PR #177, opened 2018-10-26
NC-1675 Ensure that default logging is appropriate

shemnon:
Do not log stack traces to console output.

## PR description

Do not long the stack traces to the console output.  This can be overridden by log4j2 magic.


ajsutton:
This seems like a really bad idea - stack traces are vital in understanding what went wrong if there are errors and we don’t have a different output that would provide more detail.

shemnon:
Would limiting the number of stack trace lines be better?  5, 10, 20, 30?

ekellstrand:
Per conversation in Zoom,  I think the case you are hitting just needs to be fixed,  and ideally, figure out how to get more context into that exception.  disabling trace information will really cause problems in getting incomplete issues logged.  For dev purposes, you can adjust logging on your box to fit your preferences.

shemnon:
Tabling this for now.  I think the real discussion is going to be about logging philosophy and users.  Devs want the stack traces but admins don't want it for ignorable issues.  It also dovetails into a question about log4j2 and SLF4J.

PR #203, opened 2018-10-29
Add benchmark for BlockHashOperation

ajsutton:
## PR description

Adds a micro-benchmark for the BlockHashOperation and a reusable `OperationBenchmarkHelper` to make it easier to write benchmarks for other operations.


ethereum/core/src/jmh/java/tech/pegasys/pantheon/ethereum/vm/operations/OperationBenchmarkHelper.java
```diff
@@ -0,0 +1,113 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+import com.google.common.io.MoreFiles;
+import com.google.common.io.RecursiveDeleteOption;
+
+public class OperationBenchmarkHelper {
+
+  private final Path storageDirectory;
+  private final RocksDbKeyValueStorage keyValueStorage;
```
shemnon: Does this need to be RocksDbKeyValueStorage?  Can we use the abstract parent?

ethereum/core/src/jmh/java/tech/pegasys/pantheon/ethereum/vm/operations/OperationBenchmarkHelper.java
```diff
@@ -0,0 +1,113 @@
...
+      final RocksDbKeyValueStorage keyValueStorage,
+      final MessageFrame messageFrame) {
+    this.storageDirectory = storageDirectory;
+    this.keyValueStorage = keyValueStorage;
+    this.messageFrame = messageFrame;
+  }
+
+  public static OperationBenchmarkHelper create() throws IOException {
+    final Path storageDirectory = Files.createTempDirectory("benchmark");
+    final RocksDbKeyValueStorage keyValueStorage = RocksDbKeyValueStorage.create(storageDirectory);
```
shemnon: Could we use Provider<KeyValueStorage> and pass it into create?

ethereum/core/src/jmh/java/tech/pegasys/pantheon/ethereum/vm/operations/OperationBenchmarkHelper.java
ajsutton: We need access to the `close` method which is only on `RocksDbKeyValueStorage`.  I think I might follow up with another PR to add `close` to the `KeyValueStorage` interface which would improve this kind of thing in a few places, but outside this scope of this change.

PR #204, opened 2018-10-29
Acceptance test - refactor JsonRpc to use conditionals

CjHare:
## PR description
The main goal of this PR was re-writing the JSON-RPC acceptance tests to use the conditionals and through restricting access reduce the opportunity to create non-conditional style AT.


acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/RpcApisTogglesAcceptanceTest.java
```diff
@@ -45,37 +42,40 @@ public void before() throws Exception {
...
-    rpcDisabledNode.verifyJsonRpcDisabled();
+    final String expectedMessage = "Failed to connect to /127.0.0.1:8545";
+
+    rpcDisabledNode.verify(net.netVersionExceptional(expectedMessage));
   }
 
   @Test
   public void shouldSucceedConnectingToNodeWithWsRpcEnabled() {
-    rpcEnabledNode.verifyWsRpcEnabled();
+    rpcEnabledNode.jsonRpcByWebSockets();
```
ajsutton: Maybe `useWebSocketsForJsonRpc`?

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/condition/eth/ExpectEthAccountsException.java
```diff
@@ -0,0 +1,41 @@
...
+      final EthAccountsTransaction transaction, final String expectedMessage) {
+    this.expectedMessage = expectedMessage;
+    this.transaction = transaction;
+  }
+
+  @Override
+  public void verify(final Node node) {
+    final Throwable thrown = catchThrowable(() -> node.execute(transaction));
+    assertThat(thrown).isInstanceOf(ClientConnectionException.class);
+    assertThat(thrown.getMessage()).contains(expectedMessage);
```
ajsutton: Maybe something for the future, but this feels like we could have a `ExpectExceptionCondition` that delegates to another condition (or maybe just `Runnable`) and then checks the exception thrown.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/condition/eth/ExpectEthGetWorkHasThreeNotNullValues.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+package tech.pegasys.pantheon.tests.acceptance.dsl.condition.eth;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import tech.pegasys.pantheon.tests.acceptance.dsl.condition.Condition;
+import tech.pegasys.pantheon.tests.acceptance.dsl.node.Node;
+import tech.pegasys.pantheon.tests.acceptance.dsl.transaction.eth.EthGetWorkTransaction;
+
+public class ExpectEthGetWorkHasThreeNotNullValues implements Condition {
```
ajsutton: nit: Would this be clearer if it were named `SanityCheckEthGetWorkValues`?  Just to be explicit that the requirement isn't really just that there's three arbitrary, non-null values but that's the most we can actually verify about this data.

PR #205, opened 2018-10-29
Use reliable dependency mirror while on jenkins

Errorific:
## PR description

There's now a Nexus mirror running under our control that will proxy our requirements from bintray, this mirror should prove significantly less flakey and reduce/eliminate the failures we get from dependency resolution. These changes get the build to use that mirror when it detects it's running on our jenkins. Was some additional work to get the quickstart build to be aware of it.


Errorific:
This doesn't make gradle use the new mirror for it's own dependencies, I'll follow this up with another PR to make that happen

PR #213, opened 2018-10-31
NC-1378: Adding error handling to http requests on websocket endpoint

lucassaldanha:
## PR description

Adding error handling to http requests on websocket endpoint.


ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/websocket/WebSocketService.java
```diff
@@ -103,6 +105,12 @@ public WebSocketService(
     };
   }
 
+  private Handler<HttpServerRequest> httpHandler() {
+    System.out.println("HTTP Handler");
```
CjHare: S.o.p line, why not use a Logger?

lucassaldanha: Nah, I forgot to delete this line :)
I'm impressed our static code checked doesn't complain about System.out.println statements on the code.

PR #234, opened 2018-11-03
fixes #233 Allow nodekey to be specified separately to --datadir

### LINKED ISSUE #233: Allow nodekey to be specified separately to --datadir
### Description
As an Admin, I want to be able to specify the location of the node private key separately to the data directory so that I can manage the lifecycle of the key separately (such as via a in memory filesystem). 

### Acceptance Criteria
- GIVEN a pantheon environment with an existing node key in a separate filesystem to an existing database
- WHEN I run a suitable command (similar to `--nodekey` in geth or otherwise)
- THEN Pantheon starts using the specified nodekey for its identity

### Versions (Add all that apply)
* Software version: [`pantheon --version`]
```sh
# /opt/pantheon/pantheon-entrypoint.sh --version
pantheon/0.8.1
```
* Java version: [`java -version`]
```
# java -version
openjdk version "1.8.0_171"
OpenJDK Runtime Environment (IcedTea 3.8.0) (Alpine 8.171.11-r0)
OpenJDK 64-Bit Server VM (build 25.171-b11, mixed mode)
```
* OS Name & Version: [`cat /etc/*release`]
```
# cat /etc/*release
3.8.1
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.8.1
PRETTY_NAME="Alpine Linux v3.8"
HOME_URL="http://alpinelinux.org"
BUG_REPORT_URL="http://bugs.alpinelinux.org"
```
* Kernel Version: [`uname -a`]
```
# uname -a
Linux ec4d460bdd0d 4.9.93-linuxkit-aufs #1 SMP Wed Jun 6 16:55:56 UTC 2018 x86_64 Linux
```
* Docker Version: [`docker version`]
`18.06.1-ce`

### Additional Information

- Non-blocking issue found during the awesome PegaSys hackathon 🦄 👨‍💻 👩‍💻

NicolasMassart:
## PR description
added a --nodekey option to the CLI to be able to define a separate path for the private key when starting pantheon.

## Fixed Issue(s)

fixes #233


ekellstrand:
should the --nodekey param be the actual file instead of a directory that contains "./key"?

NicolasMassart:
You're right. I will check that tomorrow. I did it this way because I was focused in using the datadir as the default value using a Path type, but it's easy to do it the way you say.

NicolasMassart:
@ekellstrand change from Path to File done

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/ThreadPantheonNodeRunner.java
```diff
@@ -59,7 +59,8 @@ public void startNode(final PantheonNode node) {
               ethNetworkConfig,
               false,
               node.getMiningParameters(),
-              true);
+              true,
+              node.homeDirectory().resolve("key").toFile());
```
ajsutton: Can we avoid the duplication of homeDirectory.resolve("key").toFile() here, in PantheonNode and in PantheonCommand somehow?  Maybe  we need `KeyPairUtil.getDefaultKeyFile(homeDirectory)`?

pantheon/src/test/java/tech/pegasys/pantheon/cli/PantheonCommandTest.java
```diff
@@ -328,14 +336,36 @@ public void NoOverrideDefaultValuesIfKeyIsNotPresentInConfigFile() throws IOExce
...
+        .build(
+            any(),
+            isNotNull(),
+            any(),
+            eq(false),
+            any(),
+            anyBoolean(),
+            fileArgumentCaptor.capture());
+
+    assertThat(fileArgumentCaptor.getValue().compareTo(file)).isZero();
```
ajsutton: This is a particularly unusual form of assertion and won't provide very useful feedback.  Is `assertThat(fileArgumentCaptor.getValue()).isEqualTo(file)` not sufficient?  If not maybe you could just compare the absolute file paths?

NicolasMassart: Will try that, thanks.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/ThreadPantheonNodeRunner.java
NicolasMassart: Ok, I will do that.

PR #235, opened 2018-11-03
Nc 1815

shemnon:
## PR description

as recommended in #231 alter the API so that the
subclasses always return their list from executePeerTask in the
`CompletableFuture` and reset the retry in `#executeTask()` when any
non-empty list is returned.


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/AbstractRetryingPeerTask.java
```diff
@@ -18,23 +18,20 @@
...
- * <p>As an additional semantic subclasses may call {@link #resetRetryCounter} so that if they have
- * partial success they can reset the retry counter and only count zero progress retries against the
- * exception limit. If this facility is used only consecutive zero progress retries count against
- * the maximum retries limit.
- *
- * @param <T> The type of the CompletableFuture this task will return.
+ * @param <T> The type as a typed list that the peer task can get partial or full results in.
  */
-public abstract class AbstractRetryingPeerTask<T> extends AbstractEthTask<T> {
+public abstract class AbstractRetryingPeerTask<T extends List<?>> extends AbstractEthTask<T> {
```
mbaxter: Does it make sense to use the more generic `Collection` here?

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/AbstractRetryingPeerTask.java
```diff
@@ -68,12 +65,16 @@ protected void executeTask() {
               if (error != null) {
                 handleTaskError(error);
               } else {
+                // If we get a partial success reset the retry counter.
+                if (peerResult.size() > 0) {
+                  retryCount = 0;
```
mbaxter: You can remove `resetRetryCounter()` now.

shemnon: Github merge restored it.

shemnon: note to self: never use github web merge again.

PR #240, opened 2018-11-05
Fix ECREC precompiled contract name

matt9ucci:
## PR description
`ECREC_PRECOMPILED_GAS_COST` -> `ECREC`

## Fixed Issue(s)

`ECREC_PRECOMPILED_GAS_COST` looks improper. It is probably derived from the constant `ECREC_PRECOMPILED_GAS_COST = Gas.of(3_000L)` in the `FrontierGasCalculator` class. But the name is not for a gas calculator but for a precompiled contract.


matt9ucci:
I think it's not such a bad idea to keep `AbstractPrecompiledContract` class. It can provide `gasRequirement()` method's default implementation, for example:

```java
public abstract class AbstractPrecompiledContract implements PrecompiledContract {
  private final GasCalculator gasCalculator;
  public Gas gasRequirement() {
    gasCalculator.calc();
  }
}
```

In order to do that, we have to redesign `GasCalculator` interface, which has many methods such as `getEcrecPrecompiledContractGasCost()`, `sha256PrecompiledContractGasCost(BytesValue input)`, etc. It would be preferred for `GasCalculator` to have just only the `calc()` method like the above example and split current methods into individual classes such as `EcrecPrecompiledContractGasCalculator.java`, `SHA256PrecompiledContractGasCalculator.java`, etc. In that case, users who develop their own private network can easily add their own custom gas calculator classes.

ajsutton:
Apologies for the delay in responding... The GasCalculator interface is designed to support the varying gas costs in different milestones so it works really well for it to be the collection of gas cost calculators. There isn't a common interface that works for all of the individual gas calculations because many of which require specific inputs such as the code size or storage values being set.

Issue #245, opened 2018-11-10
MainnetBlockBodyValidator | Invalid block: receipts root mismatch


5chdn:
### Description
As an Node, I want Consensus so that I can sync the chain. 

### Acceptance Criteria
* Import block 9554 from the Görli chain (`0x9ba7bbf8487006902c91ded8ed04cf80ad3a6efa6e27d2c53d14f05bc5e235dc`)

### Steps to Reproduce (Bug)
1. Get the Görli genesis https://github.com/goerli/testnet/blob/master/geth/goerli.genesis
2. Sync pantheon with Görli bootnodes https://github.com/goerli/testnet/blob/master/bootnodes.txt
3. Get stuck on a different fork after block 9554 http://ethstats.goerli.ethberl.in:3000/

**Expected behavior:** 
Pantheon should be in consensus with Geth and Parity.

**Actual behavior:**
- ReceiptsRoot Mismatch with Pantheon Clique, ref https://github.com/goerli/testnet/issues/6

```
2018-11-09 10:09:21.055+00:00 | EthScheduler-Workers | WARN  | MainnetBlockBodyValidator | Invalid block: receipts root mismatch (expected=0x3dc2cb692e7b8cd4276e523dfd031960d2e09bc24fcc5ccf990434c697e67d87, actual=0x2aa024623680a960cc38a96838a71e2463b0d182a897f95ea928be00cb5560c9)
2018-11-09 10:09:21.056+00:00 | EthScheduler-Workers | WARN  | PipelinedImportChainSegmentTask | Import of chain segment (9400 to 9600) failed: Failed to import block: 9554, 0x9ba7bbf8487006902c91ded8ed04cf80ad3a6efa6e27d2c53d14f05bc5e235dc.
```

Logs: https://gist.github.com/JohnnySheffield/c3deae0f33d96ec6105a3b5bc2d27060

**Frequency:** [What percentage of the time does it occur?]
Over 9000

### Versions (Add all that apply)
* Software version: master @ 2ee9cb149777ad9ed4612e2dcc2e800d8aea2633
* Java version: openjdk version "1.8.0_181"
* OS Name & Version: Ubuntu 16.04.5 LTS
* Kernel Version: Linux qa-afri-test-v20794 4.15.0-1030-azure #31~16.04.1-Ubuntu SMP Tue Oct 30 19:40:01 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

### Additional Information

[...]

PR #250, opened 2018-11-12
Common protocol schedule config handling

ajsutton:
## PR description

To ensure milestone configuration is consistent across consensus algorithms, introduce a `ProtocolScheduleBuilder` that sets up milestones.

NOTE: This changes handling of ethash genesis config files so that if milestones are not listed in config, they are not activated instead of defaulting to the Mainnet activation block numbers. Our MainNet config is unaffected as it already included the scheduled milestones (as is typical for all genesis configs).  This new behaviour seems to be consistent with other clients. It also avoids confusion when we schedule a Constantinople fork on MainNet as previously that would have automatically applied to all ethash configs that didn't explicitly disable Constantinople.

## Fixed Issue(s)

Follow up to #245 to ensure we don't forget to support future milestones in Clique and other consensus algorithms.


consensus/clique/src/main/java/tech/pegasys/pantheon/consensus/clique/CliqueProtocolSchedule.java
```diff
@@ -27,44 +35,33 @@
...
-        .getConstantinopleBlockNumber()
-        .ifPresent(
-            blockNumber -> protocolSchedule.putMilestone(blockNumber, specs.constantinople()));
+    final EpochManager epochManager = new EpochManager(cliqueConfig.getEpochLength());
+    return new ProtocolScheduleBuilder<>(
+            config,
+            builder ->
+                applyCliqueSpecificModifications(
+                    epochManager, cliqueConfig.getBlockPeriodSeconds(), localNodeAddress, builder))
+        .createProtocolSchedule(DEFAULT_CHAIN_ID);
```
rain-on: Passing in a magic default is a bit of smell - reads as though we are creating a protocol schedule of the DEFAULT_CHAIN_ID (rather than this being an override).
Also - can't quite appreciate the difference between passing everything in at construction vs in the call to createProtocolSchedule.

consensus/clique/src/main/java/tech/pegasys/pantheon/consensus/clique/CliqueProtocolSchedule.java
```diff
@@ -27,44 +35,33 @@
...
-    // Get Config Data
     final CliqueConfigOptions cliqueConfig = config.getCliqueConfigOptions();
-    final long epochLength = cliqueConfig.getEpochLength();
-    final long blockPeriod = cliqueConfig.getBlockPeriodSeconds();
-    final int chainId = config.getChainId().orElse(DEFAULT_CHAIN_ID);
-
-    final MutableProtocolSchedule<CliqueContext> protocolSchedule =
-        new MutableProtocolSchedule<>(chainId);
 
     // TODO(tmm) replace address with passed in node data (coming later)
```
rain-on: can you remove this todo? Appears its fulfilled.

ajsutton: Done

consensus/clique/src/main/java/tech/pegasys/pantheon/consensus/clique/CliqueProtocolSchedule.java
ajsutton: Yeah our handling of default chain ID is definitely weird, but not new here (I think it used to be a bit better hidden).  I suspect we should default to no chain ID if one isn't in the config file which would clean this up but that probably requires some additional research and thought.

I have changed it so that the default chain ID is passed into the constructor instead of the createProtocolSchedule method to be more consistent.

config/src/main/java/tech/pegasys/pantheon/config/GenesisConfigOptions.java
```diff
@@ -87,4 +89,47 @@ private OptionalLong getOptionalLong(final String key) {
         ? OptionalLong.of(configRoot.getLong(key))
         : OptionalLong.empty();
   }
+
+  public static class Builder {
```
rain-on: nit: This builder could/should also handle the consensus parameters - but given it is currently only used for testing, does not need repairing now.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/ProtocolScheduleBuilder.java
```diff
@@ -0,0 +1,96 @@
...
+    addProtocolSpec(
+        protocolSchedule,
+        config.getHomesteadBlockNumber(),
+        MainnetProtocolSpecs.homesteadDefinition());
+
+    config
+        .getDaoForkBlock()
+        .ifPresent(
+            daoBlockNumber -> {
+              if (daoBlockNumber > 0) {
```
rain-on: Slight concern that if the DAO fork occurs anywhere other than Frontier or Homestead, that the protocol spec will revert to an incorrect spec.

Given Dao fork should only occur on mainnet, and it definitively occurred in homestead - this is probably fine.

ajsutton: Agreed on both counts. It's messy but looking at what the Dao fork does, chances of success on anything other than MainNet at the specific time of the Dao fork are almost certain to end in tears.

Issue #251, opened 2018-11-12
CompressionException: Snappy decompression failed


5chdn:
### Description
As a Node Operator, I want ... whatever :laughing:  

### Acceptance Criteria
* Connect to other peers than Pantheon

### Steps to Reproduce (Bug)
1. Checkout latest master @ 5566eb36d83c7323ef27444a2c119dd707fea2e5
2. Connect to Görli https://github.com/goerli/testnet
3. See it fails with `CompressionException` and disconnects any peer that is not Pantheon

**Expected behavior:** 
Sync the network with all peers

**Actual behavior:**
My pantheon nodes connect to each other but not with parity or geth

**Frequency:**
100%

### Versions (Add all that apply)
```
rando@qa-afri-test-v01fa3:~/panth$ git log -n1 | head -n3
commit 5566eb36d83c7323ef27444a2c119dd707fea2e5
Author: Adrian Sutton <adrian@symphonious.net>
Date:   Mon Nov 12 12:42:20 2018 +1000
rando@qa-afri-test-v01fa3:~/panth$ java -version
openjdk version "1.8.0_181"
OpenJDK Runtime Environment (build 1.8.0_181-8u181-b13-1ubuntu0.16.04.1-b13)
OpenJDK 64-Bit Server VM (build 25.181-b13, mixed mode)
rando@qa-afri-test-v01fa3:~/panth$ cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=16.04
DISTRIB_CODENAME=xenial
DISTRIB_DESCRIPTION="Ubuntu 16.04.5 LTS"
rando@qa-afri-test-v01fa3:~/panth$ uname -a
Linux qa-afri-test-v01fa3 4.15.0-1030-azure #31~16.04.1-Ubuntu SMP Tue Oct 30 19:40:01 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```
### Additional Information

```
6|panth  | 2018-11-12 09:56:06.672+00:00 | nioEventLoopGroup-3-1 | INFO  | WaitForPeerTask | Waiting for new peer connection. 5 peers currently connected, Some peers idle.
6|panth  | 2018-11-12 09:56:06.672+00:00 | nioEventLoopGroup-3-1 | INFO  | WaitForPeerTask | Finished waiting for peer connection.
[...]

PR #254, opened 2018-11-12
Ban Peers via CLI

shemnon:
## PR description

As part of working on #251 I needed to be able to ban certain nodes from my
connection pool and let others connect.  This is a general solution to add
a --banned-nodeids CLI flag where the nodeIds of banned nodes are listed.


ajsutton:
I think if peers are blacklisted from the command line they should stay blacklisted permanently, not be subject to the usual limit of 500 blacklisted peers. It would be very confusing for users to have nodes they explicitly blacklisted wind up being able to connect.

MadelineMurray:
Raised Jira Issue NC-1874

Issue #228, opened 2018-11-01
Client stops syncing after a while because of losing p2p connections 


steffenkux:
### Description
Syncing the clinet from scratch.

### Acceptance Criteria
* get syncronized 

### Steps to Reproduce (Bug)
1. install
2. start client (on windows10)
3. wait a while

**Expected behavior:** [What you expect to happen]
* keeps syncing
* has contact to other nodes

**Actual behavior:** [What actually happens]
* stops syncing
* loses connection to other clients (?)

**Frequency:** [What percentage of the time does it occur?]
* always, somtimes it syncs a little bit, somtimes not

### Versions (Add all that apply)
* Software version: 0.8.1
* Java version: 8u191
* OS Name & Version: win10 pro 
* Kernel Version: 1803

### Additional Information
see errorlog (appended)
[errorlog_pantheon.txt](https://github.com/PegaSysEng/pantheon/files/2538034/errorlog_pantheon.txt)

steffenkux:
[errorlog_pantheon.txt](https://github.com/PegaSysEng/pantheon/files/2538038/errorlog_pantheon.txt)

lucassaldanha:
Hi @steffenkux ,

Could you try running it using the `--p2p-listen` property to advertise your public IP properly?
You just need to add the property on the startup command `--p2p-listen=host:30303`, replacing `host` with your public IP.

Also, ensure that TCP and UDP traffic is allowed on port 30303 in you firewall settings.

Let me know if this solves the issue.

5chdn:
Having the same issue: https://gist.github.com/5chdn/1939f043078905580a80ffd686c6c375

```
2018-11-01 12:38:09.984+01:00 | nioEventLoopGroup-3-1 | ERROR | Downloader | Encountered error importing blocks
tech.pegasys.pantheon.ethereum.eth.manager.exceptions.MaxRetriesReachedException: Task failed: MAX_RETRIES_REACHED
	at tech.pegasys.pantheon.ethereum.eth.manager.AbstractRetryingPeerTask.executeTask(AbstractRetryingPeerTask.java:45) [pantheon-eth-0.9.0-SNAPSHOT.jar:0.9.0-SNAPSHOT]
[...]

PR #255, opened 2018-11-12
NC-1815 Rinkeby import can stall with too many fragments

shemnon:
## PR description

Move "complete what we just got" semantic into the
DownloadHeaderSequenceTask.  Previously it was completing everything
in the chunk it was assigned.

Builds off of more generic tests by mbaxter.

## Fixed Issue(s)

NC-1815


ajsutton:
This is the more complete fix for https://github.com/PegaSysEng/pantheon/issues/228 as well right?

shemnon:
This is the more complete fix to #228, correct.

PR #283, opened 2018-11-20
NC-1890: Acceptance test smart contract

CjHare:
## PR description
Re-writing the AT for the Emitter smart contract to keep it consistent with the Node/Transaction pattern used across the other AT.


acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/web3j/EventEmitterAcceptanceTest.java
```diff
@@ -53,21 +47,23 @@ public void setUp() throws Exception {
...
           subscriptionReceived.set(true);
           assertEquals(BigInteger.valueOf(12), storedEventResponse._amount);
         });
 
-    final TransactionReceipt send = eventEmitter.store(BigInteger.valueOf(12)).send();
+    final TransactionReceipt receipt = eventEmitter.store(BigInteger.valueOf(12)).send();
 
+    assertNotNull(receipt);
     assertEquals(BigInteger.valueOf(12), eventEmitter.value().send());
     assertTrue(subscriptionReceived.get());
```
ajsutton: The assertions here are still not as declarative as I'd ideally like but it's hard to see how to improve it without losing the very handy Java wrappers that web3j provides. Probably have to live with it for now.

PR #287, opened 2018-11-20
Remove unnecessary RlpInput and RlpOutput classes

mbaxter:
## PR description
As part of removing `RlpUtils`, I noticed that we have some other unnecessary rlp classes.  This PR removes `VertxBufferRLPInput` and `VertxBufferRLPOutput`.

Our `BytesValue` library has utilities for wrapping around various types of bytes values including vertx `Buffer`'s.  Instead of using custom rlp classes for vertx buffers, we should just wrap these buffers as `BytesValue`'s and use the corresponding `BytesValueRlpInput` and `BytesValueRlpOutput` classes.  

Also removed `FileRLPInput` as it was unused (and we probably want to treat files as another type of `BytesValue` if we did need this functionality).


util/src/main/java/tech/pegasys/pantheon/util/bytes/BytesValue.java
```diff
@@ -139,6 +139,21 @@ static BytesValue wrapBuffer(final Buffer buffer) {
     return wrapBuffer(buffer, 0, buffer.length());
   }
 
+  /**
+   * Wraps a full Vert.x {@link Buffer} as a {@link BytesValue}.
+   *
+   * <p>Note that as the buffer is wrapped, any change to the content of that buffer may be
+   * reflected in the returned value.
+   *
+   * @param buffer The buffer to wrap.
```
s-matthew-english: this is a useless comment, but what about "`The {@link Buffer} to wrap.`"

mbaxter: Fair point, but I think I'll just leave it as-is so that the javadoc across this class is consistent.

PR #295, opened 2018-11-22
[NC-1742] Added host whitelist for Json-RPC.

mark-terry:
## PR description
Implemented a hostname whitelist for the JSON RPC service, as suggested by trail-of-bits to avoid DNS rebind attacks.

New behaviour:
A new CLI option has been added - `--host-whitelist`.  This argument can take a few values:
- A list of comma separated hostnames, permitting connections from the specified hosts
- `*` enables connections from all hosts

If this argument is not specified, it will default to `localhost` and only permit connections that contain `localhost` in the `Host` header of the request.

## Fixed Issue(s)
fixes #244


ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/JsonRpcHttpService.java
```diff
@@ -148,6 +156,29 @@ private void validateConfig(final JsonRpcConfiguration config) {
     return resultFuture;
   }
 
+  private Handler<RoutingContext> checkWhitelistHostHeader() {
+    return event -> {
+      Optional<String> hostHeader =
+          Optional.ofNullable(Iterables.get(Splitter.on(':').split(event.request().host()), 0));
+      if ((config.hostWhitelist().size() == 1 && config.hostWhitelist().contains("*"))
```
CjHare: Is this a validation that when `*` is used that there is also only one entry in the whitelist?

mark-terry: Yea, figured it'd be poor practice to define a wildcard _and_ explicit hostnames.  The documentation would specify that `*` can only be used by itself, to ensure people understand what they're doing.

Thoughts on my reasoning?

mark-terry: Apparently Jake is the person to raise this with. 👍

CjHare: My point was going to be a user experience perspective one. It'd be smoother to perform the validation at the earliest point and fail with an appropriate error (i.e. using * with another input), rather than at this point in the code ...as it might not be clear to the user what they have done to cause the failure.

ethereum/jsonrpc/src/test/java/tech/pegasys/pantheon/ethereum/jsonrpc/JsonRpcHttpServiceHostWhitelistTest.java
```diff
@@ -0,0 +1,160 @@
...
+  protected static final Vertx vertx = Vertx.vertx();
+
+  private static Map<String, JsonRpcMethod> rpcMethods;
+  private static JsonRpcHttpService service;
+  private static OkHttpClient client;
+  private static String baseUrl;
+  private static final MediaType JSON = MediaType.parse("application/json; charset=utf-8");
+  private static final String CLIENT_VERSION = "TestClientVersion/0.1.0";
+  private static final int CHAIN_ID = 123;
+  protected static BlockchainQueries blockchainQueries;
```
CjHare: `protected` modifier? can't `private` be used?

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/JsonRpcConfiguration.java
```diff
@@ -28,6 +28,7 @@
   private String host;
   private Collection<String> corsAllowedDomains = Collections.emptyList();
   private Collection<RpcApi> rpcApis;
+  private Collection<String> hostWhitelist = Collections.singletonList("localhost");
```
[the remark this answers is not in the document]
mark-terry: Seems reasonable.  If it were just a host, we wouldn't need a list.  👍

ethereum/jsonrpc/src/test/java/tech/pegasys/pantheon/ethereum/jsonrpc/JsonRpcHttpServiceHostWhitelistTest.java
mark-terry: Good spot.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/JsonRpcHttpService.java
mark-terry: I see, I'll see if we can move this logic back to the CLI parameter validation.

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -328,6 +328,17 @@ public RpcApi convert(final String name) throws RpcApisConversionException {
...
+  @Option(
+    names = {"--host-whitelist"},
+    paramLabel = "<hostname>",
+    split = ",",
+    arity = "1..*",
+    description =
+        "Comma separated list of hostnames to whitelist for RPC access.  default: ${DEFAULT-VALUE}",
+    defaultValue = "localhost"
+  )
+  private final Collection<String> hostWhitelist = null;
```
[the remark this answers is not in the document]
mark-terry: As discussed, I've implemented 'all'.  'none' makes sense to just turn rpc off.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/JsonRpcConfiguration.java
```diff
@@ -87,13 +88,22 @@ public void addRpcApi(final RpcApi rpcApi) {
     rpcApis.add(rpcApi);
   }
 
+  public Collection<String> hostsWhitelist() {
```
ajsutton: Should be `getHostsWhitelist`

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/JsonRpcConfiguration.java
```diff
@@ -87,13 +88,22 @@ public void addRpcApi(final RpcApi rpcApi) {
...
+    return this.hostsWhitelist;
+  }
+
+  public void hostsWhitelist(final Collection<String> hostsWhitelist) {
```
ajsutton: Should be `setHostsWhitelist`

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/JsonRpcHttpService.java
```diff
@@ -161,6 +169,29 @@ private void validateConfig(final JsonRpcConfiguration config) {
     return resultFuture;
   }
 
+  private Handler<RoutingContext> checkWhitelistHostHeader() {
+    return event -> {
+      Optional<String> hostHeader =
+          Optional.ofNullable(Iterables.get(Splitter.on(':').split(event.request().host()), 0));
```
ajsutton: What happens if an invalid Host header is sent that contains more than one colon?

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/JsonRpcHttpService.java
```diff
@@ -161,6 +169,29 @@ private void validateConfig(final JsonRpcConfiguration config) {
...
+      if (config.hostsWhitelist().contains("*")
+          || (hostHeader.isPresent()
+              && config
+                  .hostsWhitelist()
+                  .stream()
+                  .anyMatch(
+                      whitelistEntry ->
+                          whitelistEntry.toLowerCase().equals(hostHeader.get().toLowerCase())))) {
```
ajsutton: It would probably be helpful to extract the second part of this or condition to a `hostIsInWhitelist` method.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/JsonRpcHttpService.java
```diff
@@ -161,6 +169,29 @@ private void validateConfig(final JsonRpcConfiguration config) {
...
+        event.next();
+      } else {
+        event
+            .response()
+            .setStatusCode(403)
+            .putHeader("content-type", "application/json; charset=utf-8")
```
ajsutton: Shouldn't this be `Content-Type` with capitals?

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/JsonRpcHttpService.java
mark-terry: We take the first item from the split and do the checks on it.  If `*` or `all` are specified it'll get through, otherwise it'll get matched against the specified whitelist and _likely_ fail.

I think it'll still work as intended.

ajsutton: I would have though it if the host was: `some:host:8080` you should require `some:host` to be in the whitelist and reject if only `some` is in the whitelist.  That said, I'm not sure if the request would make it through vertx at all since it's a pretty seriously invalid http header.

PR #303, opened 2018-11-23
Dependencies Version upgrade

shemnon:
## PR description
Upgrade dependency versions
First, Merge all version information into one place
Next, Upgrade most version to their latest non-test version.
Finally, sort dependencies and normalize code format.

Two dependencies were not upgraded:

* errorprone - There are new checks that require build or source changes.
* vertex - The license to eclipse 2.0 which is not in our current approved
           license list.  Change is trivial but for tracability it should
           be done alone.


shemnon:
I did the gradle stuff by hand.  If I had the time I could write a spotless filter, but we will see if we just keep things sorted ourselves.

The version numbers has a plugin I added, just run 'gradle dependencyUpdates' and it will tell us what we _could_ update.  IMHO this would be the first thing to do after pushing out a release so we can 'bake in' new versions of the updates.

PR #314, opened 2018-11-26
Basic Ibft message validators

rain-on:
MessageValidator has been added which is responsible for determining the validity of a given prepare or commit message with respect to an existing Preprepare message.

It is also responsible for ensuring a supplied Preprepare message contains valid content, in line with the context of the validation (i.e. current consensus round, block is valid in the context of the existing blockchain etc).


consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/validation/MessageValidator.java
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    return validateDigestMatchesPreprepareBlock(msg.getUnsignedMessageData().getDigest(), msgType);
+  }
+
+  public boolean validateCommmitMessage(
+      final IbftSignedMessageData<IbftUnsignedCommitMessageData> msg) {
+    final String msgType = "Commit";
+
+    if (!isMessageInRoundFromValidatorAndMatchesPreprepareDigest(msg, msgType)) {
```
jframe: I think commit is only meant to validate that round number > current round number and not check the sequence number. Not sure if it matters or not.

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/validation/MessageValidator.java
```diff
@@ -0,0 +1,194 @@
...
+    if (preprepareMessage.isPresent()) {
+      return handleSubsequentPreprepareMessage(msg);
+    }
+
+    if (!msg.getUnsignedMessageData().getRoundIdentifier().equals(roundIdentifier)) {
+      LOG.info("Invalid Preprepare message, does not match current round.");
+      return false;
+    }
+
+    if (preprepareMessage.isPresent()) {
```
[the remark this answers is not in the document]
rain-on: Done.

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/ibftmessagedata/IbftUnsignedPrePrepareMessageData.java
```diff
@@ -59,4 +62,22 @@ public Block getBlock() {
...
+  public boolean equals(final Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    IbftUnsignedPrePrepareMessageData that = (IbftUnsignedPrePrepareMessageData) o;
+    return Objects.equals(block, that.block)
+        && Objects.equals(roundIdentifier, that.roundIdentifier);
```
Errorific: Should this equals just check the properties this class added and use super.equals for the parts defined in the superclass?

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/validation/MessageValidator.java
```diff
@@ -0,0 +1,194 @@
...
+    this.roundIdentifier = roundIdentifier;
+    this.headerValidator = headerValidator;
+    this.protocolContext = protocolContext;
+    this.parentHeader = parentHeader;
+  }
+
+  public boolean addPreprepareMessage(
+      final IbftSignedMessageData<IbftUnsignedPrePrepareMessageData> msg) {
+
+    if (preprepareMessage.isPresent()) {
```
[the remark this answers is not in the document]
rain-on: Can't use ifPresent, as we need to be able to return (ifPresent takes a consumer) - however, can pass the value through to prevent the extraneous get().

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/validation/MessageValidator.java
rain-on: I'm getting the impression that the commit message's consensusroundID needs to match the associated Preprepare message - so .... take it up with Roberto.

saltiniroberto: all good as we are discarding old commit messages

consensus/ibft/src/test/java/tech/pegasys/pantheon/consensus/ibft/validation/MessageValidatorTest.java
```diff
@@ -0,0 +1,274 @@
...
+            roundIdentifier,
+            headerValidator,
+            protocolContext,
+            parentHeader);
+
+    when(block.getHash()).thenReturn(Hash.fromHexStringLenient("1"));
+  }
+
+  @Test
+  public void receivingAPrepareMessageBeforePrePrepareFails() {
```
saltiniroberto: The names of these tests suggest that if we receive a Prepare or Commit message before we receive a Pre-prepare message then the Prepare/Commit message should be rejected. Perhaps there is a better name for these tests.

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/ibftmessagedata/IbftUnsignedPrePrepareMessageData.java
rain-on: Ditched this and made a custom equality function (can be dangerous using 'equals' in a heirarchy of classes.

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/validation/MessageValidator.java
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    return true;
+  }
+
+  private boolean handleSubsequentPreprepareMessage(
+      final IbftSignedMessageData<IbftUnsignedPrePrepareMessageData> existingMsg,
+      final IbftSignedMessageData<IbftUnsignedPrePrepareMessageData> newMsg) {
+    if (!existingMsg.getSender().equals(newMsg.getSender())) {
+      LOG.debug("Received subsequent invalid Preprepare message; sender differs from original.");
```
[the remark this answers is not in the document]
rain-on: Hopefully, if we receive a Preprepare from a Non-proposer, we will show an error indicating the supplied message was not from the expected proposer - and will thus be discarded.

This check can be removed, and the "not-proposer" check can catch subsequent failures.

consensus/ibft/src/test/java/tech/pegasys/pantheon/consensus/ibft/validation/MessageValidatorTest.java
rain-on: For the context, this is the expected behaviour - i.e. this class will state that a Prepare message is invalid if insufficient context exists to validate it.

PR #334, opened 2018-11-29
[MINOR] Render handler exception to System.err instead of .out

s-matthew-english:
## PR description

This is somewhat of a drive-by amelioration. I was going to test it but this is actually kind of hard to test, and I didn't want to spend a lot of time trying to make a good test for this, so I almost didn't change it- but anyway- I in fact did test it once, by making the method from `private` to `public` so I know that it behaves as expected.

### DIFF
--- a/ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/PeerDiscoveryAgent.java
+++ b/ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/PeerDiscoveryAgent.java
@@ -230,8 +230,9 @@ public class PeerDiscoveryAgent implements DisconnectCallback {
   private void handleException(final Throwable exception) {
     if (exception instanceof IOException) {
       LOG.debug("Packet handler exception", exception);
+    } else {
+      LOG.error("Packet handler exception", exception);
     }
-    LOG.error("Packet handler exception", exception);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/PeerDiscoveryAgent.java
```diff
@@ -222,12 +222,12 @@ private void initialize(final DatagramSocket socket, final int tcpPort) {
...
-   * This is the exception handler for uncontrolled exceptions ocurring in the packet handlers.
+   * This is the exception handler for uncontrolled exceptions occurring in the packet handlers.
    *
-   * @param throwable the exception that was raised
+   * @param exception the exception that was raised
    */
-  private void handleException(final Throwable throwable) {
-    System.out.println(throwable);
+  private void handleException(final Throwable exception) {
+    System.err.println(String.format("Packet handler exception: %s.", exception.getMessage()));
```
ajsutton: Good spot, but this should be sent to a logger, not `System.err` or `System.out` and it should include the stack trace, so:
`LOG.error("Packet handler exception", exception);`

We may want to check the exception type though - `IOException` is fairly expected in this area so probably should be debug level.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/PeerDiscoveryAgent.java
```diff
@@ -227,7 +228,10 @@ private void initialize(final DatagramSocket socket, final int tcpPort) {
    * @param exception the exception that was raised
    */
   private void handleException(final Throwable exception) {
-    System.err.println(String.format("Packet handler exception: %s.", exception.getMessage()));
+    if (exception instanceof IOException) {
+      LOG.debug("Packet handler exception", exception);
+    }
+    LOG.error("Packet handler exception", exception);
```
ajsutton: missing an else - this will log IOException at both debug and error level.

PR #337, opened 2018-11-30
NC-1880 High TX volume swamps block processing

shemnon:
## PR description

Move transaction processing into its own thread(s).

## Fixed Issue(s)

Fixes NC-1880


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/EthScheduler.java
```diff
@@ -43,6 +43,7 @@
 
   protected final ExecutorService workerExecutor;
```
mbaxter: Should we rename this `syncWorkerExecutor`?

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/EthScheduler.java
```diff
@@ -56,13 +57,21 @@
             new ThreadFactoryBuilder()
                 .setDaemon(true)
                 .setNameFormat(EthScheduler.class.getSimpleName() + "Timer")
+                .build()),
+        Executors.newFixedThreadPool(
+            workerCount,
```
mbaxter: We probably want a separate count for the tx workers, right?

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/EthScheduler.java
```diff
@@ -92,7 +101,11 @@ protected EthScheduler(
   }
 
   public Future<?> scheduleWorkerTask(final Runnable command) {
```
mbaxter: `scheduleSyncWorkerTask`?

shemnon: Sure.  Also scheduleTxWorkerTask, so both align with the field name.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/manager/EthProtocolManagerTest.java
```diff
@@ -744,4 +755,38 @@ public void shouldSuccessfullyRespondToGetHeadersRequestLessThanZero()
       done.get();
     }
   }
+
+  @Test
+  public void transactionMessagesGoToTheCorrectExecutor() {
```
mbaxter: nitpick - this might make more sense in the `TransactionPoolFactory` test, as the tx pool is the object deciding how to use the workers

pantheon/src/main/java/tech/pegasys/pantheon/controller/CliquePantheonController.java
```diff
@@ -142,7 +142,8 @@
             protocolContext.getBlockchain(),
             networkId,
             fastSyncEnabled,
-            syncConfig.downloaderParallelism());
+            syncConfig.downloaderParallelism(),
+            syncConfig.transactionsParallelism());
```
mbaxter: Nitpick (optional) - it looks like SyncConfig is becoming EthConfig - maybe we should rename and move it??

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/manager/EthProtocolManagerTest.java
shemnon: The message dispatch occurs within the EthProtocolManager, even though TransactionPool is the one registering the listener.  EthProtocolManager::processMessage calls are key to this test.  Also, this would result in core then depending on eth (at least the tests) and would result in a quasi-circular dependency.  While it does bring the test closer to the other half of the code that is being tested EthScheduler is also as much under test.  But the number of new quasi-circular imports that :ethereum:core_test would gain if I moved the test persuades me to leave it where it is.

pantheon/src/main/java/tech/pegasys/pantheon/controller/CliquePantheonController.java
shemnon: There may be more to it than a surface renaming.  Spun off to NC-1993.

PR #339, opened 2018-11-30
Repair CLique Proposer Selection

rain-on:
Clique Proposer Selection would choose an incorrect peer
when a signer was removed from the pool, as the algorithm worked
purely on the block count.

The algorithm has now been updated to ensure the next proposer is
incrementally the next signer in the ordered list, based on the
parent's proposer


shemnon:
This broke Görli, hard.  I think it should be rolled back until it doesn't break Görli

PR #341, opened 2018-11-30
Revert "Repair Clique Proposer Selection" #339 - Breaks Görli testnet

shemnon:
Reverts PegaSysEng/pantheon#339

Görli's testnet regularly has proposers miss their turn and their genesis block actually had a bad validator specified (that was voted out as soon as they could)  So many blocks do not have the proper "proposer" as their block generator.  As a consequence the CliqueDifficultyValidationRule fails regularly, in fact it fails at block 3, so this should be easy to re-create locally.


shemnon:
(a) it reverts too much (I just wanted the one change undone) and 
(b) Github won't merge it:
`Update branch attempt failed
Couldn't update "revert-339-cpr": Required status check "continuous-integration/jenkins/pr-merge" is expected. At least 1 approving review is required by reviewers with write access.`

So I am abandoning this one in favor of #343

PR #344, opened 2018-12-01
Plumb in three more metrics

shemnon:
## PR description

* add blockchain_height gauge
* add blockchain_difficulty_total gauge
* add blockchain_announcedBlock_ingest histogram

This involved some deep pluming such that the metrics system needs to be
created in the PantheonCommand, along with trickle down effects into other
consensus engines.  This is likely where it should live anyway.

Testing necessitated turning the PrometheusMetricsSystem::init method
into a singleton accessed via PrometheusMetricsSystem::instance.


shemnon:
I'm all ears for different names for the metrics.  We may want to see how they evolve and go on a renaming binge at some point.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/db/DefaultMutableBlockchain.java
```diff
@@ -51,10 +53,23 @@
...
+    metricsSystem.createGauge(
+        MetricCategory.BLOCKCHAIN,
+        "height",
+        "Height of the chainhead",
+        () -> (double) this.getChainHeadBlockNumber());
+    metricsSystem.createGauge(
+        MetricCategory.BLOCKCHAIN,
+        "difficulty_total",
+        "Total difficulty of the chainhead",
+        () -> (double) this.getChainHead().getTotalDifficulty().toLong());
```
ajsutton: This cast to double is proving to be an annoyingly common pattern.  Should we introduce a `LongSupplier` version of gauges?  Even if it just did the cast internally it would tidy code up.  Also, in hind-sight I should have used `DoubleSupplier`.  Might be a separate PR...

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockPropagationManager.java
```diff
@@ -64,21 +67,29 @@
...
     this.ethContext = ethContext;
 
     this.syncState = syncState;
     this.pendingBlocks = pendingBlocks;
+
+    this.announcedBlockIngestTimer =
+        metricsSystem.createTimer(
+            MetricCategory.BLOCKCHAIN,
+            "pantheon_blockchain_announcedBlock_ingest",
+            "Time to ingest a single announced block");
```
ajsutton: Is it meaningful to track importing an announced block separately to just importing a block?  ie should we just time how long `BlockImporter` takes as that will capture all block imports? I could see it being useful to track how long it takes from first hearing about a block to it being on-chain, but this metric doesn't quite capture that since it only starts the timer after the new block has been retrieved.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockPropagationManager.java
```diff
@@ -241,21 +252,25 @@ private void handleNewBlockHashesFromNetwork(final EthMessage message) {
...
+                // TODO do we time failures?  But we cannot drop a label in at this point.
                 LOG.warn(
                     "Failed to import announced block {} ({}).",
                     block.getHeader().getNumber(),
                     block.getHash());
               } else {
+                final double time = blockTimer.stopTimer();
                 LOG.info(
-                    "Successfully imported announced block {} ({}).",
+                    "Successfully imported announced block {} ({}) in {}ms.",
```
ajsutton: Include the time in the log message is a nice touch.

metrics/src/main/java/tech/pegasys/pantheon/metrics/prometheus/PrometheusMetricsSystem.java
```diff
@@ -48,9 +49,16 @@
   private static final String PANTHEON_PREFIX = "pantheon_";
   private final Map<MetricCategory, Collection<Collector>> collectors = new ConcurrentHashMap<>();
 
+  private static final Supplier<MetricsSystem> INSTANCE =
+      Suppliers.memoize(PrometheusMetricsSystem::init);
```
ajsutton: I think this is wrong as it makes the metrics system JVM-wide whereas we want to be able to run two separate Pantheon instances in the same JVM and have them entirely isolated.

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -422,6 +426,10 @@ public void run() {
       Configurator.setAllLevels("", logLevel);
     }
 
+    if (metricsSystem == null) {
+      metricsSystem = PrometheusMetricsSystem.instance();
+    }
```
ajsutton: When would `metricsSystem` not be null here?

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/db/DefaultMutableBlockchain.java
shemnon: Yes, because I also see the potential of a UInt256Supplier.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockPropagationManager.java
shemnon: I was timing the amount of time to process the block.  Its separate code paths IIRc between download and announced blocks, and different tasks (download I see being split into at least three non-adjacent tasks : headers, bodies, receipts).

I could push this down into the persistBlock task, but I was hoping to have the number for the logging statement.

metrics/src/main/java/tech/pegasys/pantheon/metrics/prometheus/PrometheusMetricsSystem.java
shemnon: My Prometheus Server code introduced the test breakage.  So this is reverted.

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
shemnon: It was an early attempt to address my server test breakage.

pantheon/src/test/java/tech/pegasys/pantheon/cli/PantheonCommandTest.java
```diff
@@ -95,6 +95,7 @@ public void initMocks() throws Exception {
     when(mockRunnerBuilder.webSocketConfiguration(any())).thenReturn(mockRunnerBuilder);
     when(mockRunnerBuilder.dataDir(any())).thenReturn(mockRunnerBuilder);
     when(mockRunnerBuilder.bannedNodeIds(any())).thenReturn(mockRunnerBuilder);
+    when(mockRunnerBuilder.metricsSystem(any())).thenReturn(mockRunnerBuilder);
```
shemnon: This was a whole lot more lines to add a new object to the builder before the builder remodel.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockPropagationManager.java
ajsutton: Having the timing in the log message is probably justification enough for this.  I think at some point having a timer inside `BlockImporter` itself which will capture every block imported makes sense, but there are going to be a number of different timings related to block import (e.g. timing import of each batch of blocks when synchronizing).

shemnon: Pluming it into tasks is a much larger PR.  I'll leave it here for now, but it may move.

PR #346, opened 2018-12-02
added --nodes-whitelist param to CLI and NodeWhitelistController

macfarla:
## PR description
Add --nodes-whitelist parameter for permissioning.

## Fixed Issue(s)
NC-1936


pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -377,6 +378,17 @@ public RpcApi convert(final String name) throws RpcApisConversionException {
   )
   private final BytesValue extraData = DEFAULT_EXTRA_DATA;
 
+  // Permissioning: A list of whitelist nodes can be passed.
+  @Option(
+    names = {"--nodes-whitelist"},
+    paramLabel = "<enode://id@host:port>",
+    description =
+        "Comma separated enode URLs for Permissioned networks. " + "Default is an empty list.",
```
lucassaldanha: It might be just me, but 'Default is an empty list' gives me a false impression that the default behaviour is not whitelisting anyone, therefore, not talking to any node. If the user doesn't use the 'nodes-whitelist' property, we won't do any permissioning.

Do you think this can be misleading or is it just me? :P

pantheon/src/main/java/tech/pegasys/pantheon/controller/NodeWhitelistController.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.List;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class NodeWhitelistController {
+
+  private static final Logger LOG = LogManager.getLogger();
+
+  private static List<String> nodeWhitelist;
```
lucassaldanha: We might need a flag to know if the user set the property or not. If the user didn't, we won't do any permissioning.

An empty list can be ambiguous. It can indicate:

The user didn't whitelist any node (node whitelisting toggle on - don't talk to any node)
The user didn't specify the 'nodes-whitelist' property (node whitelisting toggled off)

We can assume that a null list means that the property hasn't been set (as opposed to an empty list when the user set the property without a value). But I think a flag is more semantic.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/permissioning/PermissioningConfiguration.java
```diff
@@ -0,0 +1,37 @@
...
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package tech.pegasys.pantheon.ethereum.permissioning;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+public class PermissioningConfiguration {
```
lucassaldanha: How are we gonna differentiate an empty list 'cause the user didn't whitelist any nodes and an empty list 'cause the user didn't specify the property?

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -487,6 +500,13 @@ private WebSocketConfiguration webSocketConfiguration() {
     return webSocketConfiguration;
   }
 
+  private PermissioningConfiguration permissioningConfiguration() {
+    final PermissioningConfiguration permissioningConfiguration =
```
lucassaldanha: We need to validate that all bootnodes have been included in the nodes whitelist. I wonder if this would be the place to validate this...

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/permissioning/PermissioningConfigurationTest.java
```diff
@@ -0,0 +1,46 @@
...
+  @Test
+  public void defaultConfiguration() {
+    final PermissioningConfiguration configuration = PermissioningConfiguration.createDefault();
+    assertThat(configuration.getNodeWhitelist()).isEmpty();
+    assertThat(configuration.isNodeWhitelistSet()).isFalse();
+  }
+
+  @Test
+  public void setNodeWhitelist() {
+    final String[] nodes = {"enode://001@123:4567", "enode://002@123:4567", "enode://003@123:4567"};
```
lucassaldanha: Question: should we validate if the enode in the config file is valid? We are planning to validate it in the JSON-RPC API so maybe it makes sense to do the same when reading from the file.
Thoughts?

macfarla: https://pegasys1.atlassian.net/browse/NC-1981

PR #347, opened 2018-12-02
Log the names of any failed block header validations

jframe:
## PR description

Additional logging in the BlockHeaderValidator to log the names of any failed rules.


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/BlockHeaderValidator.java
```diff
@@ -70,10 +72,18 @@ private boolean applyRules(
...
+    final List<Rule<C>> failedRules =
+        rules
+            .stream()
+            .filter(filter)
+            .filter(rule -> !rule.validate(header, parent, protocolContext))
+            .collect(toList());
+    if (!failedRules.isEmpty()) {
+      List<String> failedRuleNames =
+          failedRules.stream().map(r -> r.rule.getClass().getSimpleName()).collect(toList());
+      LOG.error("Block header validation failed on these rules {}", failedRuleNames);
```
ajsutton: I suspect for many rules this will just print the lambda class name as detached validation rules get wrapped in an adapter so all rules just get treated as attached.  I'm also pretty sure this isn't an error.

Printing the classname is probably not the right approach anyway, two possible better alternatives:

 * Have header validations return an actual error message or enum value instead of just true/false. Then all the relevant info can be included there and the caller can decide whether to log it or not (this is used for transaction validation).
 * Have the header validation rule log messages to indicate why things are invalid (probably at debug level).

PR #352, opened 2018-12-03
[NC-1524] Update get validators for block hash test to start from block 1

jframe:
## PR description

This updates the Clique getSigners shouldBeAbleToGetValidatorsForBlockHash test to first wait for block 1 to be mined. This ensures the block numbers are consistent instead of assuming it will start from block 0. Same approach is done in the getValidatorsForBlockNumber test this test was missed unfortunately.

Also increased the block period to 10 as sometimes there are unexpected delays in starting nodes or processing blocks that affect the clique tests more if the block period is too short.


acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/clique/CliqueGetSignersRpcTest.java
```diff
@@ -18,53 +18,53 @@
...
+    minerNode1.waitUntil(wait.chainHeadIsAt(1));
 
+    minerNode1.execute(cliqueTransactions.createAddProposal(minerNode2));
     cluster.waitUntil(wait.chainHeadHasProgressed(minerNode1, 1));
-    cluster.verify(clique.validatorsAtBlockEqual("0x2", initialNodes));
+    cluster.verify(clique.validatorsAtBlockEqual("0x2", minerNode1, minerNode2));
 
-    minerNode1.waitUntil(wait.chainHeadHasProgressed(minerNode1, 1));
-    cluster.verify(clique.validatorsAtBlockEqual("0x3", allNodes));
+    minerNode1.execute(cliqueTransactions.createAddProposal(minerNode3));
```
rain-on: these lines probably don't add a lot of value - they can be removed (I suspect they do not increase coverage ... so .... remove?)

jframe: Agreed. Don't think they do and could make the test more brittle too.
Removed them

PR #354, opened 2018-12-03
Make DisconnectMessage reason optional

mbaxter:
## PR description
* Update `DisconnectMessage` to hold an optional reason
* Add tests for `DisconnectMessage` construction and parsing
* Send an empty reason when we disconnect from blacklisted peers

### DIFF
--- a/ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/transactions/TestNodeList.java
+++ b/ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/transactions/TestNodeList.java
@@ -32,6 +32,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
@@ -218,7 +219,7 @@ public class TestNodeList implements Closeable {
     int errCnt = 0;
     final StringBuilder sb = new StringBuilder();
     for (final TestNode node : nodes) {
-      for (final Map.Entry<PeerConnection, DisconnectReason> entry :
+      for (final Map.Entry<PeerConnection, Optional<DisconnectReason>> entry :
           node.disconnections.entrySet()) {
         final PeerConnection peer = entry.getKey();
         final String peerString = peer.getPeer().getNodeId() + "@" + peer.getRemoteAddress();
--- a/ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/transactions/TestNodeList.java
+++ b/ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/transactions/TestNodeList.java
@@ -228,8 +228,8 @@ public class TestNodeList implements Closeable {
                 + node.shortId()
                 + " has received a disconnection from "
                 + peerString
-                + " for "
-                + entry.getValue()
+                + " for reason "
+                + entry.getValue().map(DisconnectReason::toString).orElse("none")
                 + "\n";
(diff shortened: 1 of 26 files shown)


ethereum/mock-p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/testing/MockNetwork.java
```diff
@@ -246,13 +247,14 @@ public PeerInfo getPeer() {
...
-    public void terminateConnection(final DisconnectReason reason, final boolean peerInitiated) {
+    public void terminateConnection(
+        final Optional<DisconnectReason> reason, final boolean peerInitiated) {
       network.disconnect(this, reason);
     }
 
     @Override
     public void disconnect(final DisconnectReason reason) {
-      network.disconnect(this, reason);
+      network.disconnect(this, Optional.ofNullable(reason));
```
shemnon: Should we push Optional<DisconnectReason> into PeerConnection?  An alternative would be to make a no-args disconnect() and make disconnect(@NotNull final DisconnectReason reason) and handle the optional stuff in the impls.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/netty/Callbacks.java
```diff
@@ -44,12 +45,20 @@
...
       final PeerConnection connection,
-      final DisconnectReason reason,
+      final Optional<DisconnectReason> reason,
       final boolean initatedByPeer) {
     disconnectCallbacks.forEach(
         consumer -> consumer.onDisconnect(connection, reason, initatedByPeer));
   }
 
+  public void invokeDisconnect(
+      final PeerConnection connection,
```
shemnon: Any reason we need both optional and non-optional forms?

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/transactions/TestNodeList.java
```diff
@@ -218,7 +219,7 @@ public void assertNoNetworkDisconnections() {
     int errCnt = 0;
     final StringBuilder sb = new StringBuilder();
     for (final TestNode node : nodes) {
-      for (final Map.Entry<PeerConnection, DisconnectReason> entry :
+      for (final Map.Entry<PeerConnection, Optional<DisconnectReason>> entry :
```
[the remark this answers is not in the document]
mbaxter: 👍 Fixed

ethereum/mock-p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/testing/MockNetwork.java
mbaxter: I cleaned this up a little, but I have `disconnect()` and `disconnect(final DisconnectReason reason)` methods on `PeerConnection`.  I thought this was cleaner than having to wrap the reasons in `Optional`'s all over.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/netty/Callbacks.java
mbaxter: Just for convenience - there are some cases where this api is called with a reason directly, and some cases where you're passing on the reason that's already wrapped in an `Optional`.

mbaxter:
@shemnon - I think https://github.com/PegaSysEng/pantheon/pull/359 turned out better - what do you think?

shemnon:
> @shemnon - I think #359 turned out better - what do you think?

#359 is much cleaner

mbaxter:
Closing this in favor of #359

PR #360, opened 2018-12-04
ReworkVoteTallyCache  to better represent purpose

rain-on:
VoteTallyCache has been refactored to simplify some logic.

Resolved issue whereby an existing voteTally could be polluted with votes from the next block.


consensus/clique/src/main/java/tech/pegasys/pantheon/consensus/clique/VoteTallyCache.java
```diff
@@ -80,24 +92,22 @@ private VoteTally populateCacheUptoAndIncluding(final BlockHeader start) {
...
 
   private VoteTally constructMissingCacheEntries(
       final Deque<BlockHeader> headers, final VoteTally tally) {
+    final VoteTally mutableVoteTally = tally.copy();
     while (!headers.isEmpty()) {
       final BlockHeader h = headers.pop();
-      voteTallyUpdater.updateForBlock(h, tally);
-      voteTallyCache.put(h.getHash(), tally.copy());
+      voteTallyUpdater.updateForBlock(h, mutableVoteTally);
+      voteTallyCache.put(h.getHash(), mutableVoteTally.copy());
```
ajsutton: For the record, this is the key to the bug fix. Previously the VoteTally that was originally retrieved from the cache was modified and then a copy added for the new block, resulting in the original and new both being the vote tally for the new block.  Now it copies before modifying so the original value is unchanged.

PR #361, opened 2018-12-04
Time all tasks

shemnon:
## PR description

This is fairly high touch consisting of 3 things:
* Moving to Prometheus's Summary for timers
  - Timing at .2, .5, .8, .9, .99, and 1.0 (1.0 actually gets max I believe)
* Timing all abstract EthTasks
* The vast bulk of the changes: plumbing the timing context everywhere we need it

A sample of the metrics can be seen here: https://gist.github.com/shemnon/13d1f57858f8b8f8ec0edf8cb4e3d0ab

I'm going to investigate alternatives to dragging the metric all across all those constructors.


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/AbstractEthTask.java
```diff
@@ -117,6 +127,17 @@ public final T result() {
   /** Execute core task logic. */
   protected abstract void executeTask();
 
+  /** Executes the task while timed by a timer. */
+  public void executeTaskTimed() {
+    final OperationTimer.TimingContext timingContext = taskTimer.startTimer();
+    executeTask();
+    taskTimeInSec = timingContext.stopTimer();
```
ajsutton: We should probably use the `try (final OperationTimer.TimingContext timingContext = taskTimer.startTimer())` or an explicit `final` block here to ensure the timer completes even if the task fails.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/sync/tasks/GetHeadersFromPeerByNumberTaskTest.java
```diff
@@ -97,7 +101,13 @@ private void getHeadersFromHash(final int skip, final boolean reverse) {
...
     final AbstractGetHeadersFromPeerTask task =
         new GetHeadersFromPeerByNumberTask(
-            protocolSchedule, ethContext, startNumber, count, skip, reverse);
+            protocolSchedule,
+            ethContext,
+            startNumber,
+            count,
+            skip,
+            reverse,
+            labels -> () -> () -> 0.0D);
```
ajsutton: Should we make this a constant somewhere? It's a very cryptic line to have scattered around the code base so much.

metrics/src/main/java/tech/pegasys/pantheon/metrics/MetricCategory.java
```diff
@@ -17,7 +17,8 @@
   RPC("rpc"),
   JVM("jvm", false),
   PROCESS("process", false),
-  BLOCKCHAIN("blockchain");
+  BLOCKCHAIN("blockchain"),
+  TASKS("tasks");
```
ajsutton: `TASKS` feels too generic as a category - is this more `SYNCHRONIZER`?

metrics/src/test/java/tech/pegasys/pantheon/metrics/prometheus/PrometheusMetricsSystemTest.java
```diff
@@ -120,23 +112,14 @@ public void shouldCreateObservationsFromTimerWithLabels() {
...
-            new Observation(RPC, "request", null, asList("method", "sum")),
-            new Observation(RPC, "request", null, asList("method", "count")));
+            new Observation(RPC, "request", null, asList("method", "0.2")),
+            new Observation(RPC, "request", null, asList("method", "0.5")),
+            new Observation(RPC, "request", null, asList("method", "0.8")),
+            new Observation(RPC, "request", null, asList("method", "0.95")),
+            new Observation(RPC, "request", null, asList("method", "0.99")),
+            new Observation(RPC, "request", null, asList("method", "1.0")),
+            new Observation(RPC, "request_sum", null, singletonList("method")),
+            new Observation(RPC, "request_count", null, singletonList("method")));
```
ajsutton: For histograms we deliberately converted the _sum and _count metrics to labels instead of completely separate metrics. It comes out much nicer in the JSON-RPC that way.  The logic is in convertHistogramSampleNamesToLabels and is probably very similar to what summaries would require.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/AbstractEthTask.java
shemnon: We need the final block because close doesn't return a time and we want to keep that for logging at higher levels.

metrics/src/test/java/tech/pegasys/pantheon/metrics/prometheus/PrometheusMetricsSystemTest.java
shemnon: Done.  I left the histogram branch in place in case we ever come back to it.

PR #362, opened 2018-12-04
Time all tasks V.2

shemnon:
## PR description

Alternate plumbing to #361


shemnon:
Surprisingly this took more lines and is less straightforward than #361.  I'll kill this after a day or so to keep for posterity.

PR #365, opened 2018-12-05
NC-1936 discovery wiring for --node-whitelist

macfarla:
## PR description
If --node-whitelist is set, discard any messages from non-whitelisted peers. 

## Fixed Issue(s)
NC-1936


ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
```diff
@@ -106,6 +107,7 @@
 
   private final PeerDiscoveryAgent agent;
   private final PeerBlacklist peerBlacklist;
+  private final NodeWhitelistController nodeWhitelistController;
```
CjHare: Suggestion: s/nodeWhiteListController/nodeWhitelist
Reason being, the term `Controller` is unnecessary, brings no additional contextual meaning.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/permissioning/NodeWhitelistController.java
```diff
@@ -10,40 +10,42 @@
...
 
-  public boolean removeNode(final String nodeId) {
-    return nodeWhitelist.remove(nodeId);
+  public boolean removeNode(final Peer node) {
+    return nodeWhitelist.remove(node);
   }
 
-  public static boolean isNodeWhitelistSet() {
-    return nodeWhitelistSet;
+  public boolean isNodeWhitelisted(final Peer node) {
```
CjHare: Suggestion: s/isNodeWhiteListed/contains
Reason: as a public scoped method, it would reduce repetition in context when calling the method i.e. nodeWhiteList.contains() vs nodeWhitelist.isNodeWhiteListed()

CjHare: If there is no white list, then everything is whitelisted. The whitelist contains the node, as it is the set of all nodes.

PR #369, opened 2018-12-06
Metrics for messages

shemnon:
## PR description

Adds metrics counters for inbound and outbound messages, with capability, name, and message code.


ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/NetworkRunner.java
```diff
@@ -115,6 +129,12 @@ private void setupHandlers() {
             cap,
             message -> {
               final int code = message.getData().getCode();
+              inboundMessageCounter
+                  .labels(
+                      cap.toString(),
+                      protocol.messageName(cap.getVersion(), code),
+                      Integer.toString(code))
```
ajsutton: If someone sends a message with an invalid code, we'll get `null` for the message name and throw `NullPointerException` here (labels can't be null).  We probably should have an "unknown" or "invalid" name for this.  We should also collapse all unknown message codes to a single code label (e.g. or "unknown" since it's a string).  Otherwise we create a new time series for every invalid message code someone sends us.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/netty/NettyPeerConnection.java
```diff
@@ -82,6 +87,16 @@ public void send(final Capability capability, final MessageData message) throws
...
+      outboundMessagesCounter
+          .labels(
+              capability.toString(),
+              subProtocol.messageName(capability.getVersion(), message.getCode()),
+              Integer.toString(message.getCode()))
+          .inc();
+    } else {
+      outboundMessagesCounter
+          .labels("null", Integer.toString(message.getCode()), Integer.toString(message.getCode()))
+          .inc();
```
ajsutton: Messages with no capability would be one of the ones from `WireMessageCodes` so maybe could add a message name function there and provide the proper name rather than just the message code.

metrics/src/main/java/tech/pegasys/pantheon/metrics/MetricCategory.java
```diff
@@ -18,7 +18,8 @@
   JVM("jvm", false),
   PROCESS("process", false),
   BLOCKCHAIN("blockchain"),
-  SYNCHRONIZER("synchronizer");
+  SYNCHRONIZER("synchronizer"),
+  NETWORK("network");
```
ajsutton: Not for this PR, but I'm really feeling like we need hierarchical categories.  The `PEERS` category is both worthy of it's own category but also very highly network related.  Having `NETWORK->PEERS` would make a lot of sense.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/NetworkRunner.java
shemnon: I'll go with "invalid" in all lowercase, to make it distinct in case there is ever an "Invalid" message.

As for message codes without message names I don't think we should collapse them.  If we are getting strange codes we will want to know how many of what kind of code they are.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/netty/NettyPeerConnection.java
shemnon: Done.

metrics/src/main/java/tech/pegasys/pantheon/metrics/MetricCategory.java
shemnon: I'll keep it in mind if I ever mess with categories.  May be deep.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/NetworkRunner.java
ajsutton: Every unique combination of labels is an entirely separate time series and increases the RAM and storage required.  It's really important to never have open-ended labels as it effectively becomes a DoS attack.

ajsutton: Just to back that up with actual docs: https://prometheus.io/docs/practices/naming/#labels

CAUTION: Remember that every unique combination of key-value label pairs represents a new time series, which can dramatically increase the amount of data stored. Do not use labels to store dimensions with high cardinality (many different label values), such as user IDs, email addresses, or other unbounded sets of values.

shemnon: 256 is not high cardinality for a metrics system, 20K was the limit at a former employer.  The message ID is one byte (Framer, line 250), so in theory there is an upper bound and not a very effective DOS attack since they are all counters.

shemnon: If we are not going to record the invalid message code we may as well not count them.  I moved the count after the invalid code check.  If it is a bug then we can crank logging up to trace and look for the log messages, not something we are likely to do in production anyway.

PR #383, opened 2018-12-09
NC-1361 Added configurable refresh delay for SyncingSubscriptionService on start up

Shinabyss:
## PR description
Added configurable refresh delay for SyncingSubscriptionService on start up allowing user to change how often user/subscribers will sync with the node on the current state.

Config has a boundary of greater than 0 and less than 3600000 (equates to 1 hour) milliseconds.

Added Unit Testing
Not suitable for acceptance testing as acceptance testing this feature require testing the time elapse of the syncing subscriptions. This is performed by vertx hence the function is tested by vertx and secondly and more importantly, attempt to assert time delay as a form of testing can result in brittle test. For instance, setting the refresh delay to 5 seconds and run test could produce measurements on delay between 0-5 seconds and does not provide conclusive test that the config is actually set to 5 seconds. A loop can be used to retest this over and over to produce a more reliable conclusion, but this is resource and time consuming on top of still only provide a higher percentage confidence the config is correctly set and not a conclusive one. And thus as per discussion with other members of product dev team.


ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/websocket/subscription/SubscriptionManager.java
```diff
@@ -49,6 +49,15 @@
...
   private final Map<String, List<Long>> connectionSubscriptionsMap = new HashMap<>();
   private final SubscriptionBuilder subscriptionBuilder = new SubscriptionBuilder();
+  private final long refreshDelay;
+
+  public SubscriptionManager(final long refreshDelay) {
+    this.refreshDelay = refreshDelay;
+  }
+
+  public SubscriptionManager() {
+    this.refreshDelay = 5000;
```
macfarla: use the const here rather than the number

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -312,6 +312,26 @@ public RpcApi convert(final String name) throws RpcApisConversionException {
...
+  @Option(
+    names = {"--ws-refresh-delay"},
+    paramLabel = "<refresh delay>",
+    arity = "1",
+    description =
+        "Refresh delay of websocket subscription sync in milliseconds. "
+            + "default: ${DEFAULT-VALUE}",
+    defaultValue = "5000"
+  )
+  private void setRefreshDelay(final Long refreshDelay) {
```
macfarla: what happens if I specify 1.5 ? Why long not int?

Shinabyss: vertx handles the sync function after we pass it in a value and it requires a long.
If 1.5 is entered, the CLI will pick it up as an error and warn user because only positive int below 3600000 is accepted.
long datatype are integers, they cannot take in decimals unlike double precision numbers

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/websocket/subscription/SubscriptionManager.java
Shinabyss: Fixed

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -312,6 +312,26 @@ public RpcApi convert(final String name) throws RpcApisConversionException {
...
+    if (refreshDelay < 1 || refreshDelay > 3600000) {
```
CjHare: Why the magic numbers?

Suggestion: refactor the values into static constants with meaningful name

Shinabyss: 5000 milliseconds is the original default when we still had it hard coded. I wanted to keep the configuration the same as the original if user started pantheon without specifying a refresh delay. Although why we decided on 5 seconds to begin with when we had it hard coded is up for debate. Do you have a better default value you would like to share?

CjHare: That's not what I mean when I say 'magic numbers' i.e. https://en.wikipedia.org/wiki/Magic_number_%28programming%29

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -312,6 +314,26 @@ public RpcApi convert(final String name) throws RpcApisConversionException {
...
+        "Refresh delay of websocket subscription sync in milliseconds. "
+            + "default: ${DEFAULT-VALUE}",
+    defaultValue = "" + WebSocketConfiguration.DEFAULT_WEBSOCKET_REFRESH_DELAY
+  )
+  private void setRefreshDelay(final Long refreshDelay) {
+    if (refreshDelay < DEFAULT_MIN_REFRESH_DELAY || refreshDelay > DEFAULT_MAX_REFRESH_DELAY) {
+      throw new ParameterException(
+          new CommandLine(this),
+          "refreshDelay must be a positive integer smaller than 3600000 (1 hour)");
+    }
```
macfarla: recommend using MAX const in the string message here too

PR #391, opened 2018-12-10
Move Block data generator test helper to test support package

jframe:
## PR description
Move BlockDataGenerator test helper to test support package so it can be used in other unit tests for creating blocks.


ethereum/core/src/test-support/java/tech/pegasys/pantheon/ethereum/core/BlockDataGenerator.java
```diff
@@ -10,26 +10,12 @@
  * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
  * specific language governing permissions and limitations under the License.
  */
-package tech.pegasys.pantheon.ethereum.testutil;
+package tech.pegasys.pantheon.ethereum.core;
```
shemnon: Not this package, we don't want to split a package across separate modules.  What is wrong with `testutil`?  Perhaps `testsupport` if we need  it out of `testutil`.  But not `core`, that should only be fore classes in `:ethereum:core`

shemnon: As this stands right now it is a technical barrier to us adopting either Java 9 modules or OSGi (mutually exclusive choices, but we want our options open).

PR #400, opened 2018-12-11
Don't advertise eth63 for regular sync

mbaxter:
## PR description
We don't fully support eth63 as node data requests are always served with an empty response.  So, just don't advertise eth63 for now.


mbaxter:
Hmm - just testing this locally and I do get a lot of disconnects from clients that are only serving eth63.  I'm going to close this for now and bring it up for discussion.

PR #403, opened 2018-12-11
Validate enodeurl syntax from command line

lucassaldanha:
## PR description

Now we are using `EnodeToURIPropertyConverter` to parse enode URLs from the command line. This applies to `--bootnodes` and `--nodes-whitelist`. Also, we are now keeping them as URI instances instead of Strings.

Unfortunately, Picocli doesn't handle well error messages for lists that accept zero elements (`--nodes-whitelist`). Therefore, the error message when using an invalid enode URL for this property is not as helpful as when you try to do the same with `--bootnodes`. I'll raise a JI ticket to investigate exxtending Picocli `DefaultExceptionHandler` to improve the error message for lists with arity `0..*`.


lucassaldanha:
This is the JI about improving the error messages: NC-2015

pantheon/src/main/java/tech/pegasys/pantheon/cli/custom/EnodeToURIPropertyConverter.java
```diff
@@ -0,0 +1,98 @@
...
+          String.format("enode://%s@%s:%d?discport=%d", nodeId, ip, listeningPort, discoveryPort));
+    } else {
+      return URI.create(String.format("enode://%s@%s:%d", nodeId, ip, listeningPort));
+    }
+  }
+
+  private Integer getAndValidatePort(final Matcher matcher, final String portName) {
+    int port = Integer.valueOf(matcher.group(portName));
+    if (port < 0 || port > 65535) {
+      throw new IllegalArgumentException(
```
macfarla: should we have consts for these magic numbers?

pantheon/src/test/java/tech/pegasys/pantheon/cli/PantheonCommandTest.java
```diff
@@ -431,13 +445,16 @@ public void callingWithBannedNodeidsOptionButNoValueMustDisplayErrorAndUsage() {
 
   @Test
   public void bootnodesOptionMustBeUsed() {
-    final String[] nodes = {"enode://001@123:4567", "enode://002@123:4567", "enode://003@123:4567"};
+    final String[] nodes = {
+      "enode://001@192.168.0.1:4567", "enode://002@192.168.0.1:4567", "enode://003@192.168.0.1:4567"
+    };
```
macfarla: aren't these also invalid because the 001 part is an odd number of characters - should be a hex string?

pantheon/src/test/java/tech/pegasys/pantheon/cli/custom/EnodeToURIPropertyConverterTest.java
```diff
@@ -0,0 +1,156 @@
...
+
+    assertThat(thrown)
+        .isInstanceOf(IllegalArgumentException.class)
+        .hasMessage(
+            "Invalid enode URL syntax. Enode URL should have the following format 'enode://<node_id>@<ip>:<listening_port>[?discport=<discovery_port>]'.");
+  }
+
+  @Test
+  public void convertEnodeURLWithoutListeningPortAnWithDiscoveryPort() {
+    String value = "enode://6f8a80d1@192.168.0.1:?30301";
```
macfarla: s/An/And

pantheon/src/test/java/tech/pegasys/pantheon/cli/custom/EnodeToURIPropertyConverterTest.java
```diff
@@ -0,0 +1,156 @@
...
+    String value = "enode://6f8a80d1@fe80::200:f8ff:fe21:67cf:30303?discport=30301";
+    URI expectedURI = URI.create(value);
+
+    URI convertedURI = converter.convert(value);
+
+    assertThat(convertedURI).isEqualTo(expectedURI);
+  }
+
+  @Test
+  public void convertEnodeURLWithoutNodeId() {
```
macfarla: should the test name include something about the expected result eg 'ShouldError' vs 'ShouldSucceed'

pantheon/src/main/java/tech/pegasys/pantheon/cli/custom/EnodeToURIPropertyConverter.java
lucassaldanha: My bad! I thought they were quite straightforward. I'll add a constant :)

lucassaldanha: Even better! We already have a utility for that :)

`NetworkUtility.isValidPort(port)`

PR #419, opened 2018-12-13
Make Pantheon behave as a submodule

shemnon:
## PR description

version sniffing should be submodule aware, we need to use a .git as a file to point to our actual submodule data.


build.gradle
```diff
@@ -411,6 +411,10 @@ def calculateVersion() {
 
 def getCheckedOutGitCommitHash() {
   def gitFolder = "$projectDir/.git/"
+  if (!file(gitFolder).isDirectory()) {
+    // we may be in a submodule
+    gitFolder = file(gitFolder).text.substring(8).trim() + "/"
```
mbaxter: Might be worth adding an explanatory comment here

shemnon: Done

PR #429, opened 2018-12-17
Allow IBFT Round to be created using PreparedCert

rain-on:
When the local node is to be the proposer for a new round
(due to round change), the new Ibft Round is to be created using the
latest received PreparedCertificate, and a NewRound message multicast
to all validators.

It should be noted, the block in the NewRound Proposal must be
modified from that received in the PreparedCertificate such that it
contains the executing round number (all other fields remain unchanged).


consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/IbftHelpers.java
```diff
@@ -51,4 +55,28 @@ public static Block createSealedBlock(
 
     return new Block(sealedHeader, block.getBody());
   }
+
+  public static Optional<PreparedCertificate> findLatestPreparedCertificate(
```
jframe: doesn't look there is a test for this

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/IbftHelpers.java
```diff
@@ -51,4 +55,28 @@ public static Block createSealedBlock(
...
+      final Collection<SignedData<RoundChangePayload>> msgs) {
+
+    Optional<PreparedCertificate> result = Optional.empty();
+
+    for (SignedData<RoundChangePayload> roundChangeMsg : msgs) {
+      final RoundChangePayload payload = roundChangeMsg.getPayload();
+      if (payload.getPreparedCertificate().isPresent()) {
+        if (!result.isPresent()) {
+          result = Optional.of(payload.getPreparedCertificate().get());
```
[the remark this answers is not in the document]
rain-on: done.

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/IbftHelpers.java
rain-on: Was previously fully tested as part of NewRoundMessageValidator (but was then moved out here) - have created additional tests based of the previously existing tests.

PR #431, opened 2018-12-17
ibft controller and future msgs handling

jframe:
## PR description
The IBFT controller handler will create a block height managers as needed and pass on message for the current height. For block at a future height it will buffer these until such a time that a new block event is received and then replay the messages.


consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/IbftController.java
```diff
@@ -0,0 +1,186 @@
...
+
+  public void startNewHeightManager(final BlockHeader parentHeader) {
+    currentHeightManager = ibftBlockHeightManagerFactory.create(parentHeader);
+    currentHeightManager.start();
+    final long newChainHeight = currentHeightManager.getChainHeight();
+    futureMessages.getOrDefault(newChainHeight, emptyList()).forEach(this::handleMessage);
+    futureMessages.remove(newChainHeight);
+  }
+
+  public void handleRoundExpiry(final RoundExpiry expired) {
```
rain-on: nit: move this up 1 function to be next to the other "handle" functions?

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/IbftController.java
```diff
@@ -0,0 +1,186 @@
...
+    currentHeightManager.roundExpired(expired);
```
rain-on: ChainHeight Check, discard if old?
maybe with logs if it's not for the current height?

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/IbftController.java
```diff
@@ -0,0 +1,186 @@
...
+      return isMsgForKnownValidator(msg);
+    } else if (isMsgForFutureChainHeight(msgRoundIdentifier)) {
+      addMessageToFutureMessageBuffer(msgRoundIdentifier.getSequenceNumber(), rawMsg);
+    }
+    return false;
+  }
+
+  private boolean processRoundMessage(final SignedData<? extends Payload> msg) {
+    final ConsensusRoundIdentifier msgRoundIdentifier = msg.getPayload().getRoundIdentifier();
+    return isMsgForCurrentOrFutureChainHeight(msgRoundIdentifier) && isMsgForKnownValidator(msg);
```
rain-on: RoundChange and NewRound Messages are still targetted to current Chain Height, and mus be handled as future messages the same way as Prepare/Commit/Proposal messages - so this function is probably not needed, you can just route RoundChange and NewRound through the standard processMessage function.

jframe: Yeah your right. Don't know why it did that differently. I'll just change them to use the processMessage method.

jframe: done

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/IbftController.java
jframe: I think logging is good idea. I'll add some

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/IbftController.java
```diff
@@ -0,0 +1,186 @@
...
+  public void handleBlockTimerExpiry(final BlockTimerExpiry blockTimerExpiry) {
+    currentHeightManager.handleBlockTimerExpiry(blockTimerExpiry.getRoundIndentifier());
+  }
+
+  public void startNewHeightManager(final BlockHeader parentHeader) {
+    currentHeightManager = ibftBlockHeightManagerFactory.create(parentHeader);
+    currentHeightManager.start();
+    final long newChainHeight = currentHeightManager.getChainHeight();
+    futureMessages.getOrDefault(newChainHeight, emptyList()).forEach(this::handleMessage);
+    futureMessages.remove(newChainHeight);
```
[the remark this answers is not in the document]
jframe: Your thinking we should just remove all future messages below new chain height?

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/IbftController.java
```diff
@@ -0,0 +1,190 @@
...
+
+  public void handleRoundExpiry(final RoundExpiry roundExpiry) {
+    if (isMsgForCurrentHeight(roundExpiry.getView())) {
+      currentHeightManager.roundExpired(roundExpiry);
+    } else {
+      LOG.info("Round expiry event discarded as it is not for current block height");
+    }
+  }
+
+  public void startNewHeightManager(final BlockHeader parentHeader) {
```
rain-on: can this be private?

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/IbftController.java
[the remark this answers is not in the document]
rain-on: Change of mind - i think we've done the right thing - i.e. do the exact right thing, not a broad brush catch all.

PR #438, opened 2018-12-18
Insert interfaces for MessageValidator

rain-on:
To facilitate testing/stubbing, an interface has been added "under" the MessageValdiator.


rain-on:
Not going to happen now.

PR #474, opened 2018-12-20
Set base directory name to be lowercase in building.md

Matthalp:
## PR description
Keeping the base directory lower case is a common convention amongst repositories and makes the local repository directory default to all lower case during a default checkout.


Matthalp:
Thanks @rojotek !

PR #475, opened 2018-12-20
Propagate ibft newBlock messages after block import

jframe:
## PR description
Propagate ibft new block messages after block import. This adds the EthProtocolManager as a block mined observer so that it can notify peers with the new block message we import a block.


consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/IbftRoundFactory.java
```diff
@@ -24,22 +24,22 @@
...
 
 public class IbftRoundFactory {
-
   private final IbftFinalState finalState;
-
   private final IbftBlockCreatorFactory blockCreatorFactory;
   private final ProtocolContext<IbftContext> protocolContext;
   private final ProtocolSchedule<IbftContext> protocolSchedule;
-  private final Subscribers<MinedBlockObserver> minedBlockObservers = new Subscribers<>();
+  private final Subscribers<MinedBlockObserver> minedBlockObservers;
```
CjHare: With `Subscribers` being a mutable object passed  by reference to the constructor, you are allowing that object to be altered (the subscriptions to be added / removed) after the time of the factory construction. Is this the intention?

jframe: It definitely wasn't the intention, it doesn't need to be mutable. The IbftRound just needs to iterate the list of subscribers.

PR #476, opened 2018-12-20
Nodes whitelist JSON-RPC APIs

mark-terry:
## PR description
Added JSON-RPC methods to interact with the permissioning whitelist.

## Fixed Issue(s)

#NC-1938


ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/eea/PermAddNode.java
```diff
@@ -0,0 +1,69 @@
...
+      return new JsonRpcSuccessResponse(req.getId(), true);
+    } catch (IllegalArgumentException e) {
+      return new JsonRpcErrorResponse(req.getId(), JsonRpcError.INVALID_PARAMS);
+    } catch (Exception e) {
+      return new JsonRpcErrorResponse(req.getId(), JsonRpcError.INTERNAL_ERROR);
+    }
+  }
+
+  private DefaultPeer parsePeers(final String enodeURI) {
+    return DefaultPeer.fromURI(enodeURI);
```
macfarla: should this method be singular ie parsePeer ?

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/eea/PermRemoveNode.java
```diff
@@ -0,0 +1,69 @@
...
+      return new JsonRpcSuccessResponse(req.getId(), true);
+    } catch (IllegalArgumentException e) {
+      return new JsonRpcErrorResponse(req.getId(), JsonRpcError.INVALID_PARAMS);
+    } catch (Exception e) {
+      return new JsonRpcErrorResponse(req.getId(), JsonRpcError.INTERNAL_ERROR);
+    }
+  }
+
+  private DefaultPeer parsePeers(final String enodeURI) {
+    return DefaultPeer.fromURI(enodeURI);
```
macfarla: singular ie parsePeer

ethereum/jsonrpc/src/test/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/eea/PermAddNodeTest.java
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  @Test
+  public void shouldReturnCorrectMethodName() {
+    assertThat(method.getName()).isEqualTo(METHOD_NAME);
+  }
+
+  @Test
+  public void shouldThrowInvalidJsonRpcParametersExceptionWhenBadEnode() {
+    final JsonRpcRequest request = buildRequest(badEnode);
```
macfarla: should we also test when the bad one is in a list with good ones?

ethereum/jsonrpc/src/test/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/eea/PermRemoveNodeTest.java
```diff
@@ -0,0 +1,131 @@
...
+  public void shouldThrowExceptionWhenEnodeNotInWhitelist() {
+    final JsonRpcRequest request = buildRequest("");
+    final JsonRpcResponse expected =
+        new JsonRpcErrorResponse(request.getId(), JsonRpcError.INVALID_PARAMS);
+
+    final JsonRpcResponse actual = method.response(request);
+
+    assertThat(actual).isEqualToComparingFieldByFieldRecursively(expected);
+  }
+
```
macfarla: this test looks the same as shouldThrowInvalidJsonRpcParametersExceptionWhenNoEnode - missing buildRequest() param value?

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/AcceptanceTestBase.java
```diff
@@ -53,6 +56,7 @@ protected AcceptanceTestBase() {
     cliqueTransactions = new CliqueTransactions();
     clique = new CliqueConditions(ethTransactions, cliqueTransactions);
     net = new Net(new NetTransactions());
+    perm = new Perm(new PermTransaction());
```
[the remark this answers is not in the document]
mark-terry: Ah, I see.  The base class is plural, while the specific Transaction types are singular.  Will update. 👍

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/factory/PantheonNodeFactory.java
```diff
@@ -137,6 +138,27 @@ public PantheonNode createNodeWithNodesWhitelist(
             .build());
   }
 
+  public PantheonNode createNodeWithNodesWhitelistAndPermRPC(
```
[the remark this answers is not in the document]
mark-terry: That is something I was going to raise for discussion.  Will refactor.  👍

ethereum/jsonrpc/src/test/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/eea/PermRemoveNodeTest.java
mark-terry: Actually, I tested this logic in NodeWhitelistControllerTest.  I'll delete this test.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/permissioning/PermRemoveNodesFromWhitelist.java
```diff
@@ -0,0 +1,70 @@
...
+              .getStringList()
+              .parallelStream()
+              .map(this::parsePeer)
+              .collect(Collectors.toList());
+
+      p2pNetwork.getNodeWhitelistController().removeNodes(peers);
+      return new JsonRpcSuccessResponse(req.getId(), true);
+    } catch (IllegalArgumentException e) {
+      return new JsonRpcErrorResponse(req.getId(), JsonRpcError.NODE_WHITELIST_INVALID_ENTRY);
+    } catch (Exception e) {
```
lucassaldanha: I think capturing a generic exception and assuming it was caused by a missing entry can be misleading. If it is a server failure or something else throwing the exception we will never know the true cause.

It would be good to either create specific exceptions (what we don't do much in Pantheon) or create an object that represents the status of the `removeNodes` operation. And based on the status you choose the correct error response. This is the approach I took for the Accounts whitelist.

lucassaldanha: I think capturing a generic exception and assuming it was caused by a missing entry can be misleading. If it is a server failure or something else throwing the exception we will never know the true cause.

It would be good to either create specific exceptions (what we don't do much in Pantheon) or create an object that represents the status of the removeNodes operation. And based on the status you choose the correct error response. This is the approach I took for the Accounts whitelist.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/permissioning/PermAddNodesToWhitelist.java
```diff
@@ -0,0 +1,69 @@
...
+              .getStringList()
+              .parallelStream()
+              .map(this::parsePeer)
+              .collect(Collectors.toList());
+
+      p2pNetwork.getNodeWhitelistController().addNodes(peers);
+      return new JsonRpcSuccessResponse(req.getId(), true);
+    } catch (IllegalArgumentException e) {
+      return new JsonRpcErrorResponse(req.getId(), JsonRpcError.NODE_WHITELIST_INVALID_ENTRY);
+    } catch (Exception e) {
```
lucassaldanha: Same comment about the generic exception catch here.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/condition/perm/GetNodesWhitelistPopulated.java
```diff
@@ -0,0 +1,37 @@
...
+
+  public GetNodesWhitelistPopulated(final PermGetNodesWhitelistTransaction transactions) {
+    this.transaction = transactions;
+  }
+
+  @Override
+  public void verify(final Node node) {
+    final List<String> response = node.execute(transaction);
+    assertThat(response).isInstanceOf(List.class);
+    assertThat(response).size().isEqualTo(3);
```
lucassaldanha: Instead of having this hardcoded to 3 we should probably add a parameter to this condition with the expected list and use the size here.

Example: `AwaitNetPeerCount.java`

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/permissioning/PermRemoveNodesFromWhitelist.java
mark-terry: Good idea.  I'll implement a return object to make this a bit clearer and enable the generic handler to catch unexpected exceptions.  👍

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/permissioning/PermGetNodesWhitelist.java
```diff
@@ -0,0 +1,72 @@
...
+      List<String> enodeList =
+          nodesWhitelist.parallelStream().map(this::buildEnodeURI).collect(Collectors.toList());
+
+      return new JsonRpcSuccessResponse(req.getId(), enodeList);
+    } else {
+      return new JsonRpcErrorResponse(req.getId(), JsonRpcError.NODE_WHITELIST_NOT_SET);
+    }
+  }
+
+  private String buildEnodeURI(final Peer s) {
```
Errorific: would this be better moved into Peer? e.g. peer.getEnodeURI()

PR #483, opened 2018-12-21
ibft mining acceptance test

jframe:
## PR description
Acceptance test for IBFT to test that a single node and a cluster of 4 nodes work in the basic mining test sending transactions.


acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/cluster/Cluster.java
```diff
@@ -88,33 +89,85 @@ public void start(final List<RunnableNode> nodes) {
...
+        .filter(nodeInfo -> nodeInfo.isRunning)
+        .map(NodeInfo::getRunnableNode)
+        .collect(toList());
+  }
+
+  public RunnableNode getFirstRunningNode() {
+    return getRunningNodes().get(0);
+  }
+
+  private static class NodeInfo {
```
CjHare: What's the value in this new inner class, my IntelliJ shows `isRunning()` and `isRunning` are not called outside NodeInfo.

jframe: It's just way a way adding of associating a the running state with the nodes and their names. Allows me to just verify on the running nodes if I have stopped some.

jframe: True the isRunning() doesn't need to exist, I just added both the getter and setters for both fields for consistency.

CjHare: This approach with NodeInfo and that with the `verifyRunningNode` are inconsistent with the indirection used elsewhere within Cluster and AT framework.

Let me think about it, there'll be a more consistent way to this goal.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/factory/PantheonNodeFactory.java
```diff
@@ -187,25 +189,16 @@ public PantheonNode createCliqueNode(final String name) throws IOException {
...
-      return Resources.toString(uri.toURL(), Charset.defaultCharset());
-    } catch (URISyntaxException | IOException e) {
-      throw new IllegalStateException("Unable to get test clique genesis config");
-    }
+  public PantheonNode createIbftNode(final String name) throws IOException {
+    return create(
+        new PantheonFactoryConfigurationBuilder()
+            .setName(name)
+            .miningEnabled()
+            .setJsonRpcConfiguration(jsonRpcConfigWithClique())
```
rain-on: jsonRpcConfigWithClique? (not withIbft?)

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/ibft/IbftMiningAcceptanceTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+    minerNode.execute(transactions.createIncrementalTransfers(sender, receiver, 1));
+    cluster.verify(receiver.balanceEquals(1));
+
+    minerNode.execute(transactions.createIncrementalTransfers(sender, receiver, 2));
+    cluster.verify(receiver.balanceEquals(3));
+  }
+
+  @Test
+  public void shouldMineOnMultipleNodes() throws IOException {
```
rain-on: can we create a test whereby one of the nodes is not a validator?

acceptance-tests/src/test/resources/ibft/ibft.json
```diff
@@ -0,0 +1,43 @@
...
+  "config": {
+    "chainId": 4,
+    "homesteadBlock": 1,
+    "eip150Block": 2,
+    "eip150Hash": "0x0000000000000000000000000000000000000000000000000000000000000000",
+    "eip155Block": 3,
+    "eip158Block": 3,
+    "byzantiumBlock": 1035301,
+    "revisedibft": {
+      "blockperiodseconds": 5,
```
rain-on: do we want to drop these values to make tests faster? (blocktime could be 1 reasonably).

If we do timeout tests, we probably don't want to wait 25 seconds for (request) timeout ....

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/factory/PantheonNodeFactory.java
jframe: oh that's not right. i'll fix that

jframe: done. it happen to be working because it is enabled by default. I still prefer to be explicit in this case though so am doing it the same as in the clique case and adding the rpc option.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/ibft/IbftMiningAcceptanceTest.java
jframe: added a test, but have disabled as it won't work until the new block events are being propagated to peers. the PR for propagating the new_block event is up for review so will re-enable this test once that is merged.

acceptance-tests/src/test/resources/ibft/ibft.json
jframe: not sure what reasonable time is. in the clique tests I found they were not as reliable when I had the block time at such a small value. will have to think about this a bit. but in general want this to be as small as practical.
agree 25 seconds is too long. ideas on how long? I might go with 5s for now.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/ibft/IbftMiningAcceptanceTest.java
```diff
@@ -43,28 +44,52 @@ public void shouldMineOnSingleNode() throws IOException {
...
   public void shouldMineOnMultipleNodes() throws IOException {
-    final PantheonNode minerNode1 = pantheon.createIbftNode("miner1");
-    final PantheonNode minerNode2 = pantheon.createIbftNode("miner2");
-    final PantheonNode minerNode3 = pantheon.createIbftNode("miner3");
-    final PantheonNode minerNode4 = pantheon.createIbftNode("miner4");
-    cluster.start(minerNode1, minerNode2, minerNode3, minerNode4);
+    final PantheonNode validator1 = pantheon.createIbftNode("miner1");
+    final PantheonNode validator2 = pantheon.createIbftNode("miner2");
+    final PantheonNode validator3 = pantheon.createIbftNode("miner3");
+    final PantheonNode nonValidatorNode = pantheon.createIbftNode("miner4");
```
rain-on: textual name (miner4) doesn't line up with variable name (nonValidatorNode) - which should it be?

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/ibft/IbftMiningAcceptanceTest.java
```diff
@@ -43,28 +44,52 @@ public void shouldMineOnSingleNode() throws IOException {
...
-    minerNode4.execute(transactions.createIncrementalTransfers(sender, receiver, 3));
+    nonValidatorNode.execute(transactions.createIncrementalTransfers(sender, receiver, 3));
     cluster.verify(receiver.balanceEquals(6));
   }
 
+  @Test
+  @Ignore("Temporarily disabled until ibft new block events are being sent")
+  public void shouldMineOnMultipleNodesEvenWhenClusterContainsNonValidator() throws IOException {
+    final String[] validators = {"validator1", "validator2", "validator3"};
+    final PantheonNode validator1 = pantheon.createIbftNodeWithValidators("validator1", validators);
```
rain-on: name field could/should reference the validators array? [or even put this into a loop?]

jframe: rest of the code can't reference the validators array, best I could do is use the validator names in that array to create the validator nodes. I've done that but I think the test is less clear afterwards.

can't use stream as we have a checked exception (IOException) to deal with from the createIbft methods.

```
final String[] validatorNames = {"validator1", "validator2", "validator3"};
    final List<PantheonNode> validators = new ArrayList<>();
    for (String name: validatorNames) {
      validators.add(pantheon.createIbftNodeWithValidators(name, validatorNames));
    }
    final PantheonNode nonValidatorNode =
        pantheon.createIbftNodeWithValidators("non-validator", validatorNames);
    validators.add(nonValidatorNode);
    cluster.start(validators);

    final Account sender = accounts.createAccount("account1");
    final Account receiver = accounts.createAccount("account2");

    validators.get(0).execute(transactions.createTransfer(sender, 50));
    cluster.verify(sender.balanceEquals(50));

    validators.get(1).execute(transactions.createIncrementalTransfers(sender, receiver, 1));
    cluster.verify(receiver.balanceEquals(1));

    nonValidatorNode.execute(transactions.createIncrementalTransfers(sender, receiver, 2));
    cluster.verify(receiver.balanceEquals(3));
```

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/cluster/Cluster.java
jframe: done

acceptance-tests/src/test/resources/ibft/ibft.json
jframe: done

PR #491, opened 2018-12-30
Review rework on filter methods

### LINKED ISSUE #424: Update filter methods to indicate they return filter ID hashes
### Description
Filter JSON-RPC API methods (`eth_newFilter`) etc return non-sequential transaction hashes not sequential decimal IDs.

MadelineMurray:
## Fixed Issue(s)
fixes #424


docs/Reference/JSON-RPC-API-Methods.md
```diff
@@ -1102,19 +1102,19 @@ Returns the receipt of a transaction by transaction hash. Receipts for pending t
 
 ### eth_newFilter
 
-Creates a logs filter to log state changes. To poll for state changes, use [eth_getFilterChanges](#eth_getfilterchanges).
+Creates a log filter. To poll for logs, use [eth_getFilterChanges](#eth_getfilterchanges).
```
NicolasMassart: Maybe something more explicit like “To poll for logs associated with this filter” would help understand that the method is not just to retrieve logs, but logs that are returned by this specific filter. It was not super clear to me at first read so...

MadelineMurray: Updated

PR #498, opened 2019-01-02
Added documentation on default APIs and enabling non-default APIs

### LINKED ISSUE #464: Document that admin_peers included in admin API group
### Description
As a user, I want documentation so that I know which API group to enable to use admin_peers. 

### Acceptance Criteria
* Documented that admin_peers is part of admin API 
* Documentation tested

MadelineMurray:
## Fixed Issue(s)
fixes #464


docs/Reference/Using-JSON-RPC-API.md
```diff
@@ -70,6 +70,15 @@ Send individual requests as a JSON data package at each prompt:
 
 The [RPC Pub/Sub methods](../Using-Pantheon/RPC-PubSub.md) can also be used over WebSockets.
 
+### APIs Enabled by Default
+
+The `ETH`, `NET`, `WEB3`, `CLIQUE`, and `IBFT` APIs are enabled by default.
```
lucassaldanha: Are we sure these are the default APIs? In `RpcApis` class like 27 we have:

```
public static final Collection<RpcApi> DEFAULT_JSON_RPC_APIS = Arrays.asList(ETH, NET, WEB3);
```

MadelineMurray: Turns out they're also being set on the command line. I've raised a JI for this and am going to merge this as it documents the current behaviour. Note included in JI that a docs issue is required when that is addressed.

MadelineMurray: It turns out the default APIs were being specified in two places. Merging this doc as it matches the current behaviour. Have raised a JI to fix and noted that a docs JI is required when it's addressed.

PR #499, opened 2019-01-02
Initial implementation of ibft_getValidatorsByBlockNumber

rojotek:
Implements the ibft_getValidatorsByBlockNumber method. (really it is byBlockNumber not hash -- I implemented the BlockNumber version not the blockhash bersion :))


consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/jsonrpc/IbftJsonRpcMethodsFactory.java
```diff
@@ -34,10 +37,17 @@
     final Map<String, JsonRpcMethod> rpcMethods = new HashMap<>();
 
     if (jsonRpcApis.contains(IbftRpcApis.IBFT)) {
+      BlockchainQueries blockchainQueries =
+          new BlockchainQueries(context.getBlockchain(), context.getWorldStateArchive());
```
rain-on: nit: might be _slightly_ nicer to store context.X() things in local variables and use them from there ... but its 50/50.

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/jsonrpc/methods/IbftGetValidatorsByBlockNumber.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.Optional;
+
+public class IbftGetValidatorsByBlockNumber extends AbstractBlockParameterMethod
+    implements JsonRpcMethod {
+
+  private final Blockchain blockchain;
+  private final IbftBlockInterface ibftBlockInterface;
+
+  public IbftGetValidatorsByBlockNumber(
+      final BlockchainQueries blockchainQueries,
```
rain-on: Having both BlockChainQueries AND Blockchain is a smell - i.e. I'm assuming the queries should hide the underlying blockchain.

I _suspect_ the smell is that AbstractBlockParameterMethod should actually be 2 things:
1. AbstractJsonParameterHandler
2.BlockchainQuerier

That way, this class can re-use the AbstractJsonParameterHandler to do the json bits, but then wouldn't need both a BlockchainQuery AND Blockchain...

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/jsonrpc/methods/IbftGetValidatorsByBlockNumber.java
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  @Override
+  protected BlockParameter blockParameter(final JsonRpcRequest request) {
+    return parameters().required(request.getParams(), 0, BlockParameter.class);
+  }
+
+  @Override
+  protected Object resultByBlockNumber(final JsonRpcRequest request, final long blockNumber) {
+    Optional<BlockHeader> blockHeader = blockchainQueries().getBlockHeaderByNumber(blockNumber);
```
rain-on: Having chewed the fat with AJS, we wonder if its better _not_ to sully BlockChainQueries in this PR, and just reuse BlockChainQueries::blockByNumber (its slightly less efficient, but means less created clutter).

PR #505, opened 2019-01-03
Added --accounts-whitelist

### LINKED ISSUE #495: Document --accounts-whitelisting
### Description
As a user, I want to know how to use the `--accounts-whitelisting` option so I can implement permissioning for transactions. 

### Acceptance Criteria
* `--accounts-whitelisting` added to CLI reference 
* Conceptual information on using option included in Permissioning Overview
* Documentation tested

MadelineMurray:
## Fixed Issue(s)
fixes #495


docs/Reference/Pantheon-CLI-Syntax.md
```diff
@@ -13,6 +13,18 @@ Runs the Pantheon Ethereum full node client.
 
 ## Options
 
+### accounts-whitelist
+
+```bash tab="Syntax"
+--accounts-whitelist[=<hex string of account public key>[,<hex string of account public key>...]...
```
lucassaldanha: I'm not sure if the cli command needs the `[` and `]`. At least it didn't before I went on holiday :)

MadelineMurray: It doesn't need it - it's to indicate that part is optional (ie, you can specify an empty list). Also, it matches what's in CLI help and is consistent with the other commands where an empty list can be specified :-) Will follow up with @jakehaugen and @NicolasMassart to confirm for the CLI changes.

NicolasMassart: Yes [] indicates optional values, no need to use them in the actual CLI value

PR #511, opened 2019-01-04
Metrics measurement adjustment

shemnon:
## PR description

The total difficulty of mainnet is larger than a signed long, so we
need to adjust the way we get the number.


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/db/DefaultMutableBlockchain.java
```diff
@@ -69,7 +70,9 @@ public DefaultMutableBlockchain(
         MetricCategory.BLOCKCHAIN,
         "difficulty_total",
         "Total difficulty of the chainhead",
-        () -> (double) this.getChainHead().getTotalDifficulty().toLong());
+        () ->
+            new BigInteger(this.getChainHead().getTotalDifficulty().getBytes().getArrayUnsafe())
```
mbaxter: There's a utility for transforming `BytesValues` to `BigInteger`'s: `BytesValues.asUnsignedBigInteger(this.getChainHead().getTotalDifficulty().getBytes()).doubleValue()`

shemnon: I was expecting it on UInt256, which is why I stopped looking and rolled my own.

PR #523, opened 2019-01-07
NewRoundMessageValidator ignores Round Number when comparing blocks

rain-on:
NewRoundMessageValidator previously ensured the block in the newest
PreparedCertificate was the same as that in the Proposal by comparing
block hashes.
However, the hash-function for received IBFT blocks includes round
number - thus the blocks will never be deemed 'equal'.

As such, the blocks must be compared using a hash function which
ignores the round number - i.e. the OnChain function.


consensus/ibft/src/test/java/tech/pegasys/pantheon/consensus/ibft/validation/NewRoundMessageValidatorTest.java
```diff
@@ -289,4 +268,25 @@ public void lastestPreparedCertificateMatchesNewRoundProposalIsSuccessful() {
 
     assertThat(validator.validateNewRoundMessage(msg)).isTrue();
```
CjHare: Based on the name for the this test and the added comments referring to failing validation, should this assertion be a `isFalse()` rather than `isTrue()`?

rain-on: It looks like this test is proving the successful case when we have multiple PrepareCertificates, so ATM, I think its correct.

consensus/ibft/src/test/java/tech/pegasys/pantheon/consensus/ibft/validation/NewRoundMessageValidatorTest.java
```diff
@@ -147,16 +145,6 @@ public void newRoundTargetingDifferentSequenceNumberFails() {
     assertThat(validator.validateNewRoundMessage(inValidMsg)).isFalse();
   }
 
-  @Test
```
jframe: isn't this test still valid?

consensus/ibft/src/test/java/tech/pegasys/pantheon/consensus/ibft/validation/NewRoundMessageValidatorTest.java
CjHare: My bad, in GitHub interface I'd assumed choosing to expand the collapsed section fully expanded ...turns out multiple clicks were needed, the `isTrue()` is in the succeed test, not the fail test.

consensus/ibft/src/test/java/tech/pegasys/pantheon/consensus/ibft/validation/NewRoundMessageValidatorTest.java
rain-on: The mocked messageValidatorFactory (and MessageValidator) ultimately means it doesn't matter WHY a message is invalid - just that it is - thus many of the specialised failure modes can be removed.

PR #526, opened 2019-01-08
Updated --nodes-whitelist

### LINKED ISSUE #433: Doc #414 If --nodes-whitelist specified bootnodes must be subset of whitelist nodes
# Requirements
Doc:
• if you specify a network like `--rinkeby` AND specify `--nodes-whitelist` then you need to include the rinkeby bootnodes in the whitelist.
• if you specify no bootnodes, you get mainnet bootnodes - which similarly need to be in the `--nodes-whitelist`
• this means that you can't specify an empty `--nodes-whitelist` (because you can't specify an empty list of bootnodes). You could achieve an equivalent setup by adding a dummy node (that you will never connect to) to the `bootnodes` and `--nodes-whitelist`

MadelineMurray:
Updated `--nodes-whitelist` 

## Fixed Issue(s)
fixes #433


rojotek:
Engineering have approved.  I suspect this one can be updated and merged.

PR #527, opened 2019-01-08
Ibft Int Testing - Local Node is proposer

rain-on:


consensus/ibft/src/integration-test/java/tech/pegasys/pantheon/consensus/ibft/tests/LocalNodeIsProposerTest.java
```diff
@@ -0,0 +1,116 @@
...
+import tech.pegasys.pantheon.crypto.SECP256K1.Signature;
+import tech.pegasys.pantheon.ethereum.core.Block;
+
+import java.time.Clock;
+
+import org.junit.Before;
+import org.junit.Test;
+
+public class LocalNodeIsProposerTest {
+  // These tests assume the basic function of the Ibft Round State Machine has been proven
```
CjHare: Is there a reason this comment cannot be a class comment? (as it seems an appropriate level)

consensus/ibft/src/integration-test/java/tech/pegasys/pantheon/consensus/ibft/tests/LocalNodeIsProposerTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+  private final MessageFactory localNodeMessageFactory = context.getLocalNodeMessageFactory();
+
+  private Block expectedProposedBlock;
+  private SignedData<ProposalPayload> expectedTxProposal;
+  private SignedData<CommitPayload> expectedTxCommit;
+
+  @Before
+  public void setup() {
+    when(mockBlockCreationTime.millis()).thenReturn(blockTimeStamp * 1000);
```
CjHare: `blockTimeStamp * 1000` would be the 100 seconds ...what's the significance of that number?

consensus/ibft/src/integration-test/java/tech/pegasys/pantheon/consensus/ibft/tests/LocalNodeIsProposerTest.java
```diff
@@ -0,0 +1,116 @@
...
+    context.getController().start();
+    context.getController().handleBlockTimerExpiry(new BlockTimerExpiry(roundId));
+  }
+
+  @Test
+  public void basicCase() {
+    assertPeersReceivedExactly(roles.getAllPeers(), expectedTxProposal);
+
+    // NOTE: In these test roles.getProposer() will return NULL.
+    roles.getNonProposingPeers().get(0).injectPrepare(roundId, expectedProposedBlock.getHash());
```
CjHare: nit; for a readability gain, if you changed the signature of `getNonProposingPeers()`to accept an integer then `getNonProposingPeers().get(0)` could become `getNonProposingPeer(0)`.

consensus/ibft/src/integration-test/java/tech/pegasys/pantheon/consensus/ibft/tests/LocalNodeIsProposerTest.java
rain-on: Done.

consensus/ibft/src/integration-test/java/tech/pegasys/pantheon/consensus/ibft/tests/LocalNodeIsProposerTest.java
rain-on: So - as the local node is the proposer, it will create a block, with the timestamp as reported by the Clock (which is mocked in these tests).

In order to validate packets, the timestamp in the created (and transmitted) block must known - thus the _expectedBlock_ has 'blockTimestamp' in it - and the clock which will provide a timestamp to the UUT is set to the same time.

consensus/ibft/src/integration-test/java/tech/pegasys/pantheon/consensus/ibft/tests/LocalNodeIsProposerTest.java
rain-on: yup - much nicer solution.

consensus/ibft/src/integration-test/java/tech/pegasys/pantheon/consensus/ibft/tests/LocalNodeIsProposerTest.java
jframe: Could use Clock.fixed to created a clock for testing

CjHare: Right, so it's just about covering the `blocTimeStamp` to seconds ...because in that case you could use`TimeUnit.MILLISECONDS.toSeconds(blocTimeStamp)` and loose the magic number :)

consensus/ibft/src/integration-test/java/tech/pegasys/pantheon/consensus/ibft/tests/LocalNodeIsProposerTest.java
```diff
@@ -0,0 +1,119 @@
...
+  private SignedData<CommitPayload> expectedTxCommit;
+
+  @Before
+  public void setup() {
+    when(mockBlockCreationTime.millis()).thenReturn(blockTimeStamp * 1000);
+    expectedProposedBlock = context.createBlockForProposal(0, blockTimeStamp);
+    expectedTxProposal =
+        localNodeMessageFactory.createSignedProposalPayload(roundId, expectedProposedBlock);
+
+    final IbftExtraData extraData =
```
jframe: perhaps this could be moved into a common test util somewhere we are probably going to do this in many of the nodes. or perhaps just inside the createSignedCommitPayload using the proposedBlock as an arg to it?

consensus/ibft/src/integration-test/java/tech/pegasys/pantheon/consensus/ibft/tests/LocalNodeIsProposerTest.java
rain-on: Changed to clock.fixed, nicer than using Mocks if we don't need to.

PR #531, opened 2019-01-08
RocksDB Metrics

shemnon:
## PR description

* Measure read/get, write/put, remove, and commit latencies
* Measure rollback count (latency seemed a bit much)


services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/RocksDbKeyValueStorage.java
```diff
@@ -44,19 +48,45 @@
...
+          metricsSystem.createTimer(
+              MetricCategory.ROCKSDB,
+              "remove_latency_seconds",
+              "Latency of remove requests from RocksDB.");
+      writeLatency =
+          metricsSystem.createTimer(
+              MetricCategory.ROCKSDB, "write_latency_seconds", "Latency for write to RocksDB.");
+      commitLatency =
+          metricsSystem.createTimer(
+              MetricCategory.ROCKSDB, "commit_latency_seconds", "Latency for write to RocksDB.");
```
ajsutton: Description is incorrect, should be "for commit" not "for write"

services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/RocksDbKeyValueStorage.java
```diff
@@ -44,19 +48,45 @@
...
+
+      rollbackCount =
+          metricsSystem.createCounter(
+              MetricCategory.ROCKSDB, "rollback_count", "Number of transactions rolled back.");
```
ajsutton: Can we make the description "Number of RocksDB transactions rolled back."?  I know it will have rocksdb in the key as well but it would be easy to confuse this with number of EVM transactions rolled back.

services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/RocksDbKeyValueStorage.java
shemnon: Done

services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/RocksDbKeyValueStorage.java
shemnon: Done

PR #538, opened 2019-01-10
[PRIV] Prepare private transaction support

iikirilov:
## PR description

Adds an Orion client with integration tests.

The purpose of this PR is to prepare for supporting private transaction in Pantheon. To support private transaction it is required to access an enclave from Pantheon. As Orion is an enclave created by the same organisation that maintains Pantheon it seems natural to support the Orion enclave first.

No functionality is added to Pantheon in this PR.


build.gradle
```diff
@@ -83,7 +83,9 @@ allprojects {
     } else {
       jcenter()
       mavenCentral()
-      maven { url  "https://consensys.bintray.com/pegasys-repo" }
+      maven { url "https://consensys.bintray.com/pegasys-repo" }
+      maven { url 'https://jitpack.io' }
```
macfarla: We don't want to add jitpack as a dependency resolver. We're working on improving the release process for Orion which should remove the need for this (you'll be able to get the snapshot from bintray)

macfarla: or maybe the 0.1.0-SNAPSHOT in bintray has what you need?

macfarla: https://bintray.com/consensys/binaries/orion

macfarla: (also this is the reason the build fails on jenkins, because jitpack isn't configured)

antonydenyer: Agreed, will switch over to `0.1.0-SNAPSHOT`. No jitpack should 'just work' but it's not a problem if we move over to the snapshot

antonydenyer:
> Do we have to include libraries binaries in the repository ?

Yes but no! Until https://github.com/PegaSysEng/orion/issues/197 is fixed we need some kind of binary. When that's done we can switch over to using the snapshot. 

The alternative is we ignore Orion integration and acceptance tests on CI and force developers to be running Orion locally.

rojotek:
For today I am ok with the idea of us having the libs included.  As soon as Orion ships with a pom file, it will be possible to delete these libs.  Until we do that, we need to support the approach of including the files directly, in order to unblock the work in this area of the codebase.

PR #539, opened 2019-01-10
[NC-2056] Remove vertx from discovery tests

mbaxter:
## PR description
Previously, our peer discovery tests used vertx to set up real sockets to communicate between discovery agents within the tests.  This was a source of non-determinacy causing our tests to randomly fail.  This PR reorganizes the discovery code in order to isolate vertx.

Changes include:
- Created an abstract PeerDiscoveryAgent to hold code not related to vertx.  A subclass now holds the vertx-specific code.
- Create an abstract interface to wrap around vertx timer utilities
- Create some test utilities to properly mock out network communication without setting up real sockets
- Remove circular dependencies between PeerDiscoveryAgent and PeerDiscoveryController
- Remove dead code
- Fix bug where we attempted to bond with ourselves

## Fixed Issue(s)
NC-2056


ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/VertxPeerDiscoveryAgent.java
```diff
@@ -0,0 +1,193 @@
...
+
+  @Override
+  protected CompletableFuture<InetSocketAddress> listenForConnections() {
+    CompletableFuture<InetSocketAddress> future = new CompletableFuture<>();
+    vertx
+        .createDatagramSocket(new DatagramSocketOptions().setIpV6(NetworkUtility.isIPv6Available()))
+        .listen(
+            config.getBindPort(),
+            config.getBindHost(),
+            res -> {
```
shemnon: Per style guild lambdas should be 1-3 lines.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
```diff
@@ -105,7 +103,9 @@
 
   private final AtomicBoolean started = new AtomicBoolean(false);
 
-  private final PeerDiscoveryAgent agent;
+  private final SECP256K1.KeyPair keypair;
+  final DiscoveryPeer self;
```
shemnon: Not a fan of `self` outside python (nor of `this` inside of python).  Could this become `discoveryPeer` or `localPeer` or something else?  

My concern is that it maps to the same concept and so I keep reading it as `this`.

ethereum/p2p/src/test/java/tech/pegasys/pantheon/ethereum/p2p/discovery/PeerDiscoveryTestHelper.java
```diff
@@ -13,27 +13,207 @@
...
+    private boolean active = true;
+
+    public AgentBuilder(
+        final Map<BytesValue, MockPeerDiscoveryAgent> agents,
+        final AtomicInteger nextAvailablePort) {
+      this.agents = agents;
+      this.nextAvailablePort = nextAvailablePort;
+    }
+
+    public AgentBuilder setBootstrapPeers(final List<DiscoveryPeer> peers) {
```
shemnon: When I restored idiomatic getters/setters I didn't do setters for Builders (stuff that returned `this`).  For consistency could you change it to just <foo>?

ethereum/p2p/src/test/java/tech/pegasys/pantheon/ethereum/p2p/discovery/PeerDiscoveryTimestampsTest.java
```diff
@@ -26,126 +28,107 @@
...
-    final SECP256K1.KeyPair[] keypairs = PeerDiscoveryTestHelper.generateKeyPairs(2);
-    final DiscoveryPeer[] peers = PeerDiscoveryTestHelper.generateDiscoveryPeers(keypairs);
+    final List<KeyPair> keypairs = PeerDiscoveryTestHelper.generateKeyPairs(2);
+    final List<DiscoveryPeer> peers = helper.createDiscoveryPeers(keypairs);
 
-    final PeerDiscoveryAgent agent = mock(PeerDiscoveryAgent.class);
-    when(agent.getAdvertisedPeer()).thenReturn(peers[0]);
+    final MockPeerDiscoveryAgent agent = mock(MockPeerDiscoveryAgent.class);
+    when(agent.getAdvertisedPeer()).thenReturn(peers.get(0));
+    DiscoveryPeer self = peers.get(0);
```
shemnon: Again, `self` as a variable may cause python confusion with `this`, an alternate name would improve readability.

ethereum/p2p/src/test/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/PeerDiscoveryControllerTest.java
```diff
@@ -65,24 +69,19 @@
   private static final RetryDelayFunction SHORT_DELAY_FUNCTION = (prev) -> Math.max(100, prev * 2);
   private static final PeerRequirement PEER_REQUIREMENT = () -> true;
   private static final long TABLE_REFRESH_INTERVAL_MS = TimeUnit.HOURS.toMillis(1);
-  private final Vertx vertx = spy(Vertx.vertx());
-  private PeerDiscoveryAgent agent;
   private PeerDiscoveryController controller;
-  private DiscoveryPeer peer;
+  private DiscoveryPeer self;
```
shemnon: Again, `self` may not be the best name here.  Perhaps `thisPeer`?

ethereum/p2p/src/test/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/MockTimerUtil.java
```diff
@@ -0,0 +1,65 @@
...
+
+public class MockTimerUtil implements TimerUtil {
+  private final AtomicLong nextId = new AtomicLong(0);
+  private final Map<Long, TimerHandler> timerHandlers = new HashMap<>();
+  private final Map<Long, TimerHandler> periodicHandlers = new HashMap<>();
+
+  @Override
+  public long setPeriodic(final long delay, final TimerHandler handler) {
+    long id = nextId.incrementAndGet();
+    periodicHandlers.put(id, handler);
```
shemnon: Do we have any test cases that depend on different delays between peers for timers or handlers?

ethereum/p2p/src/test/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/MockTimerUtil.java
```diff
@@ -0,0 +1,65 @@
...
+
+  public void runHandlers() {
+    runTimerHandlers();
+    runPeriodicHandlers();
+  }
+
+  public void runTimerHandlers() {
+    List<TimerHandler> handlers = new ArrayList<>();
+    timerHandlers.forEach((id, handler) -> handlers.add(handler));
+    timerHandlers.clear();
```
shemnon: `handlers = timerHandlers.values();` would be better.

ethereum/p2p/src/test/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/MockTimerUtil.java
```diff
@@ -0,0 +1,65 @@
...
+
+    handlers.forEach(TimerHandler::handle);
+  }
+
+  public void runPeriodicHandlers() {
+    List<TimerHandler> handlers = new ArrayList<>();
```
shemnon: Could be a 1 liner - `periodicHandlers.forEach((id, handler) -> handler.handle);` or `periodicHandlers.values().forEach(TimerHandler::handler);`

Unless you are seeing concurrent modification in which case it would be a 2 liner storing values in a List.  If you are seeing concurrent modification it would be worth a comment.

ethereum/p2p/src/test/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/MockTimerUtil.java
mbaxter: We're not doing anything very complicated with the timers.  We can always extend this in the future if we need to.

ethereum/p2p/src/test/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/MockTimerUtil.java
mbaxter: The docs say `values()` is a view into the collection.  I'm making a copy so that timers added as we run the handlers aren't run as part of this round of processing.  Will add a comment.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/PeerDiscoveryAgent.java
```diff
@@ -53,282 +51,201 @@
...
- *       etc. based on its internal state.
- *   <li>The agent attaches a callback to the call to the Peer Table. When the Peer Table has
- *       processed the message, it'll perform a callback passing in an Optional which is populated
- *       if we recognised the Peer, and empty if we did not.
- *   <li>The agent reacts to specific messages (PING-&gt;PONG, FIND_NEIGHBORS-&gt;NEIGHBORS), if the
- *       Peer was recognised. Why doesn't the table send these messages itself? Because they don't
- *       affect the state machine of the Peer, and the table is only concerned with storing peers,
- *       keeping them alive and tracking their state. It is not bothered to service requests.
- * </ol>
+ * messages via UDP.
```
s-matthew-english:
```
/**
 * The peer discovery agent is the network component that sends and receives messages peer discovery
 * messages via UDP.
 */
```
^This is strangely worded and not really informative.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/PeerDiscoveryAgent.java
```diff
@@ -53,282 +51,201 @@
...
+    this.peerBlacklist = peerBlacklist;
+    this.nodeWhitelistController = nodeWhitelistController;
+    this.bootstrapPeers =
         config.getBootstrapPeers().stream().map(DiscoveryPeer::new).collect(Collectors.toList());
 
-    this.vertx = vertx;
     this.config = config;
     this.keyPair = keyPair;
     this.peerTable = new PeerTable(keyPair.getPublicKey().getEncodedBytes(), 16);
-    this.controller =
```
s-matthew-english: interesting, did you remove that circular dependency thing?

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/PeerDiscoveryAgent.java
```diff
@@ -53,282 +51,201 @@
...
+  protected abstract TimerUtil createTimer();
+
+  protected abstract CompletableFuture<InetSocketAddress> listenForConnections();
+
+  protected abstract CompletableFuture<Void> sendOutgoingPacket(
+      final DiscoveryPeer peer, final Packet packet);
+
+  public abstract CompletableFuture<?> stop();
+
+  public CompletableFuture<?> start() {
```
s-matthew-english: wow- interesting, so this only uses netty now?

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/PeerDiscoveryAgent.java
```diff
@@ -53,282 +51,201 @@
...
+      listenForConnections()
+          .thenAccept(
+              (InetSocketAddress localAddress) -> {
+                // Once listener is set up, finish initializing
+                this.localAddress = localAddress;
+                advertisedPeer =
+                    new DiscoveryPeer(
+                        id,
+                        config.getAdvertisedHost(),
+                        localAddress.getPort(),
```
s-matthew-english: why do we have two of `localAddress.getPort()` here? is it related to:

```
final String host = config.getBindHost();
final int port = config.getBindPort();
```

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/PeerDiscoveryAgent.java
mbaxter: yeah

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/PeerDiscoveryAgent.java
mbaxter: No, but I moved all the vertx code into a subclass. Shouldn't be too much work to migrate to netty now.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/PeerDiscoveryAgent.java
```diff
@@ -53,282 +51,201 @@
...
-      LOG.debug("Discarding invalid peer discovery packet", e);
-    } catch (final Throwable t) {
-      LOG.error("Encountered error while handling packet", t);
     }
+
+    // Notify the peer controller.
+    String host = sourceEndpoint.getHost();
+    int port = sourceEndpoint.getUdpPort();
+    final DiscoveryPeer peer = new DiscoveryPeer(packet.getNodeId(), host, port, tcpPort);
+    controller.ifPresent(c -> c.onMessage(packet, peer));
```
s-matthew-english: I kind of hate variables named `c`

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
```diff
@@ -12,17 +12,16 @@
  */
 package tech.pegasys.pantheon.ethereum.p2p.discovery.internal;
 
-import static com.google.common.base.Preconditions.checkNotNull;
 import static java.util.Collections.emptyList;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static tech.pegasys.pantheon.ethereum.p2p.discovery.internal.PeerTable.AddResult.Outcome;
 
+import tech.pegasys.pantheon.crypto.SECP256K1;
```
s-matthew-english: nice, this looks good. I don't think it'll be that bad to merge

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/VertxTimerUtil.java
```diff
@@ -0,0 +1,39 @@
+/*
```
s-matthew-english: interesting idea, I like it

ethereum/p2p/src/test/java/tech/pegasys/pantheon/ethereum/p2p/NettyP2PNetworkTest.java
```diff
@@ -296,7 +296,7 @@ public void rejectPeerWithNoSharedCaps() throws Exception {
                 new PeerBlacklist(),
                 new NoOpMetricsSystem(),
                 new NodeWhitelistController(PermissioningConfiguration.createDefault()))) {
-      final int listenPort = listener.getSelf().getPort();
+      final int listenPort = listener.getLocalPeerInfo().getPort();
```
s-matthew-english: local peer is a much more sane name, I like that

s-matthew-english:
nice I think it looks good, I only really looked at the Agent, which is where most of the important changes are I think, and then at the Controller too, 'cause it's related to what I'm working on rn, the tests I didn't look at too closely

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/PeerDiscoveryAgent.java
mbaxter: That's how it behaved before.  But it does seem wrong - I'll try to follow-up with a fix so I don't keep pushing stuff into this PR.

PR #547, opened 2019-01-14
[NC-1977] Provide a better error message when RocksDB reports an unsupported system

ajsutton:
## PR description
Provide a better error message when running on a system configuration that RocksDB doesn't support (most commonly, when using 32bit java on Windows...)

## Fixed Issue(s)
https://pegasys1.atlassian.net/browse/NC-1977


ajsutton:
For the record, we can't write a test for this because RocksDB checks the platform details in class static initialisers so we can't set them to something that isn't supported and then change them back - they have to be set before RocksDB is ever loaded and then they're locked in for the lifetime of that JVM.

PR #551, opened 2019-01-14
Add timer metrics for EVM operations

ajsutton:
## PR description
Adds a metric to time each operation performed by the EVM to allow us to see how much time is spent on each operation and how common each operation is.


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/ProtocolSpecBuilder.java
```diff
@@ -74,7 +76,8 @@
     return this;
   }
 
-  public ProtocolSpecBuilder<T> evmBuilder(final Function<GasCalculator, EVM> evmBuilder) {
+  public ProtocolSpecBuilder<T> evmBuilder(
+      final BiFunction<GasCalculator, MetricsSystem, EVM> evmBuilder) {
```
shemnon: This is why I think we need a DI system.  The Function/BiFunction/TriFuction/... becomes a Supplier<> and the context is provided by the DI system.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/ProtocolScheduleBuilder.java
```diff
@@ -13,20 +13,24 @@
...
 
 public class ProtocolScheduleBuilder<C> {
 
+  private final MetricsSystem metricsSystem;
   private final GenesisConfigOptions config;
   private final Function<ProtocolSpecBuilder<Void>, ProtocolSpecBuilder<C>> protocolSpecAdapter;
   private final int defaultChainId;
 
   public ProtocolScheduleBuilder(
+      final MetricsSystem metricsSystem,
```
shemnon: Giant constructors don't feel like the Builder pattern to me.  Could this (here or another PR) be re-written to follow the builder pattern?

metrics/src/main/java/tech/pegasys/pantheon/metrics/MetricCategory.java
```diff
@@ -13,6 +13,7 @@
 package tech.pegasys.pantheon.metrics;
 
 public enum MetricCategory {
+  EVM("evm"),
```
shemnon: Should new categories go at the end of the list?  If not should we do a more sensible ordering like alphabetization?

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/ProtocolSpecBuilder.java
ajsutton: The point of this ugliness is that you can take a `ProtocolSpecBuilder` configured for one milestone and change some of the inputs to adapt it to either a different milestone or for something like IBFT or Clique.  So Frontier defines a evmBuilder which is unchanged in homestead, but homestead provides a different GasCalculator input. I imagine you can make a DI system achieve this but I'd expect it to be a fairly advanced use case that winds up being complex either in how it's written or how deep an understanding of the DI library you need.

It is another one of those things where at some point we need to just try it and see how it actually comes out though.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/ProtocolScheduleBuilder.java
ajsutton: I suspect this should actually be named Factory not Builder so have renamed it as such.  The point is that it encapsulates the logic for interpreting a genesis config and setting up all the protocol specs at the right block numbers.  It's not designed to give the pseudo-named-parameter style you're expecting.

metrics/src/main/java/tech/pegasys/pantheon/metrics/MetricCategory.java
ajsutton: Alphabetical it is.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/ProtocolScheduleBuilder.java
shemnon: Naming it factory is a viable alternative.

PR #565, opened 2019-01-15
Ibft validates block on proposal reception

rain-on:
Prior to this change, only the header was validated on block reception - now the block will be fully processed/validated (other than the commit seals, which will not exist until later in the process).

This has necessitated splitting up the BlockImporter into an Importer and a validator - both of which are available from the protocolSpecification.

Finally - an error existed in Ibft whereby the blockheadervalidator was specified at application startup, rather than being derived from the current block number (i.e. from ProtocolSchedule) - this has now been resolved.


consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/IbftRoundFactory.java
```diff
@@ -47,6 +48,9 @@ public IbftRound createNewRound(final BlockHeader parentHeader, final int round)
     final ConsensusRoundIdentifier roundIdentifier =
         new ConsensusRoundIdentifier(nextBlockHeight, round);
 
+    BlockValidator<IbftContext> blockValidator =
```
ajsutton: nit: final

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/validation/MessageValidator.java
```diff
@@ -97,9 +97,13 @@ private boolean validateSignedProposalPayload(final SignedData<ProposalPayload>
     }
 
     final Block proposedBlock = msg.getPayload().getBlock();
-    if (!headerValidator.validateHeader(
-        proposedBlock.getHeader(), parentHeader, protocolContext, FULL)) {
-      LOG.info("Invalid Proposal message, block did not pass header validation.");
+
+    final Optional<BlockValidationOutputs> validationResult =
```
ajsutton: I feel like `validationResult` is a misleading name for this - I'd expect it to either be empty to indicate valid or have an enum to give the reason it was invalid.  This is maybe more `blockProcessingOutputs`?  The `BlockValidationOutputs` class might be better as `BlockProcessingOutputs` as well. Not sure there's an ideal name in any case.

rain-on:
Too large, so has been broken into sub PRs.

rain-on:
Closed.

PR #578, opened 2019-01-16
Refactor RoundChangeManager creation

rain-on:
Constructor of the RoundChangeManager now accepts a
RoundChangeMessageValidator rather than creating it from inputs.

This removes complexity from this class, pushes creation back to the
validator factory, and simplifies the IbftBlockHeightManagerFactory.


consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/RoundChangeManager.java
```diff
@@ -81,20 +77,13 @@ public RoundChangeCertificate createRoundChangeCertificate() {
   @VisibleForTesting
   final Map<ConsensusRoundIdentifier, RoundChangeStatus> roundChangeCache = Maps.newHashMap();
 
-  private final int quorumSize;
+  private final long quorum;
```
jframe: Bit odd we are using quorum here and quorumSize elsewhere. Though it probably should be just quorum everywhere.

rain-on: yeah agreed - will do a quick round up.

PR #584, opened 2019-01-16
Register privacy

Puneetha17:


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/ProtocolScheduleFactory.java
```diff
@@ -24,19 +25,22 @@
...
+      final PrivacyParameters privacyParameters) {
     this.metricsSystem = metricsSystem;
     this.config = config;
     this.protocolSpecAdapter = protocolSpecAdapter;
     this.defaultChainId = defaultChainId;
+    this.privacyParameters = privacyParameters;
   }
 
-  public ProtocolSchedule<C> createProtocolSchedule() {
+  public MutableProtocolSchedule<C> createProtocolSchedule() {
```
Errorific: Can this possibly continue to use the more general ProtocolSchedule interface?

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/PrecompiledContractConfiguration.java
```diff
@@ -0,0 +1,43 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package tech.pegasys.pantheon.ethereum.mainnet;
+
+import tech.pegasys.pantheon.ethereum.core.PrivacyParameters;
+import tech.pegasys.pantheon.ethereum.vm.GasCalculator;
+
+public class PrecompiledContractConfiguration {
```
Errorific: Is there any utility to this class beyond holding 2 objects? Could it's consumers take the gas calculator and privacy parameters separately without losing functionality?

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/MainnetProtocolSpecs.java
```diff
@@ -193,6 +193,12 @@ private MainnetProtocolSpecs() {}
         .name("Constantinople");
   }
 
+  public static ProtocolSpecBuilder<Void> privacyDefinition(final int chainId) {
```
Errorific: As per comments about how to build privacy into the protocol schedule, don't think privacy should be defined as a spec built on top of constantinople, rather something that can be applied to any spec during building.

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -444,6 +445,24 @@ private Long configureRefreshDelay(final Long refreshDelay) {
...
   private final Collection<String> accountsWhitelist = null;
 
+  @Option(
+    names = {"--privacy-url"},
+    description = "The URL on which enclave is running "
+  )
+  private final URI privacyUrl = PrivacyParameters.DEFAULT_ORION_URI;
+
+  @Option(
+    names = {"--privacy-public-key"},
```
NicolasMassart: We did not update the CLI guideline yet but to comply with the next CLI refactoring that will be released in the 0.9 version, you could already name this option `--privacy-public-key-file`

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -444,6 +445,24 @@ private Long configureRefreshDelay(final Long refreshDelay) {
   )
   private final Collection<String> accountsWhitelist = null;
 
+  @Option(
+    names = {"--privacy-url"},
+    description = "The URL on which enclave is running "
+  )
+  private final URI privacyUrl = PrivacyParameters.DEFAULT_ORION_URI;
```
NicolasMassart: may `DEFAULT_ORION_URI` be `DEFAULT_ORION_URL` if the content is a URL (and URL is a URI, hence the type, but semantically if you ask the user for a URL, the default value should be called a URL too IMO)

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/MainnetPrecompiledContractRegistries.java
```diff
@@ -36,20 +37,39 @@ private static void populateForFrontier(
...
+  }
+
+  public static PrecompileContractRegistry byzantium(
+      final PrecompiledContractConfiguration precompiledContractConfiguration) {
+    final PrecompileContractRegistry registry = new PrecompileContractRegistry();
+    populateForByzantium(registry, precompiledContractConfiguration.getGasCalculator());
+    return registry;
+  }
+
+  public static PrecompileContractRegistry privacy(
```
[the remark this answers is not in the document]
iikirilov: According to the eea spec - you MUST include the 8 precompiled contracts as in the Ethereum yellowpaper. So this contract MUST only be added to a registry which contains the first 8 precompiled contracts.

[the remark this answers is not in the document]
iikirilov: The constraint being - privacy can be added to any registry built on top of constantinople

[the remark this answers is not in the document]
Errorific: Hi @iikirilov,
Could you please provide a reference to the section in the EEA spec that refers to this?
Thanks.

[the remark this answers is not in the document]
Errorific: Hi @iikirilov,
After discussing it with a member of our standards team the spec indicates that the client MUST support the Constantinople milestone and MUST support a method of private transactions of which the private contract is one. Supporting both of these does not imply a dependency between them and whether both features are turned on at the same time or not is up to the users deploying the chain. It is very conceivable and allowed that users will want an earlier milestone and privacy or a later milestone and no privacy. We don't believe that the constraint you've outlined exists in the EEA spec or is the intention of the spec. Would be keen to discuss this further to find that either we're wrong or that the spec requires more clarification on this matter.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/ProtocolSpecBuilder.java
```diff
@@ -118,11 +121,17 @@
   }
 
   public ProtocolSpecBuilder<T> precompileContractRegistryBuilder(
-      final Function<GasCalculator, PrecompileContractRegistry> precompileContractRegistryBuilder) {
+      final Function<PrecompiledContractConfiguration, PrecompileContractRegistry>
```
Errorific: This function can be changed such that you don't need appendPrivacyRegistry and you don't need to pass around so many functions of functions. When the 
```
  public ProtocolSpecBuilder<T> precompileContractRegistryBuilder(
      final Function<PrecompiledContractConfiguration, PrecompileContractRegistry>
          precompileContractRegistryBuilder) {
    this.precompileContractRegistryBuilder = (precompiledContractConfiguration) -> {
      final PrecompileContractRegistry registry = precompileContractRegistryBuilder.apply(precompiledContractConfiguration);
      if (precompiledContractConfiguration.getPrivacyParameters().isEnabled()) {
        registry.put(
          Address.DEFAULT_PRIVACY,
          new PrivacyPrecompiledContract(
            precompiledContractConfiguration.getGasCalculator(),
            precompiledContractConfiguration.getPrivacyParameters()
          )
        );
      }
      return registry;
    };
    return this;
  }
```

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/MainnetPrecompiledContractRegistries.java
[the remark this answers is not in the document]
iikirilov: I just had a re-read of the spec and the way you have described it makes sense. I think I misunderstood. For reference this is where I read it https://entethalliance.github.io/client-spec/spec.html#sec-cross-client-compatibility

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/PrivacyParameters.java
```diff
@@ -0,0 +1,62 @@
...
+
+  public File getPublicKey() {
+    return publicKey;
+  }
+
+  public void setPublicKey(final File publicKey) {
+    this.publicKey = publicKey;
+  }
+
+  public static PrivacyParameters createDefault() {
```
lucassaldanha: I wonder if we should name this differently. Maybe `noPrivacy()` or something to indicate that the default is using Pantheon without any privacy features. Thoughts?

Puneetha17: Yes, that makes sense. It makes it clear that the default we use has privacy disabled.
Will update the same.

PR #592, opened 2019-01-17
[PRIV] Add Private transaction abstraction

iikirilov:
## PR description

This PR creates the private transaction abstraction

## Fixed Internal Issue(s)

[PRIV-12](https://pegasys1.atlassian.net/browse/PRIV-12)

## Reasoning

It was decided to duplicate the `Transaction` class because `PrivateTransaction` and `Transaction` should never be used in the same code path -> therefore making common interface felt unnatural

Future refactor can reduce code duplication. My initial proposition is that the static methods can be moved to a `Transactions` utility class. Then `PrivateTransaction` and `Transaction` can be simple data classes. This would remove the majority of code duplication.


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/privacy/PrivateTransaction.java
```diff
@@ -0,0 +1,606 @@
...
+      signature(computeSignature(keys));
+      sender(Address.extract(Hash.hash(keys.getPublicKey().getEncodedBytes())));
+      return build();
+    }
+
+    protected SECP256K1.Signature computeSignature(final SECP256K1.KeyPair keys) {
+      final OptionalInt optionalChainId =
+          chainId > 0 ? OptionalInt.of(chainId) : OptionalInt.empty();
+      final Bytes32 hash =
+          computeSenderRecoveryHash(nonce, gasPrice, gasLimit, to, value, payload, optionalChainId);
```
antonydenyer: Should privacy params be passed through here?

PR #593, opened 2019-01-17
fixes #522 remove mention of gas setting from eth_estimateGas

### LINKED ISSUE #522: Gas limit is irrelevant to eth_estimateGas
Currently the description of eth_estimateGas includes:
_If you do not specify a gas amount in the transaction call object, Pantheon uses the gasLimit amount from the block at the head of the chain as the upper limit. If the amount of gas needed is higher than the estimated upper limit, the transaction might not have enough gas to execute._

This is misleading:
* The result for eth_estimateGas is not changed by the gas limit (whether the gas limit comes from the transaction call object or from the chain head) 
* eth_estimateGas doesn't actually execute the transaction

donntrenton:
## PR description
In JSON-RPC-API-Methods.md/#eth_estimateGas:

- Revised the intro paragraph in the Parameters section to remove the mention of the `gas` setting:

The transaction call object parameters are the same as those for `eth_call`, except that in `eth_estimateGas`, all fields are optional. Pantheon uses the `gasLimit` amount from the block at the head of the chain as the upper limit. If the amount of gas needed is higher than the estimated upper limit, the transaction might not have enough gas to execute.

In JSON-RPC-API-Objects.md/#transaction-call-object:

- Updated description of `gas` to: "Gas provided for the transaction execution. `eth_call` consumes zero gas, but this parameter might be needed by some executions. `eth_estimateGas` ignores this value."

## Fixed Issue(s)
fixes #522


docs/Reference/JSON-RPC-API-Methods.md
```diff
@@ -820,15 +820,13 @@ gas and is not added to the blockchain as a transaction. The estimate can be gre
...
 
 **Parameters**
 
-!!!note
-    The transaction call object parameters are the same as those for [eth_call](#eth_call), except that in `eth_estimateGas`
-    all fields are optional. If you do not specify a `gas` amount in the transaction call object, Pantheon uses the
-    `gasLimit` amount from the block at the head of the chain as the upper limit. If the amount of gas needed is higher
-    than the estimated upper limit, the transaction might not have enough gas to execute.
+The transaction call object parameters are the same as those for [eth_call](#eth_call), except that in `eth_estimateGas`,
+all fields are optional. Pantheon uses the `gasLimit` amount from the block at the head of the chain as the upper limit.
```
MadelineMurray: I thought the testing you did on this (setting `gasLimit` in the genesis file) showed that changing the `gasLimit` amount from the block at the head of the chain had no effect on the estimate returned?

docs/Reference/JSON-RPC-API-Methods.md
```diff
@@ -820,15 +820,13 @@ gas and is not added to the blockchain as a transaction. The estimate can be gre
...
+If the amount of gas needed is higher than the estimated upper limit, the transaction might not have enough gas to execute.
```
MadelineMurray: `eth_estimateGas` doesn't execute the transaction -> I'm not sure this is clear here. `eth_estimateGas` is always going to execute regardless of the gas limit.

docs/Reference/JSON-RPC-API-Methods.md
donntrenton: Yes--as we discussed, I'll remove this misleading text: "Pantheon uses the `gasLimit` amount from the block at the head of the chain as the upper limit. If the amount of gas needed is higher than the estimated upper limit, the transaction might not have enough gas to execute."

docs/Reference/JSON-RPC-API-Methods.md
donntrenton: I'll remove this sentence too as noted above.

docs/Reference/JSON-RPC-API-Methods.md
```diff
@@ -817,20 +817,18 @@ You can interact with contracts using [eth_sendRawTransaction or eth_call](../Us
 
 ### eth_estimateGas
 
-Returns an estimate of how much gas is needed for the transaction to complete. The estimate process does not use
-gas and is not added to the blockchain as a transaction. The estimate can be greater than the amount of gas that the
-transaction actually uses, for reasons including EVM mechanics and node performance.
+Returns an estimate of how much gas is needed for a future transaction to complete. The estimation process does not use
```
MadelineMurray: I think 'future' is redundant and could be removed (it's covered in the following sentence).

docs/Reference/JSON-RPC-API-Methods.md
```diff
@@ -817,20 +817,18 @@ You can interact with contracts using [eth_sendRawTransaction or eth_call](../Us
...
+gas and is not added to the blockchain as a transaction. Setting a gas limit is irrelevant to the estimation process
```
MadelineMurray: 'and is not added' -> 'and the transaction is not added' 

I think it would be clearer to move the part about setting a gas limit having no effect the the parameters section and say specifically something like 'The `gas` property on the transaction all object is ignored by `eth_estimateGas`

docs/Reference/JSON-RPC-API-Methods.md
donntrenton: Deleted 'future'.

PR #596, opened 2019-01-17
Fix deprecation warnings

ajsutton:
## PR description

Vertx deprecated a few methods which caused some new warnings when building. Two of them have simple replacements (you can now use the router directly instead of needing a method reference) and the other had to be suppressed as the method is being changed in vertx 4 but I don't see a suitable replacement while still on vertx 3.

And while in the files, fix the other warnings IntellIJ flagged (mostly adding final).


ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/JsonRpcHttpService.java
```diff
@@ -156,7 +155,7 @@ private void validateConfig(final JsonRpcConfiguration config) {
               }
               httpServer = null;
               final Throwable cause = res.cause();
-              if (cause instanceof BindException || cause instanceof SocketException) {
+              if (cause instanceof SocketException) {
```
ajsutton: For the record, BindException is a subclass of SocketException.

PR #597, opened 2019-01-17
[NC-2129] Add GetNodeDataFromPeerTask

ajsutton:
## PR description
Adds GetNodeDataFromPeerTask that requests node data from a peer and matches up the results.

## Fixed Issue(s)
NC-2129


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/tasks/GetNodeDataFromPeerTask.java
```diff
@@ -0,0 +1,89 @@
...
+    final NodeDataMessage nodeDataMessage = NodeDataMessage.readFrom(message);
+    final List<BytesValue> nodeData = nodeDataMessage.nodeData();
+    if (nodeData.isEmpty()) {
+      return Optional.empty();
+    } else if (nodeData.size() > hashes.size()) {
+      // Can't be the response to our request
+      return Optional.empty();
+    }
+
+    if (nodeData.stream().anyMatch(data -> !hashes.contains(Hash.hash(data)))) {
```
mbaxter: Might be nice to save this hash calculation, so we get back a mapping from hash -> data.  Just saves us from duplicative hash calculations if we're already going to do that here.

mbaxter:
Going to go ahead and merge since I need this :)

ajsutton:
Ok, since it's merged I'll leave it with you to save the hash calculation if it's useful to the calling code.  It should be pretty straight-forward.

PR #598, opened 2019-01-18
[NC-1344] Add GetReceiptsFromPeerTask

ajsutton:
## PR description

Adds a GetReceiptsFromPeerTask to support getting transaction receipts from a peer.  Receipts are actually requested based on the *block* hash, not the receipt root, but then we have to match the returned receipts back up to the headers using the receipt root.  The task deduplicates headers with the same receipts root to minimise data transferred.

The task returns a map of `BlockHeader` to `List<TransactionReceipt>`.  It could have just used the block hash as the key but that could easily be confused for the receipt root which is also of type `Hash`.

## Fixed Issue(s)
NC-1344


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/tasks/GetReceiptsFromPeerTask.java
```diff
@@ -0,0 +1,126 @@
...
+  protected ResponseStream sendRequest(final EthPeer peer) throws PeerNotConnected {
+    LOG.debug("Requesting {} receipts from peer {}.", blockHeaders.size(), peer);
+    // Since we have to match up the data by receipt root, we only need to request receipts
+    // for one of the headers with each unique receipt root.
+    final List<Hash> blockHashes =
+        headersByReceiptsRoot
+            .values()
+            .stream()
+            .map(headers -> headers.get(0).getHash())
+            .collect(toList());
```
mbaxter: We probably want to filter out requests for empty receipt sets.  Fine to follow-up with that optimization if you want to go ahead and merge this though.

ajsutton: I thought about that but if everything in the list winds up being an empty receipt we wouldn't send a request at all and then we don't have a `ResponseStream` to return and everything gets weird.  So I *think* it's actually better to handle that case back in the calling code - the downside being that it doesn't automatically get re-used if the task winds up called from a second place.

mbaxter: hmm - wonder if we could have this return an `Optional<ResponseStream>` or something?  In any case, I say we make a ticket to look into it and just merge this as-is for now.

ajsutton: Yeah that's sensible - then just need to have a way for the task to know it should insert the empty result, which probably boils down to just having a flag set or a condition in the process results section.

ajsutton: I've added NC-2148 as a subtask so I don't forget this as I work through the actual downloading.

PR #601, opened 2019-01-18
NC-2147 Implement Petersburg hardfork

shemnon:
## PR description

* create a new fork definition
* make it Constantinople without EIP-1283
* update tests where appropriate

## Fixed Issue(s)

NC-2147


ethereum/core/src/jmh/java/tech/pegasys/pantheon/ethereum/vm/operations/BlockHashOperationBenchmark.java
```diff
@@ -42,6 +42,7 @@
   @Setup
   public void prepare() throws Exception {
     operationBenchmarkHelper = OperationBenchmarkHelper.create();
+    // ?? do we create another benchmark ??
```
ajsutton: No, I think the Constantinople benchmark is all we need here as the operation of block hash is unaffected (technically we could use a much earlier milestone) and the gas calculator isn't what we're trying to benchmark anyway.

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/core/TransactionTest.java
```diff
@@ -93,6 +93,11 @@ public void constantinople() {
     milestone("Constantinople");
   }
 
+  @Test
+  public void petersburg() {
+    milestone("Petersburg");
+  }
+
```
ajsutton: Since the transaction tests haven't been updated for "Petersburg" we shouldn't add this.  I doubt there will be additional transaction tests added since they're just about parsing transactions, not about executing them and that's unaffected.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/MainnetProtocolSpecs.java
```diff
@@ -193,6 +193,12 @@ private MainnetProtocolSpecs() {}
         .name("Constantinople");
   }
 
+  public static ProtocolSpecBuilder<Void> petersburgDefinition(final int chainId) {
+    return constantinopleDefinition(chainId)
+        .gasCalculator(TangerineWhistleGasCalculator::new)
```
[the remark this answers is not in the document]
ajsutton: Good spot - we will because Constantinople gas calculator also provides the gas cost for EXTCODEHASH and CREATE2 which aren't available in tangerine whistle. It should omit the `calculateStorageCost` and `calculateStorageRefundAmount` overrides that are in `ConstantinopleGasCalculator`.

ethereum/core/src/jmh/java/tech/pegasys/pantheon/ethereum/vm/operations/BlockHashOperationBenchmark.java
shemnon: Now that I added a PetersburgGasCalculator I moved it over to that one, just to keep current.

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/core/TransactionTest.java
shemnon: Done

shemnon:
Ready for another review.
* Petersburg became ConstantinopleFix
* Fork ordering verified (had to fix Clique Tests)
* Ref Tests pass at the same rate as before

PR #603, opened 2019-01-19
Added IBFT get pending votes

rojotek:
Pulled out an AbstractVoteProposerMethod base class from the CliqueProposals json rpc method, and used this for both the CliqueProposals end point and the newly created, corresponding IbftGetPendingVotes endpoint.


consensus/common/src/main/java/tech/pegasys/pantheon/consensus/common/jsonrpc/AbstractVoteProposerMethod.java
```diff
@@ -0,0 +1,45 @@
...
+import tech.pegasys.pantheon.ethereum.jsonrpc.internal.JsonRpcRequest;
+import tech.pegasys.pantheon.ethereum.jsonrpc.internal.response.JsonRpcResponse;
+import tech.pegasys.pantheon.ethereum.jsonrpc.internal.response.JsonRpcSuccessResponse;
+
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class AbstractVoteProposerMethod {
+
+  protected final VoteProposer voteProposer;
```
rain-on: protected or private?

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/jsonrpc/methods/IbftGetPendingVotes.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+package tech.pegasys.pantheon.consensus.ibft.jsonrpc.methods;
+
+import tech.pegasys.pantheon.consensus.common.VoteProposer;
+import tech.pegasys.pantheon.consensus.common.jsonrpc.AbstractVoteProposerMethod;
+import tech.pegasys.pantheon.ethereum.jsonrpc.internal.methods.JsonRpcMethod;
+
+public class IbftGetPendingVotes extends AbstractVoteProposerMethod implements JsonRpcMethod {
+
+  public static final String PENDING_VOTES = "ibft_getPendingVotes";
```
rain-on: nit: should this be just called "JSON_RPC_NAME"? Then we could use the same thing everywhere ...

consensus/ibft/src/test/java/tech/pegasys/pantheon/consensus/ibft/jsonrpc/methods/IbftGetPendingVotesTest.java
```diff
@@ -0,0 +1,73 @@
...
+    method = new IbftGetPendingVotes(voteProposer);
+  }
+
+  @Test
+  public void returnsCorrectMethodName() {
+    assertThat(method.getName()).isEqualTo(METHOD_NAME);
+  }
+
+  @Test
+  public void testConversionFromVoteTypeToBoolean() {
```
rain-on: should this be moved into the abstract class's test?

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/jsonrpc/methods/IbftGetPendingVotes.java
rain-on: Actually - I just saw that ibftGetValidatorsByBlockHash doesn't bother with the static member (i.e. the string name is just defined in the getName() function)... not sure which way I like more - but we should be consistent.

PR #615, opened 2019-01-21
Ibft Integration Tests - Spurious Behaviour

rain-on:
This commit adds additional tests which fall outside of the existing
suite, and target spuriously behaved peers - i.e. illegal seals,
illformed RLP data etc.


consensus/ibft/src/integration-test/java/tech/pegasys/pantheon/consensus/ibft/tests/SpuriousBehaviourTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+    expectedPrepare =
+        context
+            .getLocalNodeMessageFactory()
+            .createSignedPreparePayload(roundId, proposedBlock.getHash());
+    expectedCommit =
+        createSignedCommitPayload(
+            roundId, proposedBlock, context.getLocalNodeParams().getNodeKeyPair());
+  }
+
```
CjHare: Do these two illegal RLP cover the control flow if junk RLP (RLP that doesn't 'fit' into the expected data structure)?

rain-on: The first test ensures RLP which doesn't match a given message ID doesn't prevent ongoing IBFT operation.

The latter test ensures if the seal in a commit msg is bad, that it is ignored.

PR #621, opened 2019-01-21
Trigger a peer refresh immediately upon startup.

ajsutton:
## PR description
Previously `PeerDiscoveryController` waited 30 seconds after startup before initiating a peer table refresh which guaranteed Pantheon would sit around for 30 seconds doing nothing.

With this change, it immediately starts peer discovery so things start happening sooner.


ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
```diff
@@ -156,15 +156,17 @@ public PeerDiscoveryController(
...
         .forEach(node -> bond(node, true));
 
     final long timerId =
         timerUtil.setPeriodic(
             Math.min(REFRESH_CHECK_INTERVAL_MILLIS, tableRefreshIntervalMs),
-            () -> refreshTableIfRequired());
+            this::refreshTableIfRequired);
     tableRefreshTimerId = OptionalLong.of(timerId);
 
+    refreshTableIfRequired();
```
mbaxter: I don't think this really makes sense - on start we begin the bonding process with the bootstrap peers.  That is basically a table refresh operation: once we bond, we start requesting peers.

ajsutton: We don't send a FIND_NEIGHBOURS request when bonding with a node completes so bonding to the bootstrap nodes doesn't actually perform any discovery.

ajsutton: You're right it is - I missed that.  Wonder why it never seems to actually find any nodes...

PR #629, opened 2019-01-22
[NC-1273] Consider peer count insufficient until minimum peers for fast sync are connected

ajsutton:
## PR description
When fast sync is enabled, consider the peer count insufficient until we have enough to start fast sync. `PeerDiscoveryController` will then refresh the peer table more often to find extra peers more quickly.


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/EthProtocolManager.java
```diff
@@ -61,21 +62,23 @@
...
+  private final int requiredPeers;
   private List<Capability> supportedCapabilities;
   private final Blockchain blockchain;
 
   EthProtocolManager(
       final Blockchain blockchain,
       final WorldStateArchive worldStateArchive,
       final int networkId,
-      final boolean fastSyncEnabled,
+      final SynchronizerConfiguration syncConfig,
```
mbaxter: Originally, I intended `EthProtocolManager` to be something like a generic service provider - a base layer for handling communication over the network related to the `eth` subprotocol without holding any process-specific code.  So, I tried to keep the sync package dependent on the management layer without the management layer depending on sync-specific code.  Not a huge deal, but I try to keep an eye on circular dependencies between packages.  

One alternative idea would be to expose something like a `PeerRefreshRequests` object on `EthContext`.  And the manager could then check and clear that when `hasSufficientPeers()` is invoked.

PR #631, opened 2019-01-22
[PRIV] Prepare private transaction support

antonydenyer:
## PR description
Adds an Orion client with integration tests.

The purpose of this PR is to prepare for supporting private transaction in Pantheon. To support private transaction it is required to access an enclave from Pantheon. As Orion is an enclave created by the same organisation that maintains Pantheon it seems natural to support the Orion enclave first.

No functionality is added to Pantheon in this PR.

A copy of https://github.com/PegaSysEng/pantheon/pull/538 to work round build issues.

## Fixed Issue(s)
fixes #538 build issues


antonydenyer:
This was raised to try and get around issues with jenkins, but is now redundant.

PR #635, opened 2019-01-22
NC-2104 Clique and iBFT should not be defined by default in RPC APIs

NicolasMassart:
## PR description
Clique and iBFT should not be defined by default in HTTP and Websocket RPC APIs options.
This PR removes the default values for both Clique and iBFT and removes the hard coded string used as default value for description and replaces it with a constant default that's can be changed in one place.

## Fixed Issue(s)

fixes NC-2104

### DIFF
--- a/ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/RpcApi.java
+++ b/ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/RpcApi.java
@@ -45,6 +45,6 @@ public class RpcApi {
 
   @Override
   public String toString() {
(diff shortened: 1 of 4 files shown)


ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/RpcApi.java
```diff
@@ -45,6 +44,6 @@ public int hashCode() {
 
   @Override
   public String toString() {
-    return MoreObjects.toStringHelper(this).add("cliValue", cliValue).toString();
+    return cliValue;
```
ajsutton: I think this is probably a place where a comment is worth while explaining that the CLI help output uses `toString()` as it's very unexpected otherwise (the test will pick it up but nice to get feedback sooner than that).

PR #657, opened 2019-01-24
[NC-1344] Create a simple WorldStateDownloader

mbaxter:
## PR description
This PR contains a basic algorithm for downloading world state from the network along with supporting changes.  This is just the first step in implementing the `WorldStateDownloader`.  Additional functionality will need to be implemented in subsequent PRs including:
- A real queue implementation that handles serialization / deserialization of data persisted to disk
- Cancellation functionality
- Handling stalled downloads
- Periodic commits of downloaded world state


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/storage/keyvalue/KeyValueStorageWorldStateStorage.java
```diff
@@ -29,23 +32,46 @@ public KeyValueStorageWorldStateStorage(final KeyValueStorage keyValueStorage) {
   }
 
   @Override
-  public Optional<BytesValue> getCode(final Hash codeHash) {
-    return keyValueStorage.get(codeHash);
+  public Optional<BytesValue> getCode(final Bytes32 codeHash) {
```
mbaxter: Some of these methods use `Hash` and some use `Bytes32` - just changed them all to `Bytes32` since that is more generic.

mbaxter: It's also kind of annoying to have to wrap `Bytes32` in `Hash` all over.  Kind of wondering if we really need the `Hash` type ...

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/worldstate/DefaultMutableWorldState.java
```diff
@@ -49,7 +49,7 @@
   private final WorldStateStorage worldStateStorage;
 
   public DefaultMutableWorldState(final WorldStateStorage storage) {
-    this(MerklePatriciaTrie.EMPTY_TRIE_ROOT_HASH, storage);
+    this(MerklePatriciaTrie.EMPTY_TRIE_NODE_HASH, storage);
```
mbaxter: There's nothing special about a root node, so updated this constant name so that it can be used more generally.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/storage/keyvalue/KeyValueStorageWorldStateStorage.java
```diff
@@ -29,23 +32,46 @@ public KeyValueStorageWorldStateStorage(final KeyValueStorage keyValueStorage) {
...
+    if (hash.equals(Hash.EMPTY)) {
+      return Optional.of(BytesValue.EMPTY);
+    } else {
+      return getter.apply(hash);
+    }
+  }
+
+  private Optional<BytesValue> getValue(
+      final Bytes32 hash, final Function<Bytes32, Optional<BytesValue>> getter) {
+    return getTrieValue(hash, (h) -> getCodeValue(h, getter));
```
ajsutton: I think we've wound up applying the check for empty twice in the `getNodeData` path - once in `WorldStateArchive` and once here.  This is probably the better place, but I actually prefer the more explicit code from `WorldStateArchive` - I know it essentially duplicates some code but explicitly checking for EMPTY and EMPTY_TRIE_NODE_HASH is just a bit more readable than delegating through optionals.

I'm not sure it matters, but it also avoids creating an extra object instances that this lambda version requires (to create a closure over the `getter` param).

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/storage/keyvalue/KeyValueStorageWorldStateStorage.java
ajsutton: I like `Hash` providing a very clear type to say this is a hash, not just an arbitrary 32 bytes but I don't think it has to be a hard and fast "every hash must use the `Hash` type" kind of thing.  Particularly in this low level kind of place where the meaning is unambiguous using `Bytes32` makes sense to me.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/worldstate/WorldStateDownloader.java
```diff
@@ -0,0 +1,199 @@
...
+          sendAndProcessRequests(peer, toRequest)
+              .whenComplete(
+                  (res, error) -> {
+                    if (outstandingRequests.decrementAndGet() == 0 && pendingRequests.isEmpty()) {
+                      // We're done
+                      worldStateStorageUpdater.commit();
+                      future.complete(null);
+                    } else {
+                      // Send out additional requests
+                      requestNodeData();
```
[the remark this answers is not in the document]
ajsutton: So it does. I wonder if it's worth extracting a method for the send single request function to make that while a bit easier to spot (though sometimes there's just no helping fools like me...).

Line 128 above:
`if (outstandingRequests.decrementAndGet() == 0 && pendingRequests.isEmpty()) {`
will then depend on `isEmpty()` being thread safe and completely accurate.  For the in memory implementation it would be enough to delegate to `ConccurrentLinkedQueue.isEmpty()` rather than using the separately tracked `size` variable I think.  We know we're the last request to finish so there's nothing else coming in at the same time, but we need to be sure that any pending requests added by the request prior to us have actually been reflected in the `pendingRequests` properly.

services/queue/src/main/java/tech/pegasys/pantheon/services/queue/InMemoryBigQueue.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+package tech.pegasys.pantheon.services.queue;
+
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.atomic.AtomicLong;
+
+public class InMemoryBigQueue<T> implements BigQueue<T> {
+  private final AtomicLong size = new AtomicLong(0);
+  private final Queue<T> internalQueue = new ConcurrentLinkedQueue<>();
```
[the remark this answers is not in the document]
mbaxter: Removed the separate size tracking.  I _think_ that the way `isEmpty` is being used in the downloader should be okay because the flow is:  
- send request and increment outstandingRequest count
- receive response and process it
  - queue additional requests
- decrement outstandingRequests and check if the queue is empty

Assuming that once requests are queued (enqueue returns), an updated count should be visible to other threads, I think that flow works.  Does that seem right to you?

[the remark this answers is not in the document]
ajsutton: As long as the outstanding request count is incremented *before* the request is actually sent (which it was from memory) then yes I agree that works.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/storage/keyvalue/KeyValueStorageWorldStateStorage.java
mbaxter: Fair enough - updated

PR #658, opened 2019-01-25
[NC-2137] Start world downloader

ajsutton:
## PR description
Extends the fast sync download process to include starting off the (currently unimplemented) world state and chain downloads. If either fails the other is cancelled and the overall fast sync process doesn't complete until both are complete.


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/fastsync/WorldStateDownloader.java
```diff
@@ -0,0 +1,21 @@
...
+ * specific language governing permissions and limitations under the License.
+ */
+package tech.pegasys.pantheon.ethereum.eth.sync.fastsync;
+
+import tech.pegasys.pantheon.ethereum.core.BlockHeader;
+
+import java.util.concurrent.CompletableFuture;
+
+public interface WorldStateDownloader {
+  CompletableFuture<Void> downloadWorldState(BlockHeader pivotBlockHeader);
```
ajsutton: @mbaxter this is the API I'd ideally like for the world state downloader which doesn't quite match #657 where the pivot block header is passed into the constructor.  There are two ways around this which both seem reasonable:

1. Make this interface act as a factory that creates a new instance of the world state downloader as it exists
2. Change the world state downloader run method to take the pivot block header

The main thing is to ensure we don't wind up with two concurrent world state downloads which is maybe slightly more straight forward with option 2, but the way `FastSyncDownloader` manages the returned futures should take care of it anyway.

mbaxter: I like option #2 - that probably make more sense than fixing the header in the constructor.

PR #659, opened 2019-01-25
[NC-2138] Extract out generic parts of Downloader

ajsutton:
## PR description
Extract out the reusable parts of the chain download process into a separate class with pluggable implementations for finding the sync target and creating the block import task. This provides the foundation for adding a fast sync chain downloader.


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/ChainDownloader.java
```diff
@@ -98,11 +95,16 @@ public void start() {
     }
   }
 
+  @VisibleForTesting
+  public CompletableFuture<?> getCurrentTask() {
```
mbaxter: Can we make this package-private?

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/SyncTargetManager.java
```diff
@@ -0,0 +1,123 @@
...
+                syncTargetDisconnectListenerId =
+                    bestPeer.subscribeDisconnect(this::onSyncTargetPeerDisconnect);
+                return CompletableFuture.completedFuture(syncTarget);
+              });
+    } else {
+      return waitForPeerAndThenSetSyncTarget();
+    }
+  }
+
+  protected abstract Optional<EthPeer> selectBestAvailableSyncTarget();
```
mbaxter: This difference here for fast sync is an extra check that the sync target agrees the pivot block is on the canonical chain?

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/ChainDownloader.java
ajsutton: Sadly no because `FullSyncDownloader` is under `...sync.fullsync` package and `FastSyncDownloader` will be under `...sync.fullsync` so this lives under `...sync`.  I think it's with this bit of pain to be able to have separate packages for fast sync vs full sync specific bits.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/SyncTargetManager.java
ajsutton: Yep exactly.

PR #661, opened 2019-01-25
NC-2120 --discovery-enabled option refactoring

NicolasMassart:
## PR description
now able to have --discovery-enabled option with true as default
and --discovery-enabled=true or --discovery-enabled=false on CLI
and discovery-enabled=true or discovery-enabled=false in YAML config file.

## Fixed Issue(s)

fixes NC-2120


docs/Configuring-Pantheon/Networking.md
```diff
@@ -36,7 +36,9 @@ Trailing peers cannot be used to get new blocks and are more likely to be reques
 
 ## No Discovery
 
-The [`--no-discovery`](../Reference/Pantheon-CLI-Syntax.md#no-discovery) command line option disables P2P peer discovery.  Only use this option if you are running a test node or a test network with fixed nodes.
+The [`--rpc-ws-enabled`](../Reference/Pantheon-CLI-Syntax.md#rpc-ws-enabled) command line option
```
MadelineMurray: Should rpc-ws-enabled be discover-enabled?

docs/Configuring-Pantheon/Networking.md
```diff
@@ -36,7 +36,9 @@ Trailing peers cannot be used to get new blocks and are more likely to be reques
 
 ## No Discovery
 
-The [`--no-discovery`](../Reference/Pantheon-CLI-Syntax.md#no-discovery) command line option disables P2P peer discovery.  Only use this option if you are running a test node or a test network with fixed nodes.
+The [`--rpc-ws-enabled`](../Reference/Pantheon-CLI-Syntax.md#rpc-ws-enabled) command line option 
+enables P2P peer discovery.
+Only set this option to `false` if you are running a test node or a test network with fixed nodes.
```
MadelineMurray: 'Only set this option to' -> 'Set to' -> uses less words and still has the same meaning

docs/Reference/Pantheon-CLI-Syntax.md
```diff
@@ -388,18 +388,18 @@ P2P network identifier.
 This option can be used to override your current network ID.
 The default value is the current network chain ID which is defined in the genesis file.
 
-### no-discovery
+### discovery-enabled
 
 ```bash tab="Syntax"
---no-discovery
+--discovery-enabled=false
```
MadelineMurray: For p2p-enabled I specified the syntax as `--p2p-enabled=<true|false>` because you can specify true.  I think either way is fine but we should make them the same so just let me know what you end up doing.

PR #663, opened 2019-01-25
Fixed Difficulty calculator

rojotek:
Fixed difficulty calculator. 
Goal is to have a simple replacement in genesis blocks for devmode options currently floating around.


pantheon/src/main/java/tech/pegasys/pantheon/controller/PantheonController.java
```diff
@@ -53,7 +55,16 @@
 
     final GenesisConfigOptions configOptions = genesisConfigFile.getConfigOptions();
 
-    if (configOptions.isEthHash()) {
+    if (FixedDifficultyCalculators.isFixedDifficultyInConfig(configOptions)) {
```
rain-on: is it possible to prevent FixedDifficulty from leaking this far?
I.e. this level just checks ibft vs clique vs ethhash - the ProtocolSchedule is where we check literal config options for the consensus mechanism.
(I say this not knowing how hard this would be to do in the ProtocolSchedule (i also appreciate the mainnetprotocolschedule is a bad name)).

rojotek: Moved to MainnetPantheonController and it feels better there.  Thanks for the feedback.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/mining/FixedMiningDifficultyAcceptanceTest.java
```diff
@@ -0,0 +1,53 @@
...
+package tech.pegasys.pantheon.tests.acceptance.mining;
+
+import tech.pegasys.pantheon.tests.acceptance.dsl.AcceptanceTestBase;
+import tech.pegasys.pantheon.tests.acceptance.dsl.account.Account;
+import tech.pegasys.pantheon.tests.acceptance.dsl.node.Node;
+
+import org.junit.Before;
+import org.junit.Test;
+
+public class FixedMiningDifficultyAcceptanceTest extends AcceptanceTestBase {
```
ajsutton: I think this can just be removed now that dev mode is using the fixed difficulty calculator - pretty much every AT is using it and mining transactions so if it ever breaks they'll all grind to a halt because of the increased difficulty.

acceptance-tests/src/test/resources/fixed/fixed.json
```diff
@@ -0,0 +1,40 @@
+{
```
ajsutton: Can just use `dev.json` now.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/factory/PantheonNodeFactory.java
```diff
@@ -120,6 +121,24 @@ public PantheonNode createArchiveNodeWithP2pDisabled(final String name) throws I
             .build());
   }
 
+  public Node createFixedDifficultyMinerNode(final String name) throws IOException {
```
ajsutton: `--network DEV` will be enough to get this now.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/mining/FixedMiningDifficultyAcceptanceTest.java
rojotek: removed

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/factory/PantheonNodeFactory.java
rojotek: or I could remove the method since fixed difficulty is being used in tests by default and the fixed difficulty AT has been deleted.

acceptance-tests/src/test/resources/fixed/fixed.json
rojotek: 👍

PR #669, opened 2019-01-27
Added privacy CLI options 

MadelineMurray:
Added privacy CLI options and note privacy not available until v1.1


docs/Reference/Pantheon-CLI-Syntax.md
```diff
@@ -510,6 +510,60 @@ The default is 30303.
 !!!note
     This option is not used when running Pantheon from the [Docker image](../Getting-Started/Run-Docker-Image.md#exposing-ports). 
 
+### privacy-enabled
+
+```bash tab="Syntax"
+--privacy-enabled=<true|false>
```
NicolasMassart: value is optionnal and without value, will swithc the boolean. So as default is false, if passed `--privacy-enabled` without value it will make it true. Then the following would be better IMO:
```suggestion
--privacy-enabled[=<true|false>]
```

docs/Reference/Pantheon-CLI-Syntax.md
```diff
@@ -510,6 +510,60 @@ The default is 30303.
...
+
+```bash tab="Example Configuration File"
+privacy-enabled=false
+```
+
+Set to enable private transactions. 
+The default is false.
+
+!!!note
+    Privacy is under development and will be available in v1.1.
```
NicolasMassart: If not available in this version, why documenting it ? Does it mean it works partially ? if yes, what part can be used ? Otherwise, we should hide the options in the help and in the doc.

PR #678, opened 2019-01-28
Metrics Push Gateway Options

shemnon:
## PR description

Don't overload `--metrics-host` and `--metrics-port` for push mode,
instead use `--metrics-push-host` and `--metrics-push-port`.  The former
is inbound, the latter is outbound.


pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -674,6 +692,8 @@ MetricsConfiguration metricsConfiguration() {
     metricsConfiguration.setMode(metricsMode);
     metricsConfiguration.setHost(metricsHost.toString());
     metricsConfiguration.setPort(metricsPort);
+    metricsConfiguration.setPushHost(metricsPushHost.toString());
+    metricsConfiguration.setPushPort(metricsPushPort);
```
mbaxter: Is there a way to throw an error if the user configures a set of host / port values that don't correspond to the chosen mode?

shemnon: picoCLI doesn't let us see what CLI flags there are, just the resulting values and a default is indistinguishable from a flag with a default value.

shemnon: So, no.  Not with our current CLI/TOML library.

NicolasMassart: Will be possible using what's in #679

NicolasMassart:
I can't if using push enabled instead of a mode option is better for users. Otherwise the code looks correct given this initial choice.

shemnon:
Adding new rules to #679

PR #679, opened 2019-01-28
NC-1737 Enabled warning on CLI dependent options

NicolasMassart:
## PR description
This mechanism is aimed at preventing the use of options if the associated
service is disabled. This is not a generic system as our need is simple here.
But in future version of PicoCLI some options dependency mechanism may be
implemented that could replace this.
See https://github.com/remkop/picocli/issues/295
Our current mechanism apply to p2p, rpc (http+ws), mining, privacy and metrics.

Also renamed inconsistent isRpcHttpEnabled option variable and moved
--privacy-enabled option to be on top of other privacy options.

Related tests where added or updated and renamed.
## Fixed Issue(s)

fixes NC-1737


jakehaugen:
If I'm thinking about this correctly, the goal of this change is to make it clear to the user what options are dependent on each other. The end goal is to avoid people assuming something is "on" because they've configured it. We could accomplish this by allowing them to configure the options but show a "warning" that they will have no impact until the feature is enabled. So, essentially change this from an "error" to a "warning". Does that solve our problem by communicating the dependency but still allowing things to be temporarily disabled/enabled in situations like @ajsutton called out?

ajsutton:
Yep, logging a warning makes a lot of sense to me.

pantheon/src/main/java/tech/pegasys/pantheon/cli/CommandLineUtils.java
```diff
@@ -0,0 +1,68 @@
...
+                  option ->
+                      Arrays.stream(option.names()).anyMatch(dependentOptionsNames::contains)
+                          && !option.stringValues().isEmpty())
+              .map(option -> option.names()[0])
+              .collect(
+                  Collectors.collectingAndThen(
+                      Collectors.toList(), StringUtils.joiningWithLastDelimiter(", ", " and ")));
+
+      if (!affectedOptions.isEmpty()) {
+        logger.warn("{} require {} option to be true.", affectedOptions, mainOptionName);
```
shemnon: I think the verbiage could be better.  Perhaps Madeline or Jake could weigh in.
```suggestion
        logger.warn("{} will have no effect unless {} is in the command line.", affectedOptions, mainOptionName);
```

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -630,6 +680,19 @@ private WebSocketConfiguration webSocketConfiguration() {
   }
 
   MetricsConfiguration metricsConfiguration() {
+
+    CommandLineUtils.checkOptionDependencies(
```
shemnon: This changed in #678 - here's the new rules

```suggestion
    CommandLineUtils.checkOptionDependencies(
        logger,
        commandLine,
        "--metrics-enabled",
        !isMetricsEnabled,
        Arrays.asList(
            "--metrics-mode",
            "--metrics-host",
            "--metrics-port"));
            
    CommandLineUtils.checkOptionDependencies(
        logger,
        commandLine,
        "--metrics-push-enabled",
        !isMetricsEnabled,
        Arrays.asList(
            "--metrics-push-host",
            "--metrics-push-port",
            "--metrics-push-interval",
            "--metrics-push-prometheus-job"));
```

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -654,6 +704,35 @@ MetricsConfiguration metricsConfiguration() {
...
+        logger,
+        commandLine,
+        "--metrics-push-enabled",
+        !isMetricsPushEnabled,
+        Arrays.asList(
+            "--metrics-push-host",
+            "--metrics-push-port",
+            "--metrics-push-interval",
+            "--metrics-push-prometheus-job"));
+        logger,
```
shemnon: github suggestions left in lines 727 to 735.   Looks like they don't support multiline suggestions or I did my comment wrong.

NicolasMassart: I fixed it by hand but yes, I think it means that it's single line only.

PR #689, opened 2019-01-29
NC-2043: Toml authentication provider

lucassaldanha:
## PR description

- Creates Toml Vert.x auth provider
- Add CLI option to output hashed password in the cmd line


ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/authentication/TomlUser.java
```diff
@@ -0,0 +1,86 @@
...
+  public void setAuthProvider(final AuthProvider authProvider) {
+    // we only use Toml for authentication
+    throw new UnsupportedOperationException("Not implemented");
+  }
+
+  @Override
+  protected void doIsPermitted(
+      final String permission, final Handler<AsyncResult<Boolean>> resultHandler) {
+    // we only use Toml for authentication
+    throw new UnsupportedOperationException("Not implemented");
```
Errorific: Given we have groups and permissions in the toml user & file we should be able to do is permitted

PR #696, opened 2019-01-29
[PRIV] Implement privacy precompiled contract

Puneetha17:
## PR description

Implement Privacy Precompiled Contract to call the Orion `receive` API.


ethereum/core/build.gradle
```diff
@@ -55,6 +55,8 @@ dependencies {
   integrationTestImplementation 'junit:junit'
   integrationTestImplementation 'org.assertj:assertj-core'
   integrationTestImplementation 'org.mockito:mockito-core'
+  integrationTestImplementation project(':testutil')
+  integrationTestImplementation 'net.consensys:orion:0.1.0-SNAPSHOT'
```
lucassaldanha: All dependencies versions are being set in the `gradle/versions.gradle` file. Can we do the same with this dependency?

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/precompiles/privacy/PrivacyPrecompiledContract.java
```diff
@@ -12,27 +12,57 @@
...
 
   @Override
   public BytesValue compute(final BytesValue input) {
+    try {
+      String key = new String(input.extractArray(), UTF_8);
+      ReceiveRequest receiveRequest = new ReceiveRequest(key, orionPublicKey);
+      ReceiveResponse receiveResponse = orion.receive(receiveRequest);
+      LOG.info("Got the response as ", receiveResponse.getPayload());
+      return BytesValue.wrap(receiveResponse.getPayload());
+      // pass it to private tx processor
```
lucassaldanha: Did we forget to implement something or is this comment here as a reminder to do something in the future?

Puneetha17: Its a reminder to implement in the future.

PR #697, opened 2019-01-29
Metrics Service configuration incorrect

shemnon:
## PR description

--metrics-push-enabled is not eabling metrics push and --metrics-enabled
is enabling push instead of polling.


metrics/src/main/java/tech/pegasys/pantheon/metrics/prometheus/MetricsService.java
```diff
@@ -26,9 +26,9 @@ static MetricsService create(
       final MetricsConfiguration configuration,
       final MetricsSystem metricsSystem) {
     if (configuration.isEnabled()) {
```
mbaxter: Nit - can we clarify the names around this config (perhaps in another PR)?  It's not very clear what `isEnabled` refers to, and it's also not clear how that relates to `isPushEnabled`.  It might be enough to change `isEnabled` to something like `isHttpServiceEnabled`, `isPullEnabled`, or something similar ...

shemnon: The names map directly to the --metrics-* fields, isEnabled is --metrics-enabled and isPushEnabled maps to --metrics-push-enabled.  IMHO any change to the name would need to be reflected in the CLI, or this would be the only special case.  If anything we could make two separate configs.

PR #698, opened 2019-01-29
[NC-2207] Add a basic queue on top of RocksDB

mbaxter:
## PR description
Add a basic queue implementation on top of RocksDB.


services/queue/src/main/java/tech/pegasys/pantheon/services/queue/RocksDbQueue.java
```diff
@@ -0,0 +1,154 @@
...
+    try {
+      loadNativeLibrary();
+      options =
+          new Options()
+              .setCreateIfMissing(true)
+              // TODO: Support restoration from a previously persisted queue
+              .setErrorIfExists(true);
+      db = RocksDB.open(options, storageDirectory.toString());
+
+      final String classLabelName = "class";
```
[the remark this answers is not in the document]
mbaxter: The thought was that this implementation is pretty basic and we may follow-up with something better to compare against.  Thinking it through though, I guess we could just check the before / after metrics for that change without having distinct labels.

services/util/src/main/java/tech/pegasys/pantheon/services/util/RocksDbUtil.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+package tech.pegasys.pantheon.services.util;
+
+import tech.pegasys.pantheon.util.InvalidConfigurationException;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.rocksdb.RocksDB;
+
+public class RocksDbUtil {
```
mbaxter: I created this helper package for common code within services (which is for now just this rocksdb init code).  WDYT @ajsutton ?

ajsutton: Yep that works for me.

PR #704, opened 2019-01-30
Limit ibft msg queues

jframe:
## PR description
Add a hard cap to the number unprocessed messages in the event queue and the seen messages tracked.


config/src/test/java/tech/pegasys/pantheon/config/IbftConfigOptionsTest.java
```diff
@@ -26,6 +26,7 @@
   private static final int EXPECTED_DEFAULT_EPOCH_LENGTH = 30_000;
   private static final int EXPECTED_DEFAULT_BLOCK_PERIOD = 1;
   private static final int EXPECTED_DEFAULT_REQUEST_TIMEOUT = 1;
+  private static final int EXPECTED_DEFAULT_MESSAGE_BUFFER_SIZE = 10_000;
```
rain-on: Are these oddly named? I.e. what does "EXPECTED_" indicate?

consensus/ibft/src/integration-test/java/tech/pegasys/pantheon/consensus/ibft/support/TestContextBuilder.java
```diff
@@ -158,7 +159,8 @@ public TestContext build() {
     // Use a stubbed version of the multicaster, to prevent creating PeerConnections etc.
     final StubValidatorMulticaster multicaster = new StubValidatorMulticaster();
 
-    final UniqueMessageMulticaster uniqueMulticaster = new UniqueMessageMulticaster(multicaster);
+    final UniqueMessageMulticaster uniqueMulticaster =
+        new UniqueMessageMulticaster(multicaster, MESSAGE_BUFFER_SIZE);
```
rain-on: should Multicaster be updated as part of this change? Or is that a separate problem?
Is the size of the multicaster buffer related to the incoming event queue?

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/IbftBlockHeightManager.java
```diff
@@ -64,7 +64,7 @@
   private final BlockTimer blockTimer;
   private final IbftMessageTransmitter transmitter;
   private final MessageFactory messageFactory;
-  private final Map<Integer, RoundState> futureRoundStateBuffer = Maps.newHashMap();
+  private final Map<Integer, RoundState> futureRoundStateBuffer;
```
rain-on: is this related to the incoming event queue size?

consensus/ibft/src/integration-test/java/tech/pegasys/pantheon/consensus/ibft/support/TestContextBuilder.java
jframe: The multicaster should be updated to have a limited size buffer. Would be better to have a seperate config option though.

config/src/main/java/tech/pegasys/pantheon/config/IbftConfigOptions.java
```diff
@@ -39,4 +41,12 @@ public int getBlockPeriodSeconds() {
   public int getRequestTimeoutSeconds() {
     return ibftConfigRoot.getInteger("requesttimeoutseconds", DEFAULT_ROUND_EXPIRY_SECONDS);
   }
+
+  public int getSeenMessageQueueSize() {
```
CjHare: s/getSeenMessageQueueSize/getSeenMessageQueueLimit

QueueSize could be mistaken as being the current queue size

jframe:
> Please don't use `size` when you mean `limit`. Consider common use of `size` in places such as the Collection classes

Intent was to be the max size

config/src/main/java/tech/pegasys/pantheon/config/IbftConfigOptions.java
jframe: I like limit better. Still don't like these config names though regardless of size or limit suffix. So if have any suggestions...

config/src/main/java/tech/pegasys/pantheon/config/IbftConfigOptions.java
```diff
@@ -39,4 +41,12 @@ public int getBlockPeriodSeconds() {
...
+    return ibftConfigRoot.getInteger("seenmessagequeuesize", DEFAULT_SEEN_MESSAGES_QUEUE_SIZE);
+  }
+
+  public int getEventQueueSize() {
```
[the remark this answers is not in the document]
jframe: done. changed eventQueueSize -> messageQueueLimit and seenMessageQueueSize -> gossipedHistoryLimit

config/src/test/java/tech/pegasys/pantheon/config/IbftConfigOptionsTest.java
jframe: They are bit oddly named I guess. The "EXPECTED_" indicates this expected values for the assertion. These values are used to assert the default values are correct in the tests. It's existing test code so I just followed the existing pattern.

config/src/main/java/tech/pegasys/pantheon/config/IbftConfigOptions.java
```diff
@@ -21,6 +21,8 @@
   private static final long DEFAULT_EPOCH_LENGTH = 30_000;
   private static final int DEFAULT_BLOCK_PERIOD_SECONDS = 1;
   private static final int DEFAULT_ROUND_EXPIRY_SECONDS = 1;
+  private static final int DEFAULT_GOSSIPED_HISTORY_LIMIT = 10_000;
```
rain-on: this FEELS like a huge number ... in a good system, should this not be about ... 100?

jframe: 10,000 is probably large than it needs to be. It will however only take a small amount of memory as we are only storing the hashes. So I don't think we need to be too exact and can be a little generous.
Perhaps 1000 is a more appropriate default? I think this would handle to tracking the message hashes of few rounds. And a size of 100 would only just cover the total number of messages for the current round if we going to use 20 validators.

ajsutton: Might be worth thinking about if there's some metrics you could capture to help find the right value for this setting (even if you need to add some expensive ones temporarily). ie every time you find a message you've already gossiped, record how many messages are recorded after it in your history. That will likely have to be a temporary thing because you'll need to use a less efficient structure to hold the history (worst case a list you iterate from most recent to least recent), but it would give you some concrete data to drive decisions from.

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/IbftEventQueue.java
```diff
@@ -26,16 +26,20 @@
 public class IbftEventQueue {
   private final BlockingQueue<IbftEvent> queue = new LinkedBlockingQueue<>();
 
-  private static final int MAX_QUEUE_SIZE = 1000;
   private static final Logger LOG = LogManager.getLogger();
+  private final int maxQueueSize;
```
CjHare: What happened to using `queueLimit`?

jframe: It's still there. I missed updating the existing field name in the IbftEventQueue.

jframe: updated to make it consistent with the new config names

config/src/main/java/tech/pegasys/pantheon/config/IbftConfigOptions.java
jframe: We had some further discussion and settled on 1000 for now as this will cover the current height with several rounds.
Would be great to get some metrics on this. And would also allow users to tune this setting if they wanted to.

PR #716, opened 2019-01-30
CLI refactoring to make host options able to accept localhost strings

shemnon:
## PR description

Stop using Guava's HostAndPort and HostSpecifier and just use Strings.


pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -517,6 +510,7 @@ public void parse(
     commandLine = new CommandLine(this);
 
     commandLine.setCaseInsensitiveEnumValuesAllowed(true);
+    //    commandLine.setToggleBooleanFlags(false);
```
NicolasMassart: Can you remove this commented line?

shemnon: Done

shemnon:
We were not doing any validation on the host side when it was host:port, so that isn't a change in functionality.  We could move to InetAddress as the field as it will validate IP4 and IP6 addresses and resolve textual addresses.

shemnon:
I put the localhost test on p2p host.  Should I do the test for all host arguments?

shemnon:
Looked at InteAddress, I don't think we should use it.  It initializes from either an ip address or hostname, but we cannot easily get back what was passed in.  It has to be an IP address or a hostname, not whatever was specified at CLI, so we lose information.  With CORS headers this could be a huge thing.

shemnon:
IPv4 tests are implicit in the host/port used test.  added localhost and ipv6 tests.

PR #717, opened 2019-01-30
Change Görli config to final launch config

ajsutton:
## PR description
Görli will be rebooted and launched as a stable testnet. This updates the config to match updated config at https://github.com/goerli/testnet


5chdn:
Thanks!

PR #722, opened 2019-01-31
Rename ibftrevised to ibft2

jframe:
## PR description
Rename IBFT 2.0 config section from revisedibft to ibft2.


config/src/main/java/tech/pegasys/pantheon/config/GenesisConfigOptions.java
```diff
@@ -21,15 +21,15 @@
 
   boolean isIbft();
 
-  boolean isRevisedIbft();
+  boolean isIbft2();
```
rain-on: don't love that this is only 1 character different from isIbft

config/src/main/java/tech/pegasys/pantheon/config/GenesisConfigOptions.java
```diff
@@ -21,15 +21,15 @@
...
 
   boolean isClique();
 
   IbftConfigOptions getIbftConfigOptions();
 
   CliqueConfigOptions getCliqueConfigOptions();
 
-  IbftConfigOptions getRevisedIbftConfigOptions();
+  IbftConfigOptions getIbft2ConfigOptions();
```
rain-on: as above.

config/src/main/java/tech/pegasys/pantheon/config/JsonGenesisConfigOptions.java
```diff
@@ -21,7 +21,7 @@
 
   private static final String ETHASH_CONFIG_KEY = "ethash";
   private static final String IBFT_CONFIG_KEY = "ibft";
-  private static final String REVISED_IBFT_CONFIG_KEY = "revisedibft";
+  private static final String REVISED_IBFT_CONFIG_KEY = "ibft2";
```
rain-on: probably should change the constant name (replace "Revised" with something else)

jframe: yep the constant should renamed as well

jframe: done

config/src/main/java/tech/pegasys/pantheon/config/GenesisConfigOptions.java
jframe: Not sure what to do about that. The name is ibft 2. Thought perhaps could make the old version ibftLegacy in the method names?

jframe: done. renamed old ibft methods to ibftLegacy to distinguish from ibft2

config/src/main/java/tech/pegasys/pantheon/config/GenesisConfigOptions.java
jframe: done. as above

PR #729, opened 2019-01-31
[NC-2119] permissioning config path CLI

macfarla:
## PR description
add CLI for --permissions-config-path which overrides the default value

## Fixed Issue(s)
NC-2119


pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -630,6 +637,13 @@ private void ensureAllBootnodesAreInWhitelist(
     }
   }
 
+  private String getPermissionsConfigPath() {
```
mark-terry: This doesn't feel right.  We're just after a String that contains the path, yeah?  Could we just concatenate the absolute path from dataDir() and the DefaultCommandValues field?

pantheon/src/test/java/tech/pegasys/pantheon/cli/PantheonCommandTest.java
```diff
@@ -351,6 +386,58 @@ public void tomlThatConfiguresEverythingExceptPermissioning() throws IOException
     assertThat(options.stream().map(CommandLine.Model.OptionSpec::longestName)).isEmpty();
   }
 
+  @Test
+  public void tomlThatConfiguresEverythingPlusPermissioning() throws IOException {
+    assumeTrue(isFullInstantiation());
+
+    // Load a TOML that configures literally everything (except permissioning)
```
mark-terry: Could bits of the `tomlThatConfiguresEverythingExceptPermissioning` method be extracted to avoid dupe here?

macfarla: actually examining how the spec options work, this method doesn't test anything extra. removed.

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
macfarla: to concatenate them, you need to insert the file separator which is platform dependent.

PR #735, opened 2019-02-01
IPV6 bootnodes don't work

shemnon:
## PR description

The IPV6 URL syntax for IPv6 addresses has the addresses enclosed in brackets.
Without the brackets the address could consume the port number.


ajsutton:
Is there a test we can add to ensure we don't break IPv6 syntax in the future? Might be able to do it in the `PantheonCommandTest` to ensure it's parsed and accepted at least.

shemnon:
The converter tests can pick apart what we need.  The URI wasn't being parsed out w/o the brackets, so I made sure we checked the URI parts.

PR #741, opened 2019-02-01
Handle validation failures when fast importing blocks

ajsutton:
## PR description
When a fast import block fails validation we need to halt the import and retry. Previously we incorrectly ignored the failure and continued trying to import future blocks (which would also fail validation because the parent doesn't exist).


ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/sync/fastsync/FastSyncBlockHandlerTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+    final CompletableFuture<List<BlockWithReceipts>> result =
+        blockHandler.validateAndImportBlocks(blocksWithReceipts);
+
+    assertThat(result).isCompleted();
+    verify(blockImporter).fastImportBlock(protocolContext, BLOCK, emptyList(), LIGHT_SKIP_DETACHED);
+  }
+
+  @Test
+  public void shouldReturnExceptionallyCompletedFutureWhenBlockImportFails() {
```
mbaxter: (optional) Might also be good to have a test that imports a sequence of blocks and fails early in that sequence. Then we can check that `fastImportBlock` doesn't get called for the blocks after the failure.

ajsutton: Good idea.  Done.

PR #742, opened 2019-02-01
Update RoundChangeManager to use flattened message

rain-on:


consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/RoundChangeManager.java
```diff
@@ -65,7 +64,12 @@ public boolean roundChangeReady() {
     public RoundChangeCertificate createRoundChangeCertificate() {
       if (roundChangeReady()) {
         actioned = true;
-        return new RoundChangeCertificate(receivedMessages.values());
+        return new RoundChangeCertificate(
+            receivedMessages
+                .values()
+                .stream()
+                .map(message -> message.getSignedPayload())
```
jframe: nit: can use method reference

PR #746, opened 2019-02-01
[NC-2207] Integrate rocksdb-based queue into WorldStateDownloader

mbaxter:
## PR description
Integrate rocksdb-based queue into WorldStateDownloader.


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/DefaultSynchronizer.java
```diff
@@ -114,6 +134,22 @@ public void start() {
     }
   }
 
+  private Path getStateQueueDirectory(final Path dataDirectory) {
+    Path queueDataDirectory = dataDirectory.resolve("fastsync/statequeue");
+    File queueDataFile = queueDataDirectory.toFile();
+    queueDataFile.mkdirs();
+    // Clean up this data for now (until fast sync resume functionality is in place)
+    queueDataFile.deleteOnExit();
```
ajsutton: I don't think `deleteOnExit` will work for a directory that we'll wind up creating files inside.  The naming of `queueDataDirectory` and `queueDataFile` is a bit misleading here as well since they're both actually a directory just of type `Path` or `File`.  Not quite sure what the right naming is but I think I'd have `queueDataDirectory` and `queueDataDir` since they're the same thing in slightly different forms.

mbaxter: Yeah, you're right - this won't clean up non-empty directories.  I'm still testing this locally, so I'll clean it up a bit and ping you when its ready to look at.

mbaxter: Added a shutdown hook for the cleanup: https://github.com/PegaSysEng/pantheon/pull/746/commits/bf2306fd9e6b9ed95a60f7c332affdca3afe4676

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/DefaultSynchronizer.java
```diff
@@ -98,9 +116,25 @@ public DefaultSynchronizer(
...
+                () -> {
+                  try {
+                    // Clean up this data for now (until fast sync resume functionality is in place)
+                    MoreFiles.deleteRecursively(
+                        stateQueueDirectory, RecursiveDeleteOption.ALLOW_INSECURE);
+                  } catch (IOException e) {
+                    LOG.error("Unable to clean up fast sync files: {}", stateQueueDirectory, e);
+                  }
+                }));
+  }
```
ajsutton: It would probably be better to put this in a `stop()` method which is called from `Runner.close`. Pantheon already sets up a shutdown hook to call `Runner.close` and then when tests shutdown `Runner` without killing the whole JVM this clean up will run then and not be delayed.

ajsutton: Also, I can't see where the Queue gets closed. It didn't matter with the in memory implementation but we do need to ensure that rocksdb is closed.  This may be a good place (though we'd need to ensure `WorldStateDownloader` is stopped and not going to try to access the queue anymore.

mbaxter: Good feedback!  Pushed the cleanup into a `stop()` method here: https://github.com/PegaSysEng/pantheon/pull/746/commits/f3d4e817885923a522d5c36c99f085677e88fb71

ajsutton: Yep, that all LGTM.  Merge away. :)

PR #747, opened 2019-02-02
[FIX] Refactored permissioning components to be Optional.

mark-terry:
## PR description
The Permissioning functionality is optional, so the implementation has been changed to reflect this.


acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/ThreadPantheonNodeRunner.java
```diff
@@ -93,7 +94,7 @@ public void startNode(final PantheonNode node) {
             .bannedNodeIds(Collections.emptySet())
             .metricsSystem(noOpMetricsSystem)
             .metricsConfiguration(node.metricsConfiguration())
-            .permissioningConfiguration(node.getPermissioningConfiguration())
+            .permissioningConfiguration(Optional.of(node.getPermissioningConfiguration()))
```
Errorific: Prefer to see the RunnerBuilder default to having an empty permissioning config and this setter be passed a PermissioningConfig without the Optional and pivot whether we call this based on whether we have a config or not.

mark-terry: A good suggestion.  Will update.

pantheon/src/main/java/tech/pegasys/pantheon/RunnerBuilder.java
```diff
@@ -88,7 +88,7 @@
   private Collection<String> bannedNodeIds;
   private MetricsConfiguration metricsConfiguration;
   private MetricsSystem metricsSystem;
-  private PermissioningConfiguration permissioningConfiguration;
+  private Optional<PermissioningConfiguration> permissioningConfiguration;
```
Errorific: As per earlier comment, initialise this to empty and fill it if the setter is called. This also avoids potential null pointers in build()

pantheon/src/main/java/tech/pegasys/pantheon/RunnerBuilder.java
```diff
@@ -141,7 +141,7 @@ public RunnerBuilder webSocketConfiguration(final WebSocketConfiguration webSock
   }
 
   public RunnerBuilder permissioningConfiguration(
-      final PermissioningConfiguration permissioningConfiguration) {
+      final Optional<PermissioningConfiguration> permissioningConfiguration) {
```
Errorific: as per earlier comments, make this take a PermissioningConfiguration rather than an Optional

Issue #662, opened 2019-01-25
Support contract code in the genesis file


JohnnySheffield:
### Description
As an clique node operator, I want to be able to connect to and sync with Geth in a network with custom genesis.json where a contract is defined in the genesis.

### Acceptance Criteria
* contract defined in genesis is created and available
* geth and pantheon calculate the same genesis hash

### Steps to Reproduce (Bug)
1. start pantheon with custom genesis as defined here: https://github.com/hpoa/lisinski/blob/master/geth/lisinski.genesis
2. add one of the available geth peers through rpc or as bootnodes:
enode://92435c3c0ecc81acecb59a1a108f6a9266e4569331b7cddc511ab060a9a0a3dc4993a2f175643166f88b0dcbc09b138fd8bffc26eaac1681f71b107fa8c1f52f@142.93.230.46:30303
or
enode://367cece61c3d48ccef6fa71694936ea5263069726a5799d47bd27a45fb160f3db0c8e949341dff54f8ffccae8befd5bfdb25fa8516ed498e7bef1af1f214fb6f@31.147.205.39:30303
3.  set logging level to debug
4. Observe "Disconnecting from peer with matching network id (385), but non-matching genesis hash" error in the logs.

Full logs here: https://gist.github.com/JohnnySheffield/9ada92c057053413635996b472b7fcf4

**Expected behavior:** Expect calculated genesis hash to be the same, as the genesis.json file provided is the same

**Actual behavior:** Calculated genesis hash is different

### Versions (Add all that apply)
docker image pulled from pegasyseng/pantheon:latest, at the time of pull latest is 0.8.5

* Java version: [`java -version`]
openjdk version "1.8.0_181"
OpenJDK Runtime Environment (IcedTea 3.9.0) (Alpine 8.181.13-r0)
OpenJDK 64-Bit Server VM (build 25.181-b13, mixed mode)
* OS Name & Version: [`cat /etc/*release`]
3.8.1
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.8.1
PRETTY_NAME="Alpine Linux v3.8"
HOME_URL="http://alpinelinux.org"
BUG_REPORT_URL="http://bugs.alpinelinux.org"
* Kernel Version: [`uname -a`]
Linux pantheon-89-xkdr8 3.10.0-862.6.3.el7.x86_64 #1 SMP Fri Jun 15 17:57:37 EDT 2018 x86_64 

### Additional Information
[...]

PR #749, opened 2019-02-03
Support genesis contract code

kziemianek:
## PR description
Support for contract code read from genesis file.

## Fixed Issue(s)
fixes: #662


ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/chain/GenesisStateTest.java
```diff
@@ -85,6 +85,19 @@ public void createFromJsonNoAllocs() throws Exception {
     assertThat(header.getParentHash()).isEqualTo(Hash.ZERO);
   }
 
+  @Test
+  public void createFromJsonWithContract() throws Exception {
```
kziemianek: I might remove this test and edit `createFromJsonWithAllocs` but i left it here in order to prove that it passes the case described in issue.

ajsutton: Yeah it could go either way but I think it's worth having a test that covers this more complex case specifically.  I've added a few assertions for the contract account's code and storage for good measure.

PR #757, opened 2019-02-04
Flatten IBFT Message API

rain-on:
The IBFT messages have been flattened such that the content within the signed subtype can be accessed directly from the message interface.


consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/messagewrappers/NewRound.java
```diff
@@ -24,10 +29,31 @@ public NewRound(final SignedData<NewRoundPayload> payload) {
...
+  }
+
+  public Block getBlock() {
+    return getProposalPayload().getPayload().getBlock();
+  }
+
+  @Override
+  public BytesValue encode() {
+    final BytesValueRLPOutput rlpOut = new BytesValueRLPOutput();
+    rlpOut.startList();
```
jframe: Why are these encoded inside an additional list whilst some others also with a single e.g. Commit are writing the output as is?

PR #763, opened 2019-02-04
[NC-1968] Permissioning whitelist persistence.

mark-terry:
## PR description
Implemented functionality to persist permissioning whitelists across restarts.

## Fixed Issue(s)

Fixes #NC-1968


acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/condition/perm/WhiteListContainsKeyAndValue.java
```diff
@@ -0,0 +1,44 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.stream.Stream;
+
+public class WhiteListContainsKeyAndValue implements Condition {
+  private final String val;
+  private final Path tempFile;
+
+  public WhiteListContainsKeyAndValue(final String val, final Path tempFile) {
```
lucassaldanha: Instead of passing the expected string, why can't we use TOML parsing logic inside the `verify()` and use a Map<String, Object> as val here?
It will make the test syntax easier to understand and less error-prone.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/permissioning/WhitelistPersistorAcceptanceTest.java
```diff
@@ -0,0 +1,102 @@
...
+  public void manipulatedAccountsWhitelistIsPersisted() {
+    node.verify(
+        perm.expectPermissioningWhitelistFileKeyValue(
+            String.format("%s=[\"%s\"]", "accounts-whitelist", senderA.getAddress()), tempFile));
+
+    node.execute(transactions.addAccountsToWhitelist(senderB.getAddress()));
+    node.verify(perm.expectAccountsWhitelist(senderA.getAddress(), senderB.getAddress()));
+    node.verify(
+        perm.expectPermissioningWhitelistFileKeyValue(
+            String.format(
```
lucassaldanha: This is an example where we would benefit from the Map<String, Object> approach.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/permissioning/PermAddAccountsToWhitelist.java
```diff
@@ -64,6 +64,8 @@ public JsonRpcResponse response(final JsonRpcRequest request) {
         case ERROR_DUPLICATED_ENTRY:
           return new JsonRpcErrorResponse(
               request.getId(), JsonRpcError.ACCOUNT_WHITELIST_DUPLICATED_ENTRY);
+        case ERROR_WHITELIST_PERSIST_FAIL:
+          return new JsonRpcErrorResponse(request.getId(), JsonRpcError.WHITELIST_PERSIST_FAILURE);
```
lucassaldanha: I feel like this type of error is too fine grained. Assuming we verify at startup that the Pantheon process has to write access to the permissions file, we should assume that any writing errors will be unrecoverable.

That means that we can let it fail as an internal error or something instead of adding one more error message. Thoughts?

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/permissioning/NodeWhitelistController.java
```diff
@@ -45,45 +59,65 @@ private boolean removeNode(final Peer node) {
...
       }
     }
+
+    final List<Peer> oldWhitelist = new ArrayList<>(this.nodesWhitelist);
+
     peers.forEach(this::removeNode);
+    try {
+      updateConfigurationFile(peerToEnodeURI(nodesWhitelist));
+    } catch (IOException e) {
+      revertState(oldWhitelist);
```
lucassaldanha: I think that we aren't ensuring that writing to the file is an atomic operation. In this case, I believe the best thing would be comparing the in-memory old list with the latest list read from the file to verify that they are the same. If they are not, it means that the file has been compromised.

ethereum/permissioning/src/main/java/tech/pegasys/pantheon/ethereum/permissioning/WhitelistPersistor.java
```diff
@@ -0,0 +1,93 @@
...
+
+  public synchronized void updateConfig(
+      final WHITELIST_TYPE whitelistType, final Collection<String> updatedWhitelistValues)
+      throws IOException {
+    removeExistingConfigItem(whitelistType);
+    addNewConfigItem(whitelistType, updatedWhitelistValues);
+  }
+
+  @VisibleForTesting
+  public void removeExistingConfigItem(final WHITELIST_TYPE whitelistType) throws IOException {
```
lucassaldanha: Is this method handling TOML files that define the whitelist in a multiline array?
e.g.

```
"nodes-whitelist"= [
      "foo",
      "bar"
]
```

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/permissioning/PermAddAccountsToWhitelist.java
mark-terry: Makes sense, it's hard to think of a likely case where a node will be able to recover from a state where it can't write to the file.

ethereum/permissioning/src/main/java/tech/pegasys/pantheon/ethereum/permissioning/WhitelistPersistor.java
mark-terry: Not at all.  Definitely needs that.  👍

ethereum/permissioning/src/main/java/tech/pegasys/pantheon/ethereum/permissioning/WhitelistPersistor.java
```diff
@@ -0,0 +1,93 @@
...
+    List<String> otherConfigItems;
+    try (Stream<String> configKeys = Files.lines(configurationFile.toPath())) {
+      otherConfigItems =
+          configKeys
+              .filter(line -> !line.contains(whitelistType.getTomlKey()))
+              .collect(Collectors.toList());
+    }
+
+    Files.write(
```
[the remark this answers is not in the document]
mark-terry: Updating the PR with checks for this case. 👍

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/condition/perm/WhiteListContainsKeyAndValue.java
mark-terry: Done.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/permissioning/PermAddAccountsToWhitelist.java
mark-terry: I've left this in because:
- I've added read + write checks on the file when we validate the CLI config
- We've added another condition for when the file is out of sync, but decided to allow the client to continue in this case
- This case is different to the previous case ^

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/permissioning/NodeWhitelistController.java
mark-terry: Updated.

ethereum/permissioning/src/main/java/tech/pegasys/pantheon/ethereum/permissioning/WhitelistPersistor.java
mark-terry: This has been implemented.

ethereum/permissioning/src/main/java/tech/pegasys/pantheon/ethereum/permissioning/WhitelistPersistor.java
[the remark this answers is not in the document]
mark-terry: PR has been updated to include config file state verification, which verifies the before and after state and will (at least attempt to) revert to the before state if the operation fails.

PR #767, opened 2019-02-05
EthScheduler additions

shemnon:
## PR description

Add the services thread pool and a computation thread pool to the
EthScheduler.

* Services are long running, sequential, and infrequently start tasks
  such as Full Sync and Fast Sync.
* Computations are short and high CPU intensity tasks such as ECDSA
  signature extractions and POW validation.  The intent is that each
  runnable represents one such extraction and the extractions from a
  block are saturated across available processing power.  These
  computations should have zero dependencies outside their object and
  thread.

### DIFF
--- a/ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/EthScheduler.java
+++ b/ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/EthScheduler.java
@@ -44,8 +44,11 @@ public class EthScheduler {
   protected final ExecutorService syncWorkerExecutor;
   protected final ScheduledExecutorService scheduler;
   protected final ExecutorService txWorkerExecutor;
+  private final ExecutorService servicesExecutor;
+  private final ExecutorService computationExecutor;
 
-  EthScheduler(final int syncWorkerCount, final int txWorkerCount) {
+  EthScheduler(
+      final int syncWorkerCount, final int txWorkerCount, final int computationWorkerCount) {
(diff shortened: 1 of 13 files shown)


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/EthScheduler.java
```diff
@@ -194,25 +218,46 @@ public void stop() {
...
+      LOG.error("{} services executor did not shutdown cleanly.", this.getClass().getSimpleName());
+      servicesExecutor.shutdownNow();
+      servicesExecutor.awaitTermination(2L, TimeUnit.MINUTES);
+    }
+    if (!computationExecutor.awaitTermination(2L, TimeUnit.MINUTES)) {
+      LOG.error(
+          "{} computation executor did not shutdown cleanly.", this.getClass().getSimpleName());
+      computationExecutor.shutdownNow();
+      computationExecutor.awaitTermination(2L, TimeUnit.MINUTES);
+    }
```
ajsutton: This is starting to add up to a lot of waiting.  Maybe we should set it up so that we call `shutdown` on all of them, wait then `shutdownNow` on any not shutdown, then wait and log any that still failed to shutdown?

shemnon: sounds like a good first issue - https://github.com/PegaSysEng/pantheon/issues/768

PR #770, opened 2019-02-05
[NC-2233] Randomly perform full validation when fast syncing blocks

ajsutton:
## PR description
During a fast sync, most blocks skip proof of work validation to speed up import. However we randomly select blocks to perform full validation on so we have confidence it really is a valid chain.

There's a slight oddity here because of the way we've modelled header validation - we actually perform header validation in two separate places, when the header is downloaded (detached - parent may not have been imported yet) and when it's about to be imported (attached - it's parent block is already on our chain). This means we have two independent places where we decide on the type of validation. With this implementation, fast sync is randomly selecting light or full separately at both those points.  There is actually only ever one rule skipped by using light validation - `ProofOfWorkValidation` when using ethash or `IbftCommitSealsValidationRule` when using IBFT. `ProofOfWorkValidation` runs with the detached rules, `IbftCommitSealsValidationRule` runs with the attached rules. Net result, when using ethash the choice between full or light only makes a difference when running the detached rules and for IBFT it only makes a difference when running the attached rules.


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/ValidationPolicy.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package tech.pegasys.pantheon.ethereum.eth.sync;
+
+import tech.pegasys.pantheon.ethereum.mainnet.HeaderValidationMode;
+
+public interface ValidationPolicy {
```
mbaxter: Might be worth adding `@FunctionalInterface` here

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/sync/fastsync/FastSyncValidationPolicyTest.java
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  @Test
+  public void shouldEventuallyUseBothModesWhenValidationPolicyIsHalf() {
+    final FastSyncValidationPolicy policy =
+        new FastSyncValidationPolicy(0.5f, LIGHT, FULL, NO_OP_LABELLED_COUNTER);
+    boolean seenLight = false;
+    boolean seenFull = false;
+    // It's theoretically possible to flip a coin 2^31-1 times and only ever get heads but
+    // we're probably more likely to get intermittent failures from the hard drive failing...
```
mbaxter: 😂

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/ValidationPolicy.java
ajsutton: Good idea.

PR #772, opened 2019-02-05
[NC-2180] Wait for a peer with an estimated chain height before selecting a pivot block

ajsutton:
## PR description
When selecting a pivot block we need to ensure the estimated chain height for at least one peer is actually available.  This may fail because the chain header block hasn't yet been retrieved from the peer or because the peer connected, causing the wait for peer task to complete, but then disconnected before we could select a pivot block.  This change updates `FastSyncActions.selectPivotBlock` to ignore peers without a chain height estimate and retry if a peer isn't available.


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/tasks/PipelinedImportChainSegmentTask.java
```diff
@@ -233,25 +233,30 @@ public void completeChunkPipelineAndMaybeLaunchNextOne(
...
+              result.complete(headers.subList(1, headers.size()));
+            } else {
+              result.completeExceptionally(
+                  new InvalidBlockException(
+                      "Provided first header does not connect to last header.",
+                      parentHeader.getNumber(),
+                      parentHeader.getHash()));
+            }
+            return result;
+          } catch (Throwable t) {
```
mbaxter: Curious - why do you need this try / catch?

ajsutton: Oh we shouldn't need it and I've removed it again.  That said, somewhere we're swallowing exceptions silently so will need to track that down and fix separately.

PR #778, opened 2019-02-06
Track fast sync head as a separate property to chain head

ajsutton:
## PR description
During fast sync, the chain head remains at 0, but for each block imported via fast sync we update a new fast sync head property in the blockchain.  The actual chain head is then only updated when the pivot block is reached and world state is available, completing the fast sync process.

Two advantages of this:

1. As far as any other part of the system is concerned, the chain is still back at genesis so it won't try to access the blocks part way through fast sync or fail because the world state isn't available.
2. If Pantheon exists before fast sync completes, the chain is still "empty" so fast sync is allowed and it resumes downloading from the fast sync chain head (where it left off).


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/SynchronizerConfiguration.java
```diff
@@ -31,7 +31,7 @@
   // TODO: Determine reasonable defaults here
   public static final int DEFAULT_PIVOT_DISTANCE_FROM_HEAD = 500;
   public static final float DEFAULT_FULL_VALIDATION_RATE = .1f;
-  public static final int DEFAULT_FAST_SYNC_MINIMUM_PEERS = 5;
+  public static final int DEFAULT_FAST_SYNC_MINIMUM_PEERS = 1;
```
shemnon: Is this a dev setting?

ajsutton: dang it.  We really need to expose these as command line options so I stop forgetting to change that back...

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/chain/DefaultMutableBlockchain.java
```diff
@@ -192,18 +229,25 @@ private UInt256 calculateTotalDifficulty(final Block block) {
   private BlockAddedEvent updateCanonicalChainData(
       final BlockchainStorage.Updater updater,
       final Block newBlock,
-      final UInt256 totalDifficulty) {
+      final UInt256 totalDifficulty,
+      final boolean isFastSync) {
     final Hash chainHead = blockchainStorage.getChainHead().orElse(null);
     if (newBlock.getHeader().getNumber() != BlockHeader.GENESIS_BLOCK_NUMBER && chainHead == null) {
       throw new IllegalStateException("Blockchain is missing chain head.");
     }
```
mbaxter: I think you need a check somewhere to make sure we're not interleaving regular blocks with fast-sync blocks, which could leave the chain data in an inconsistent state.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/chain/DefaultMutableBlockchain.java
```diff
@@ -192,18 +229,25 @@ private UInt256 calculateTotalDifficulty(final Block block) {
...
     final Hash newBlockHash = newBlock.getHash();
     try {
-      if (chainHead == null || newBlock.getHeader().getParentHash().equals(chainHead)) {
+      if (chainHead == null
+          || newBlock.getHeader().getParentHash().equals(chainHead)
+          || isFastSync) {
         // This block advances the chain, update the chain head
         updater.putBlockHash(newBlock.getHeader().getNumber(), newBlockHash);
-        updater.setChainHead(newBlockHash);
+        if (isFastSync) {
```
mbaxter: You should check that the new fast-sync block is connected to the previous fast-sync block or chainhead.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/queries/BlockchainQueries.java
```diff
@@ -75,6 +75,10 @@ public long headBlockNumber() {
     return blockchain.getChainHeadBlockNumber();
   }
 
+  public Optional<Long> fastSyncHeadBlockNumber() {
```
mbaxter: I don't think this is used anywhere?  We should probably have a separate ticket for handling json-rpc requests - a bunch of those will need to be modified to handle fast sync.

ajsutton: True. Was used in a hacked-in JSON-RPC method I added initially so I could see what was happening.  Removed.

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/chain/DefaultMutableBlockchainTest.java
```diff
@@ -135,6 +135,57 @@ public void appendBlockWithMismatchedReceipts() {
...
+  public void fastAppendBlock() {
+    final BlockDataGenerator gen = new BlockDataGenerator();
+
+    final KeyValueStorage kvStore = new InMemoryKeyValueStorage();
+    final Block genesisBlock = gen.genesisBlock();
+    final DefaultMutableBlockchain blockchain = createBlockchain(kvStore, genesisBlock);
+
+    final BlockOptions options =
+        new BlockOptions().setBlockNumber(1L).setParentHash(genesisBlock.getHash());
+    final Block newBlock = gen.block(options);
```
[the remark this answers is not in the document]
ajsutton: Oh nice. Done.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/chain/DefaultMutableBlockchain.java
ajsutton: I think the code around this prevents that anyway but I've added an explicit check to help us sleep better at night.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/chain/DefaultMutableBlockchain.java
ajsutton: We actually need to support re-orgs on the fast sync chain because we may need to pick a different pivot block (especially at the moment where restarting will definitely do that).  I imagine it's extremely rare to get a re-org that far back but better to handle it cleanly.

It deliberately doesn't track forks in fast sync - whatever chain we're currently importing is the one that's considered canonical because that's the one that will get us to the pivot block and have a world state matched up.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/chain/DefaultMutableBlockchain.java
```diff
@@ -355,7 +363,17 @@ private BlockAddedEvent handleChainReorg(
         });
     clearIndexedTransactionsForBlock(updater, removedTransactions);
 
-    // Update tracked forks
+    if (!isFastSync) {
+      updateForksForReorg(updater, newChainHead, oldChainHead);
```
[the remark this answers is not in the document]
mbaxter: I _think_ we're tracking the fork heads so that we can go back and clean them up at some point.  In which case, it seems useful to keep that set accurate.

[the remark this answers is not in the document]
ajsutton: Yeah I was initially aiming to limit the number of headers we'd encounter without world states but #824 handles that case so probably is better to track forks made during fast sync too.

ajsutton:
Going to go with the approach in #836 rather than this so that the blockchain can remain agnostic to how blocks are being synced.

PR #783, opened 2019-02-06
Permissioning API methods

MadelineMurray:
Added permissioning API methods


docs/Reference/JSON-RPC-API-Methods.md
```diff
@@ -2191,4 +2191,193 @@ Proposes [adding or removing a validator](../Consensus-Protocols/IBFT.md#adding-
...
+    ```
+### perm_addNodesToWhitelist
+
+Adds nodes to the nodes whitelist. 
+
+**Parameters** 
+
+`list of strings` - List of enode URLs 
+
+The enode URL is `enode://<id>@<host:port>` where:
```
macfarla: is this enode explanation repeated elsewhere? Should it be made a separate thing that other pages can link to?

docs/Reference/JSON-RPC-API-Methods.md
```diff
@@ -2191,4 +2191,193 @@ Proposes [adding or removing a validator](../Consensus-Protocols/IBFT.md#adding-
...
+
+Removes accounts (participants) from the accounts whitelist. 
+
+**Parameters** 
+
+`list of strings` - List of account addresses 
+
+**Returns** 
+
+`result` - `Success` or `error`
```
macfarla: do we explain when you would expect an error?

docs/Reference/JSON-RPC-API-Methods.md
```diff
@@ -2191,4 +2191,193 @@ Proposes [adding or removing a validator](../Consensus-Protocols/IBFT.md#adding-
...
+
+!!! example
+    ```bash tab="curl HTTP request"
+    $ curl -X POST --data '{"jsonrpc":"2.0","method":"perm_addAccountsToWhitelist","params":[["0xb9b81ee349c3807e46bc71aa2632203c5b462032", "0xb9b81ee349c3807e46bc71aa2632203c5b462034"]], "id":1}' <JSON-RPC-http-endpoint:port>
+    ```
+    
+    ```bash tab="wscat WS request"
+    {"jsonrpc":"2.0","method":"perm_addAccountsToWhitelist","params":[["0xb9b81ee349c3807e46bc71aa2632203c5b462032", "0xb9b81ee349c3807e46bc71aa2632203c5b462034"]], "id":1}
```
NicolasMassart: Explain why we need double square brackets because it's not obvious, I first thought it was a typo, but then tested and understood that the params list contains a param that is a list even if this param is the only one in the data list. If we miss the list in the list, it returns a message like `{"jsonrpc":"2.0","id":1,"error":{"code":-32603,"message":"Internal error"}}`

docs/Reference/JSON-RPC-API-Methods.md
MadelineMurray: Have added to the Node Keys page and referenced.  Removed it from the couple of other places it appeared and added the reference there too.

docs/Reference/JSON-RPC-API-Methods.md
MadelineMurray: Added

docs/Reference/JSON-RPC-API-Methods.md
MadelineMurray: Added an explanation. You're right - this caught me the first time as well.

PR #804, opened 2019-02-07
Updated for fixeddifficulty in genesis file

MadelineMurray:
Updated private network tutorial to use fixeddifficulty so it will mine blocks at a reasonable rate


docs/Tutorials/Create-Private-Network.md
```diff
@@ -56,6 +57,7 @@ Copy the following genesis definition to a file called `privateNetworkGenesis.js
 {
   "config": {
```
NicolasMassart: Shouldn't the genesis contain a chain id parameter ? also if it has one, setting the network id in the cli would be useless as the default network id is the genesis chain id.

docs/Tutorials/Create-Private-Network.md
```diff
@@ -56,6 +57,7 @@ Copy the following genesis definition to a file called `privateNetworkGenesis.js
...
     "ethash": {
+      "fixeddifficulty": 1000
```
NicolasMassart: Why 1000 ? why not 100 like in development network dev.json ?

docs/Tutorials/Create-Private-Network.md
MadelineMurray: Updated

docs/Tutorials/Create-Private-Network.md
MadelineMurray: As discussed - not quite as fast so you can see transactions going into the blocks.

PR #808, opened 2019-02-07
[NC-2058] Improve block propagation time

s-matthew-english:
## PR description

Addresses feedback raised in previous round of review and contains a test that verifies the invocation of the newly added `broadcastBlock()` method

**NOTE**: 

what do you make of the passage of  `BlockBroadcaster` in the `DefaultSynchronizer`? 

**Next Step:**

add some tests that verify that the propagated  block is in fact being received by peers


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/EthPeer.java
```diff
@@ -130,6 +133,15 @@ public ResponseStream send(final MessageData messageData) throws PeerNotConnecte
     }
   }
 
+  public void propagateBlock(final Block block, final UInt256 totalDifficulty) {
+    final NewBlockMessage newBlockMessage = NewBlockMessage.create(block, totalDifficulty);
+    final Capability capability = Capability.create("eth", 63);
```
ajsutton: We don't want to create a new capability here and need to use the passed in `protocolName`.  I suspect `connection.sendForProtocol(protocolName, newBlockMessage)` is what you want.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockPropagationManager.java
```diff
@@ -144,12 +144,21 @@ private void onBlockAdded(final BlockAddedEvent blockAddedEvent, final Blockchai
     }
   }
 
-  private void handleNewBlockFromNetwork(final EthMessage message) {
+  void broadcastBlock(final Block block, final UInt256 difficulty) {
```
ajsutton: Is there a reason this is package private instead of private?

I'd also suggest as you move forward to extract this out into a separate `BlockBroadcaster` class. It just provides a better separation of concerns and will let us write a more focussed unit test for the approach to broadcasting blocks.  Happy enough for it to live here for now though.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockPropagationManager.java
```diff
@@ -144,12 +144,21 @@ private void onBlockAdded(final BlockAddedEvent blockAddedEvent, final Blockchai
...
+    final List<EthPeer> availablePeers =
+        ethContext.getEthPeers().availablePeers().collect(Collectors.toList());
+    for (EthPeer ethPeer : availablePeers) {
+      ethPeer.propagateBlock(block, difficulty);
+    }
+  }
```
CjHare: If you're feeling adventurous you could merge the collection into a list and forLoop into a single line: 
`ethContext.getEthPeers().availablePeers().forEach(ethPeer -> ethPeer.propagateBlock(block, difficulty));`

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockBroadcaster.java
```diff
@@ -0,0 +1,71 @@
...
+
+  @VisibleForTesting
+  void broadcastBlock(final Block block, final UInt256 difficulty) {
+    ethContext
+        .getEthPeers()
+        .availablePeers()
+        .forEach(ethPeer -> ethPeer.propagateBlock(block, difficulty));
+  }
+
+  void effectuateBroadcast() {
```
shemnon: _optional_ Too many syllables.  Perhaps `maybeBroadcast` or `possiblyBroadcast.` . Effectuate communicates to me that the block will unconditionally be broadcast, while what we need to communicate is that if it passes some tests it will be broadcast.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockPropagationManager.java
```diff
@@ -77,12 +78,14 @@
       final EthContext ethContext,
       final SyncState syncState,
       final PendingBlocks pendingBlocks,
-      final LabelledMetric<OperationTimer> ethTasksTimer) {
+      final LabelledMetric<OperationTimer> ethTasksTimer,
+      final BlockBroadcaster<C> blockBroadcaster) {
```
shemnon: This field is always overwritten, no need for it in the constructor.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockBroadcaster.java
```diff
@@ -0,0 +1,71 @@
...
+import tech.pegasys.pantheon.ethereum.mainnet.ProtocolSchedule;
+import tech.pegasys.pantheon.ethereum.mainnet.ProtocolSpec;
+import tech.pegasys.pantheon.util.uint.UInt256;
+
+import java.util.Optional;
+
+import com.google.common.annotations.VisibleForTesting;
+
+class BlockBroadcaster<C> {
+
```
shemnon: No real state is kept.  Could this be rewitten as a static function?  Or perhaps make block a parameter in `effectuateBroadcast`, but re-use gets tricky when the schedule changes.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockPropagationManager.java
```diff
@@ -63,6 +63,7 @@
   private final EthContext ethContext;
   private final SyncState syncState;
   private final LabelledMetric<OperationTimer> ethTasksTimer;
+  private BlockBroadcaster<C> blockBroadcaster;
```
shemnon: This field is always overwritten prior to use. Consider making it a variable instead.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockPropagationManagerTest.java
```diff
@@ -75,6 +75,7 @@ public static void setupSuite() {
     fullBlockchain = BlockchainSetupUtil.forTesting().importAllBlocks();
   }
 
+  @SuppressWarnings("unchecked")
```
shemnon: I really don't like disabling the unchecked warning.  Consider using <?> when you access no templated methods.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockPropagationManager.java
s-matthew-english: as I wrote in the description, it's easier to test if it's in the constructor. do you suggest making an explicit getter and setter?

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockPropagationManagerTest.java
s-matthew-english: `@SuppressWarnings("unchecked")` is already used in this class in `shouldNotImportBlocksThatAreAlreadyBeingImported`

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockBroadcaster.java
s-matthew-english: I'm not sure how to interpret this comment

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockPropagationManagerTest.java
```diff
@@ -458,11 +461,13 @@ public void ignoresOldNewBlockAnnouncement() {
     assertThat(blockchain.contains(oldBlock.getHash())).isFalse();
   }
 
+  @SuppressWarnings("unchecked")
   @Test
   public void purgesOldBlocks() {
     final int oldBlocksToImport = 3;
     syncConfig =
         SynchronizerConfiguration.builder().blockPropagationRange(-oldBlocksToImport, 5).build();
+    BlockBroadcaster<Void> blockBroadcaster = mock(BlockBroadcaster.class);
```
ajsutton: We should be able to make the `blockBroadcaster` from `setUp` a field and just re-use it here.  Which then avoids the need for the `@SuppressWarnings("unchecked")` on this method.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockPropagationManager.java
shemnon: The fact it is overwritten is the real concern and never used.  If it's fixed later then it can stay.

s-matthew-english: yeah- good point about it getting overwritten- mock passed to constructor never getting used 😬- so then we should have a getter and setter then?

ajsutton: No, passing it into the constructor is good, but `BlockBroadcaster` shouldn't take `Bock` into its constructor, only pass in the block to broadcast to the method that actual broadcasts it.  That way you can reuse the same `BlockBroadcaster` instance to broadcast multiple blocks and never need to create a new one in `BlockPropagationManager`.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockPropagationManager.java
```diff
@@ -63,6 +66,7 @@
   private final EthContext ethContext;
   private final SyncState syncState;
   private final LabelledMetric<OperationTimer> ethTasksTimer;
+  private BlockBroadcaster blockBroadcaster;
```
ajsutton: nit: this should be final

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockPropagationManager.java
```diff
@@ -105,9 +110,34 @@ private void setupListeners() {
         .subscribe(EthPV62.NEW_BLOCK_HASHES, this::handleNewBlockHashesFromNetwork);
   }
 
+  @VisibleForTesting
+  void broadcastBlock(final Block block) {
+    blockBroadcaster = new BlockBroadcaster(ethContext);
```
ajsutton: Don't create a new `BlockBroadcaster` here, just use the one passed into the constructor.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockPropagationManager.java
```diff
@@ -105,9 +110,34 @@ private void setupListeners() {
...
+    final BlockHeader parent =
+        protocolContext
+            .getBlockchain()
+            .getBlockHeader(block.getHeader().getParentHash())
+            .orElseThrow(
+                () ->
+                    new IllegalArgumentException(
+                        "Incapable of retrieving header from non-existent parent of "
+                            + block.getHeader().getNumber()
+                            + "."));
```
ajsutton: This is ok for this PR, but could you follow up with a PR to change `importOrSavePendingBlock` so that instead of just using `protocolContext.getBlockchain().contains(block.getHeader().getParentHash())` it actually gets the parent header and checks if it's present.  Then you can pass that header into `broadcastBlock` and not need this exception.  Just be careful that the lookup of the parent header stays inside the `synchronized` block.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockPropagationManager.java
```diff
@@ -105,9 +110,34 @@ private void setupListeners() {
         .subscribe(EthPV62.NEW_BLOCK_HASHES, this::handleNewBlockHashesFromNetwork);
   }
 
+  @VisibleForTesting
+  void broadcastBlock(final Block block) {
```
ajsutton: Since the validation is being done in here this should probably be called `validateAndBroadcastBlock`.  And hopefully we don't need it to be `@VisibleForTesting` - it's a code smell that tests need access to private internals of the class being tested.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockPropagationManager.java
```diff
@@ -105,9 +110,34 @@ private void setupListeners() {
...
+      final UInt256 totalDifficulty =
+          parent.getDifficulty().plus(block.getHeader().getDifficulty());
+      blockBroadcaster.propagate(block, totalDifficulty);
+    }
+  }
+
   private void onBlockAdded(final BlockAddedEvent blockAddedEvent, final Blockchain blockchain) {
     // Check to see if any of our pending blocks are now ready for import
     final Block newBlock = blockAddedEvent.getBlock();
+    broadcastBlock(newBlock);
```
ajsutton: This isn't the right place to trigger block propagation.  It will be called after we've processed transactions and imported to the chain - we want to do it before then, so in `importOrSavePendingBlock` just before the `PersistBlockTask`.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockPropagationManagerTest.java
```diff
@@ -561,4 +566,37 @@ public void shouldNotImportBlocksThatAreAlreadyBeingImported() {
 
     verify(ethScheduler, times(1)).scheduleSyncWorkerTask(any(Supplier.class));
   }
+
+  @Test
+  public void verifyBroadcastBlockInvocation() {
+    final BlockPropagationManager<Void> blockPropagationManager =
+        spy(
```
ajsutton: I'd suggest using a mock here instead of a spy.  Spies are generally overly complicated and we don't need the real block propagation - just want to check that the broadcast method was called.

s-matthew-english: why do you say they're overly complicated? if I change that to mock everything breaks in a confusing way

s-matthew-english: something related to that generic component

ajsutton: Because a spy is a combination of a mock and a real implementation.  So you're simultaneously saying this thing is within the scope of the unit I'm testing (thus using the real one) but also that it's outside of the unit (thus I want to check how it's interacted with).

And the voodoo required to make a spy work leads to a number of weird implications (e.g. that test you and I hit a few weeks back where moving a method reference to the constructor broke the tests because the spy no longer picked it up).

So for this test you don't want to spy on the `BlockPropagationManager`, you already have a mock `BlockBroadcaster` that you can assert on:

```


  @Test
  public void verifyBroadcastBlockInvocation() {
    blockchainUtil.importFirstBlocks(2);
    final Block block = blockchainUtil.getBlock(2);
    blockPropagationManager.start();

    // Setup peer and messages
    final RespondingEthPeer peer = EthProtocolManagerTestUtil.createPeer(ethProtocolManager, 0);

    final UInt256 totalDifficulty = fullBlockchain.getTotalDifficultyByHash(block.getHash()).get();
    final NewBlockMessage newBlockMessage = NewBlockMessage.create(block, totalDifficulty);

    // Broadcast message
    EthProtocolManagerTestUtil.broadcastMessage(ethProtocolManager, peer, newBlockMessage);

    final Responder responder = RespondingEthPeer.blockchainResponder(fullBlockchain);
    peer.respondWhile(responder, peer::hasOutstandingRequests);

    verify(blockBroadcaster, times(1)).propagate(block, totalDifficulty);
  }
```

And that test will fail because the total difficulty doesn't match.  The test is right, the code is wrong. :)  Total difficulty is being calculated as:
```
final UInt256 totalDifficulty =
          parent.getDifficulty().plus(block.getHeader().getDifficulty());
```
which is just parent difficulty + child difficulty when you need the sum of the difficulty for every block in the chain to get the total difficulty.  `final UInt256 totalDifficulty = protocolContext.getBlockchain().getTotalDifficultyByHash(parent.getHash()).get().plus(block.getHeader().getDifficulty());` will get you what you want.

s-matthew-english: interesting. so that's just the difficulty of that one parent block in isolation, the parent block doesn't incorporate the difficulty of the blocks that came before it. && more interesting still -- the genesis block has an actual difficulty. anyway - thanks for pointing this out.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockBroadcaster.java
```diff
@@ -0,0 +1,33 @@
...
+  BlockBroadcaster(final EthContext ethContext) {
+    this.ethContext = ethContext;
+  }
+
+  void propagate(final Block block, final UInt256 difficulty) {
+    ethContext
+        .getEthPeers()
+        .availablePeers()
+        .forEach(ethPeer -> ethPeer.propagateBlock(block, difficulty));
```
ajsutton: We'll need to filter out peers that have already seen this block:
`.filter(ethPeer -> ~ethPeer.hasSeenBlock(block.getHash()))`
and then should add a test for that as well.

ajsutton: And in `EthPeer.propagateBlock` we should add the block to the `knownBlocks` set.

PR #815, opened 2019-02-08
[NC-2046] JWT authentication for JSON-RPC

macfarla:
## PR description
check authentication and permissions for JSON-RPC methods

## Fixed Issue(s)
NC-2046


acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/PantheonNode.java
```diff
@@ -90,6 +90,8 @@
   private HttpRequestFactory httpRequestFactory;
   private Optional<EthNetworkConfig> ethNetworkConfig = Optional.empty();
   private boolean useWsForJsonRpc = false;
+  private boolean useAuthenticationTokenInHeaderForJsonRpc = false;
```
Errorific: Can probably simplify this boolean out by either making token an Optional<String> or checking token != null

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/PantheonNode.java
```diff
@@ -255,6 +261,22 @@ public void useWebSocketsForJsonRpc() {
     useWsForJsonRpc = true;
   }
 
+  /** All future JSON-RPC calls will include the authentication token. */
+  @Override
+  public void useAuthenticationTokenInHeaderForJsonRpc(final String token) {
+
+    if (jsonRequestFactories != null) {
+      jsonRequestFactories.shutdown();
```
Errorific: need to add `jsonRequestFactoies = null;` here, it's also wrong in useWebSocketsForJsonRpc but we haven't managed to race condition it yet.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/jsonrpc/HttpServiceLoginAcceptanceTest.java
```diff
@@ -22,12 +25,17 @@
 import org.junit.Test;
 
 public class HttpServiceLoginAcceptanceTest extends AcceptanceTestBase {
+  private Cluster authenticatedCluster;
   private Node node;
 
   @Before
   public void setUp() throws IOException, URISyntaxException {
+    final ClusterConfiguration clusterConfiguration =
+        new ClusterConfigurationBuilder().setAwaitPeerDiscovery(false).build();
```
Errorific: With the await peer discovery being off is there still something that is awaiting the node being ready? Want to avoid the login call further down occurring before the service has finished starting in the node.

macfarla: there is nothing awaiting the node being ready. If you can't do any JSON RPC calls, is there another way to ask the node if it's ready?

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/PantheonNode.java
macfarla: 👍

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/PantheonNode.java
macfarla: 👍

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/jsonrpc/HttpServiceLoginAcceptanceTest.java
macfarla: ^ if this is a problem, it's a problem for a few tests

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/PantheonNode.java
macfarla: This caused one AT to fail so putting it back in
java.lang.RuntimeException: org.web3j.protocol.exceptions.ClientConnectionException: Invalid response received: 400; Websocket endpoint can't handle HTTP requests

macfarla: RpcApisTogglesAcceptanceTest.shouldSucceedConnectingToNodeWithWsRpcEnabled

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/jsonrpc/HttpServiceLoginAcceptanceTest.java
lucassaldanha: If we want to test that the services are up (JSON-RPC service), can't we send a request to the 'login' route and expect any response?

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/JsonRpcMethod.java
```diff
@@ -31,4 +34,16 @@
    * @return output from applying the JSON-RPC method to the input.
    */
   JsonRpcResponse response(JsonRpcRequest request);
+
+  /**
+   * The list of Permissions that correspond to this JSON-RPC method. e.g. [net/*, net/listening]
+   *
+   * @return list of permissions that match this method.
+   */
+  default List<String> getPermissions() {
```
lucassaldanha: Have we considered supporting `*/*` as a valid permission config?

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/jsonrpc/HttpServiceLoginAcceptanceTest.java
macfarla: that's exactly what the test does. I think Chris is asking is there a chance that the login call happens before the node has started?

lucassaldanha: What I meant is that we could have a check (prior to the test) if the JSON-RPC service is responding. Currently, we use the admin_peers I guess, we could have an alternative (for when we disable the discovery check) that would send a request to the `login` route and expect it to be handled (it doesn't need to be a successful login, just need the request to be "received" by the node).

macfarla: gotcha. awaitLogin()

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/JsonRpcMethod.java
Errorific: kinda, we need to have a conversation about where and how we inflate the permissions list

macfarla: 👍

PR #818, opened 2019-02-08
Peer disconnects should not result in stack traces.

shemnon:
## PR description

Replace two known stack traces an info level.  Replace with a single
debug line because disconnects are basically expected.


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/ChainDownloader.java
```diff
@@ -105,6 +106,10 @@ public ChainDownloader(
                       LOG.trace("Download cancelled", t);
                     } else if (rootCause instanceof InvalidBlockException) {
                       LOG.debug("Invalid block downloaded", t);
+                    } else if (rootCause instanceof PeerDisconnectedException) {
```
mbaxter:
```suggestion
                    } else if (rootCause instanceof EthTaskException) {
                        LOG.debug(rootCause.getMessage());
                    }
```

mbaxter: While you're at it, you could probably just handle all `EthTaskException`'s ^ as those are "expected" exceptions.  It seems that `LOG.error` below should be there to catch unexpected exceptions.

shemnon: Yep.  We don't need stack traces when we fail in expected ways.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/ChainDownloader.java
```diff
@@ -105,6 +106,8 @@ public ChainDownloader(
                       LOG.trace("Download cancelled", t);
                     } else if (rootCause instanceof InvalidBlockException) {
                       LOG.debug("Invalid block downloaded", t);
+                    } else if (rootCause instanceof EthTaskException) {
+                      LOG.debug(rootCause.toString());
```
shemnon: toString is deliberate, to show we are not attempting to log the throwable.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/sync/fullsync/FullSyncDownloaderTest.java
```diff
@@ -550,7 +550,7 @@ public void recoversFromSyncTargetDisconnect() {
     // Downloader should recover and sync to next best peer, but it may stall
     // for 10 seconds first (by design).
     await()
-        .atMost(20, TimeUnit.SECONDS)
+        .atMost(30, TimeUnit.SECONDS)
```
mbaxter: Did you mean to include this change here?

PR #824, opened 2019-02-11
Handle unavailable world states

ajsutton:
## PR description 

Make `WorldStateArchive` return an `Optional<WorldState>` and explicitly track which world states are available.

Updates callers to at least reasonably handle cases where the world state they need is not available.


ethereum/core/src/test-support/java/tech/pegasys/pantheon/ethereum/core/ExecutionContextTestFixture.java
```diff
@@ -55,6 +55,7 @@ private ExecutionContextTestFixture(
             new NoOpMetricsSystem());
     this.stateArchive =
         new WorldStateArchive(new KeyValueStorageWorldStateStorage(keyValueStorage));
+    genesisState.writeStateTo(stateArchive.getMutable());
```
mbaxter: There's another call to `writeStateTo` a few lines down - looks like one of these should be deleted.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/storage/keyvalue/KeyValueStorageWorldStateStorage.java
```diff
@@ -12,17 +12,22 @@
...
 import tech.pegasys.pantheon.services.kvstore.KeyValueStorage;
 import tech.pegasys.pantheon.util.bytes.Bytes32;
 import tech.pegasys.pantheon.util.bytes.BytesValue;
+import tech.pegasys.pantheon.util.bytes.BytesValues;
 
 import java.util.Optional;
 
 public class KeyValueStorageWorldStateStorage implements WorldStateStorage {
 
+  private static final BytesValue AVAILABILITY_PREFIX = BytesValue.of("available".getBytes(UTF_8));
```
mbaxter: I was thinking we could handle this by flipping the approach and saving "provisional" state roots in a separate location.  For now, the only provisional root would be from fast sync.  When fast sync is done, we could then confirm that provisional state root and move it into regular storage.  The advantages of flipping the approach are: no need to migrate existing dbs and more efficient storage (no need to store extra data for every valid root).

ethereum/core/src/test-support/java/tech/pegasys/pantheon/ethereum/core/ExecutionContextTestFixture.java
ajsutton: So there is. Fixed.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/storage/keyvalue/KeyValueStorageWorldStateStorage.java
ajsutton: Switched to considering a state as available if the root node is present.  Fast sync delays committing the root node until the rest of the state has been downloaded.  Also realised that when the world state downloader is first started, rather than checking if the world state is empty, it can check if the world state it's after is available and skip downloading if it is (the empty world state is always available).

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/worldstate/WorldStateArchive.java
```diff
@@ -28,20 +28,23 @@ public WorldStateArchive(final WorldStateStorage storage) {
     this.storage = storage;
   }
 
-  public WorldState get(final Hash rootHash) {
-    return getMutable(rootHash);
+  public Optional<WorldState> get(final Hash rootHash) {
+    return getMutable(rootHash).map(state -> state);
```
mbaxter: Why is this `map` here?

ajsutton: To make the compiler happy - it changes the type from `Optional<MutableWorldState>` to `Optional<WorldState>`.  Since `MutableWorldState` is an instance of `WorldState` you don't need to change the object but `Optional<MutableWorldState>` is not an instance of `Optional<WorldState>` so you do need to do a mapping.  

I'm sure someone can insert an appropriate rant about the limitations of Java generics here. :)

PR #828, opened 2019-02-11
Extract creation and init of ProtocolContext into a re-usable class

ajsutton:
## PR description

Extracts creation of `BlockchainStorage`, `MutableBlockchain`, `WorldStateStorage`, `WorldStateArchive` and `ProtocolContext` into a reusable factory. Apart from removing duplication from the different `PantheonController` implementations, this will give us a single place to hook in data migrations that we anticipate will be required soon.


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/storage/DataStores.java
```diff
@@ -0,0 +1,70 @@
...
+
+public class DataStores<T> {
+
+  private final ProtocolContext<T> protocolContext;
+  private final WorldStateStorage worldStateStorage;
+
+  private DataStores(
+      final ProtocolContext<T> protocolContext, final WorldStateStorage worldStateStorage) {
+    this.protocolContext = protocolContext;
+    this.worldStateStorage = worldStateStorage;
```
mbaxter: `WorldStateStorage` is currently exposed from `WorldStateArchive.getStorage()`, so I think you actually only need to keep the `protocolContext` as an instance var.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/storage/DataStores.java
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  public static <T> DataStores<T> initDataStores(
```
mbaxter: (discussion) Wondering if this should move to `ProtocolContext.init` - since it seems to just deal with setting up that object?

ajsutton: Yes, that would work quite nicely now that you've pointed out the existence of `WorldStateArchive.getStorage`. I swear I went looking for that and it wasn't there yesterday.... :)

ajsutton: Done, much nicer now - just moving duplicated code into a `ProtocolContext.init`.

PR #833, opened 2019-02-11
Clique should use beneficiary of zero on epoch blocks

jframe:
## PR description
Also ensure on that on epoch block that a beneficiary of 0x0 is used.


consensus/clique/src/main/java/tech/pegasys/pantheon/consensus/clique/CliqueBlockInterface.java
```diff
@@ -59,15 +59,16 @@ public Address getProposerOfBlock(final BlockHeader header) {
 
   public static BlockHeaderBuilder insertVoteToHeaderBuilder(
       final BlockHeaderBuilder builder, final Optional<ValidatorVote> vote) {
+    final BlockHeaderBuilder voteHeaderBuilder = BlockHeaderBuilder.fromBuilder(builder);
```
rain-on: I know this was a bad function - but maybe "insert" doesn't make sense given we're not changing the passed in Builder - is it simply createBuilderFrom(x, y)?

jframe: I've gone with createHeaderBuilderWithVoteHeaders

PR #834, opened 2019-02-11
[NC-2118] Method to reload permissions file

lucassaldanha:
## PR description
 Implemented perm_reloadPermissionsFile API method that will trigger a reload of the permissions file, updating the accounts and nodes whitelists.


ethereum/jsonrpc/src/test/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/permissioning/PermReloadPermissionsFromFileTest.java
```diff
@@ -0,0 +1,91 @@
...
+            Optional.of(accountWhitelistController), Optional.of(nodeWhitelistController));
+  }
+
+  @Test
+  public void getNameShouldReturnExpectedName() {
+    assertThat(method.getName()).isEqualTo("perm_reloadPermissionsFromFile");
+  }
+
+  @Test
+  public void whenControllersAreNotPresentMethodShouldDoNothingAndReturnSuccess() {
```
mark-terry: Not sure about this one.  If both controllers aren't present, that should mean that permissioning hasn't been enabled and this method isn't available.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/permissioning/NodeWhitelistController.java
```diff
@@ -180,6 +191,29 @@ public boolean isPermitted(final Peer node) {
...
+              configuration.isAccountWhitelistEnabled());
+
+      readNodesFromConfig(updatedConfig);
+      this.configuration = updatedConfig;
+    } catch (Exception e) {
+      LOG.warn(
+          "Error reloading permissions file. In-memory whitelisted nodes will be reverted to previous valid configuration. "
+              + "Details: {}",
+          e.getMessage());
+      this.nodesWhitelist.addAll(currentAccountsList);
```
mark-terry: There'd be a chance that reading from config could fail mid-execution, yeah?  Should there be a `nodesWhitelist.clear();` call before the `addAll` ?

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/permissioning/NodeWhitelistController.java
```diff
@@ -180,6 +191,29 @@ public boolean isPermitted(final Peer node) {
...
+
+    try {
+      final PermissioningConfiguration updatedConfig =
+          PermissioningConfigurationBuilder.permissioningConfigurationFromToml(
+              configuration.getConfigurationFilePath(),
+              configuration.isNodeWhitelistEnabled(),
+              configuration.isAccountWhitelistEnabled());
+
+      readNodesFromConfig(updatedConfig);
+      this.configuration = updatedConfig;
```
mark-terry: It's probably an unlikely case, but should the config be reverted if the Exception is thrown?

ethereum/permissioning/src/main/java/tech/pegasys/pantheon/ethereum/permissioning/AccountWhitelistController.java
```diff
@@ -135,15 +146,37 @@ public boolean contains(final String account) {
...
+              configuration.isNodeWhitelistEnabled(),
+              configuration.isAccountWhitelistEnabled());
+      readAccountsFromConfig(updatedConfig);
+      this.configuration = updatedConfig;
+    } catch (Exception e) {
+      LOG.warn(
+          "Error reloading permissions file. In-memory whitelisted accounts will be reverted to previous valid configuration. "
+              + "Details: {}",
+          e.getMessage());
+      this.accountWhitelist.addAll(currentAccountsList);
```
mark-terry: See NodeWhitelistController comments.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/permissioning/NodeWhitelistController.java
lucassaldanha: We read the file and build the list before adding them to the controller's whitelist. When we call `readNodesFromConfig(updatedConfig);` the list has been completely read from the file.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/permissioning/NodeWhitelistController.java
lucassaldanha: We only update the config as the last line inside the try-catch block. Any exception thrown before that will result in the configuration object not being updated.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/permissioning/NodeWhitelistController.java
lucassaldanha: I guess we can have the clear() just in case :)

PR #836, opened 2019-02-12
Store pivot block header

ajsutton:
## PR description
When the pivot block header has been downloaded, store it to disk. Currently this is always removed on shutdown because the world state downloader can't currently handle resuming.  However, once that is complete and we stop deleting the stored header on shutdown:

 * Fast sync is enabled when requested if a fast sync is still in progress, even if the chain head is no longer genesis
 * `FastSyncActions` will use that same pivot block when restarting

Note that the validation which disabled fast sync when the chain wasn't empty has been moved out of `SynchronizerConfiguration` into the new `FastSynchronizer`. The validation therefore happens later than previously and the `EthProtocolManager` will require `Eth63` if fast sync is requested, even if it is disallowed due to the chain not being empty.  Given how rare `Eth62`-only clients are these days and that we expect to be able to find `Eth63` clients on the first run when fast sync is used this seems like a reasonable trade off. The code is now simpler and the decision to use fast sync is all in one place now.

The introduction of `FastSynchronizer` moves a lot of fast-sync specific code out of `DefaultSynchronizer`, keeping it simple and providing better isolation of concerns.


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/fastsync/FastSyncActions.java
```diff
@@ -126,6 +131,19 @@ private void waitForAnyPeer(final CompletableFuture<Void> result) {
   }
 
   public CompletableFuture<FastSyncState> selectPivotBlock() {
+    return loadPivotBlockFromStorage().orElseGet(this::selectPivotBlockFromPeers);
```
mbaxter: (optional / discussion) What about loading this from storage in the constructor?  Otherwise, successive calls to this method may always return the same value, which you may not want or may be unexpected.  Also thinking it might be nice to move the reading / writing of this to the `FastSyncState` class with something like `FastSyncState.create(dataDir)` that loads or creates a new state.

ajsutton: yeah I think there are definitely some improvements along these lines. The changes have wound up being reasonable large though so I'm going to merge this and put up a follow-up PR so we can review them and discuss more clearly.

Issue #839, opened 2019-02-12
Windows: "The input line is too long" error


glethuillier:
### Description

As a Windows (7, 8.1, or 10) user, I want to be able to install Pantheon in any directory and irrespectively of my username, so that I can run it.

### Acceptance Criteria

- On Windows **Command Prompt (cmd.exe)**, from the directory where the Pantheon binary has been built by user **tester.username** *(or any other username containing at least 15 chars)*, `pantheon -V` returns the current version of Pantheon.
- On Windows **PowerShell**, from the directory where the Pantheon binary has been built by user **tester.username** *(or any other username containing at least 15 chars)*, `./pantheon -V` returns the current version of Pantheon.

### Steps to Reproduce

1. Create a Windows account `tester.username`

2. Install Pantheon in `C:\Users\tester.username\` [using the official documentation](https://docs.pantheon.pegasys.tech/en/latest/Installation/Build-From-Source/#installation-on-windows).

   *(This is the directory that appears by default when starting Git bash)*

3. Start cmd.exe or PowerShell

4. `cd C:\Users\tester.username\pantheon\build\distributions\pantheon-1.0.0-SNAPSHOT\bin`

5. `dir`

   *(This command shall show two files: `pantheon` and `pantheon.bat`)*

6.  cmd.exe: `pantheon -V` or PowerShell: `./pantheon -V`

**Expected behavior:**

The current version of Pantheon is displayed.

Example: 

`C:\Users\tester.username\pantheon\build\distributions\pantheon-1.0.0-SNAPSHOT\bin>pantheon -V`
`pantheon/v1.0.0-dev-a82e91a5/windows-x86_64/oracle-java-1.8`

**Actual behavior:**

An error is displayed:

`C:\Users\tester.username\pantheon\build\distributions\pantheon-1.0.0-SNAPSHOT>bin\pantheon -V`
`The input line is too long.`
`The syntax of the command is incorrect.`

**Frequency:** 

100%

### Versions 

- Software version: *pantheon/v1.0.0-dev-a82e91a5/windows-x86_64/oracle-java-1.8*
- Java version: *Java 1.8.0_202*
- OS Name & Version: *Windows 7 Ultimate (64-bit version) / Windows 8.1 Pro (64-bit version) / Windows 10 Enterprise Edition* 
- Virtual Machine software & version: *VirtualBox Version 6.0.4 r128413 (Qt5.6.2)*

### Additional Information

**Observation**

[...]

PR #843, opened 2019-02-12
Fix 'the Input Is Too Long' Error on Windows

glethuillier:
## PR Description

Fix for the *“The input line is too long”* error on Windows.

This fix makes Gradle generate a shortened but operational command in `pantheon.bat`. 

The modified `pantheon.bat`, instead of explicitly referencing each .jar individually when initializing the classpath, uses the wildcard character `*` instead (on this point, see the [this link](https://docs.oracle.com/javase/6/docs/technotes/tools/windows/classpath.html)).

A longer-term solution could be to generate a fat jar, for instance. Whatever the chosen solution, the suggested fix is designed to be easily removed when the time comes. 

It is therefore intended as a non-intrusive fix.

Successfully tested on Windows 7 Ultimate (64-bit version), Windows 8.1 Pro (64-bit version), and Windows 10 Enterprise Edition.

## Fixed Issue

Fixes [#839](https://github.com/PegaSysEng/pantheon/issues/839).


build.gradle
```diff
@@ -336,6 +336,9 @@ startScripts {
   doLast {
     unixScript.text = unixScript.text.replace('PANTHEON_HOME', '\$APP_HOME')
     windowsScript.text = windowsScript.text.replace('PANTHEON_HOME', '%~dp0..')
```
ekellstrand: I believe this will still fail trying to generate the long classpath (which is then ignored) because the `set CLASSPATH=%APP_HOME%\lib\pantheon-1.0.0-SNAPSHOT.jar;...` is still too long. 
If I remove that `set CLASSPATH` line ( REM is not enough ), then things start to work.

glethuillier: I am with you on that. 

A solution could be to ultimately replace `set CLASSPATH=%APP_HOME%\lib\pantheon-1.0.0-SNAPSHOT.jar;...` with `set CLASSPATH=%APP_HOME%\lib\*`. What do you think?

shemnon:
@glethuillier - We are trying to do an RC tomorrow.  Would you like to do the honors of pressing "Squash and merge" (and probably "update branch")?

PR #844, opened 2019-02-12
Do parallel extract signatures in the parallel block importer.

shemnon:
## PR description

Much like the file importer do the signature extractions in parallel.


shemnon:
I don't intend to land this unless I see positive benchmarking results.

PR #846, opened 2019-02-13
[NC-2218] RPC auth config validation + tests.

mark-terry:
## PR description
Added validation for RPC authentication configuration at start up.

## Fixed Issue(s)

#NC-2218


pantheon/src/test/resources/auth_correct.toml
```diff
@@ -0,0 +1,20 @@
+[Users.userA]
+password = "$2a$10$l3GA7K8g6rJ/Yv.YFSygCuI9byngpEzxgWS9qEg5emYDZomQW7fGC"
+groups = ["admin"]
+permissions = ["eth:*", "perm:*"]
+roles = ["net"]
```
macfarla: in my PR I used / instead of : 
what you have done is what the design doc says. I can't think right now why I used / so I guess that needs changing

macfarla: The / will be JsonRpcMethod and NetListeningTest and one or two auth.toml files
If you feel like doing it, otherwise I will get to it later this afternoon

PR #848, opened 2019-02-13
Support resuming fast-sync downloads

ajsutton:
## PR description
Support resuming fast sync downloads. 

When download world state data, skip requesting data that we already have in our local store. To ensure we have all the child nodes continue traversing the trie.  Net result is that we still walk the entire trie but only download data we don't already have.

When shutting down Pantheon, don't delete the fast sync data

When starting up, don't fail if the world state pending nodes queue exists.  The way we use that RocksDB database, it will actually be treated as empty on reload regardless of existing content. This is due to it using two fields to track the head and tail of the queue - when a new queue is created those are both set to 0 so the queue is empty regardless of what's on disk. Since we're walking the entire trie anyway now we'll identify all the nodes we need anyway.


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/SynchronizerConfiguration.java
```diff
@@ -25,7 +25,7 @@
   // TODO: Determine reasonable defaults here
   public static final int DEFAULT_PIVOT_DISTANCE_FROM_HEAD = 500;
   public static final float DEFAULT_FULL_VALIDATION_RATE = .1f;
-  public static final int DEFAULT_FAST_SYNC_MINIMUM_PEERS = 5;
+  public static final int DEFAULT_FAST_SYNC_MINIMUM_PEERS = 1;
```
mbaxter: Looks like this might've been a change for debugging?

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/worldstate/WorldStateDownloader.java
```diff
@@ -146,6 +146,13 @@ private void requestNodeData(final BlockHeader header) {
             if (pendingRequest == null) {
               break;
             }
+            final Optional<BytesValue> existingData =
+                pendingRequest.getExistingData(worldStateStorage);
+            if (existingData.isPresent()) {
+              pendingRequest.setData(existingData.get());
+              queueChildRequests(pendingRequest);
+              continue;
```
mbaxter: Looks like you may need to change the containing `for` loop to a `while` loop so that we don't send out smaller than intended requests when we run into known nodes.

PR #849, opened 2019-02-13
Ibft notifies EthPeer when remote node has a better block

rain-on:
There is a failure mode of IBFT whereby a validator fails to import
a block, and also fails to receive the NewBlock message from its
peers. This means said validator is unable to participate in
subsequent rounds, and may cause the network to halt.

To overcome this issue, if an IBFT validator receives messages from
a future height, it will update the "BestEstimatedHeight" of the
corresponding EthPeer object, such that the Synchroniser will
(eventually) download the requisite blocks - thus allowing the
IBFT network to continue to operate.


consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/BaseSynchronizerUpdater.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package tech.pegasys.pantheon.consensus.ibft;
+
+import tech.pegasys.pantheon.ethereum.p2p.api.PeerConnection;
+
+public interface BaseSynchronizerUpdater {
```
ajsutton: How is this a "base" synchronizer updater?  Seems like this should be `SynchronizerUpdater` and the current `SynchronizerUpdater` should be `EthSynchronizerUpdater`.

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/SynchronizerUpdater.java
```diff
@@ -0,0 +1,45 @@
...
+  public SynchronizerUpdater(final EthPeers ethPeers) {
+    this.ethPeers = ethPeers;
+  }
+
+  @Override
+  public void updatePeerChainState(
+      final long knownBlockNumber, final PeerConnection peerConnection) {
+    final EthPeer ethPeer = ethPeers.peer(peerConnection);
+    if (ethPeer == null) {
+      LOG.warn("Received message from a peer with no corresponding EthPeer.");
```
ajsutton: I wouldn't log this at WARN level. It generally just means the peer disconnected between when the message was received and the processing got to this point.  Debug is probably the right level.

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/SynchronizerUpdater.java
```diff
@@ -0,0 +1,45 @@
...
+      return;
+    }
+
+    if (ethPeer.chainState().getEstimatedHeight() < knownBlockNumber) {
+      ethPeer.chainState().updateHeightEstimate(knownBlockNumber);
```
ajsutton: You don't need the if condition here,  `updateHeightEstimate` already checks the new height is higher than the previous. And it can do it safely inside a `synchronized` block.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/ChainState.java
```diff
@@ -88,7 +88,7 @@ public void updateForAnnouncedBlock(
     }
   }
 
-  private void updateHeightEstimate(final long blockNumber) {
+  public void updateHeightEstimate(final long blockNumber) {
     estimatedHeightKnown = true;
     if (blockNumber > estimatedHeight) {
       estimatedHeight = blockNumber;
```
ajsutton: This will need a `synchronized` block around the whole method content.  Previously it was only called from other methods which already had `synchronized`.

consensus/ibft/src/integration-test/java/tech/pegasys/pantheon/consensus/ibft/support/StubbedSynchronizerUpdater.java
```diff
@@ -0,0 +1,37 @@
...
+import tech.pegasys.pantheon.consensus.ibft.BaseSynchronizerUpdater;
+import tech.pegasys.pantheon.ethereum.p2p.api.PeerConnection;
+
+import java.util.Collection;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class StubbedSynchronizerUpdater implements BaseSynchronizerUpdater {
+
+  private Map<PeerConnection, ValidatorPeer> validatorNodes;
```
CjHare: What am I missing here, doesn't validatorNodes need initialising to avoid NPE?

consensus/ibft/src/test/java/tech/pegasys/pantheon/consensus/ibft/SynchronizerUpdaterTest.java
```diff
@@ -0,0 +1,122 @@
...
+    final long suppliedChainHeight = 4L;
+    when(ethPeers.peer(any())).thenReturn(ethPeer);
+    when(ethPeer.chainState()).thenReturn(chainState);
+    when(chainState.getEstimatedHeight()).thenReturn(knownChainHeight);
+
+    final SynchronizerUpdater updater = new SynchronizerUpdater(ethPeers);
+
+    updater.updatePeerChainState(suppliedChainHeight, createAnonymousPeerConnection());
+
+    verify(chainState, times(0)).updateHeightEstimate(anyLong());
```
CjHare: Should the `anyLong()` be `knownChainHeight`? ...unless any value is really acceptable?

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/BaseSynchronizerUpdater.java
rain-on: I was really hoping someone would help out with the naming.... :)

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/SynchronizerUpdater.java
rain-on: done.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/ChainState.java
rain-on: done.

consensus/ibft/src/integration-test/java/tech/pegasys/pantheon/consensus/ibft/support/StubbedSynchronizerUpdater.java
rain-on: There's no NPE as the addNetworkPeers creates the map and assigns the created map to this variable.

consensus/ibft/src/test/java/tech/pegasys/pantheon/consensus/ibft/SynchronizerUpdaterTest.java
rain-on: given this is a verify (rather than a "thenReturns") - I'd like to make sure that this function wasn't invoked at all - regardless of the parameter.

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/SynchronizerUpdater.java
rain-on: done.

consensus/ibft/src/integration-test/java/tech/pegasys/pantheon/consensus/ibft/support/RoundSpecificPeers.java
```diff
@@ -125,6 +126,10 @@ public void commitForNonProposing(final ConsensusRoundIdentifier roundId, final
     nonProposingPeers.forEach(peer -> peer.injectCommit(roundId, hash));
   }
 
+  public void forNonProposing(final Consumer<ValidatorPeer> assertion) {
+    nonProposingPeers.forEach(peer -> assertion.accept(peer));
```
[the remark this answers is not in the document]
rain-on: way better.nice.

consensus/ibft/src/integration-test/java/tech/pegasys/pantheon/consensus/ibft/support/StubbedSynchronizerUpdater.java
```diff
@@ -0,0 +1,37 @@
...
+  @Override
+  public void updatePeerChainState(
+      final long knownBlockNumber, final PeerConnection peerConnection) {
+    validatorNodes.get(peerConnection).updateEstimatedChainHeight(knownBlockNumber);
+  }
+
+  public void addNetworkPeers(final Collection<ValidatorPeer> nodes) {
+    validatorNodes =
+        nodes.stream()
+            .collect(Collectors.toMap(ValidatorPeer::getPeerConnection, validator -> validator));
```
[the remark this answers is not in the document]
rain-on: done.

consensus/ibft/src/test/java/tech/pegasys/pantheon/consensus/ibft/SynchronizerUpdaterTest.java
```diff
@@ -0,0 +1,122 @@
...
+    final long suppliedChainHeight = 4L;
+    when(ethPeers.peer(any())).thenReturn(ethPeer);
+    when(ethPeer.chainState()).thenReturn(chainState);
+    when(chainState.getEstimatedHeight()).thenReturn(knownChainHeight);
+
+    final EthSynchronizerUpdater updater = new EthSynchronizerUpdater(ethPeers);
+
+    updater.updatePeerChainState(suppliedChainHeight, createAnonymousPeerConnection());
+
+    verify(chainState, times(0)).updateHeightEstimate(anyLong());
```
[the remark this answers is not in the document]
rain-on: done.

consensus/ibft/src/integration-test/java/tech/pegasys/pantheon/consensus/ibft/support/StubbedSynchronizerUpdater.java
CjHare: 👍

consensus/ibft/src/integration-test/java/tech/pegasys/pantheon/consensus/ibft/support/StubbedSynchronizerUpdater.java
```diff
@@ -0,0 +1,38 @@
...
+
+  private Map<PeerConnection, ValidatorPeer> validatorNodes;
+
+  @Override
+  public void updatePeerChainState(
+      final long knownBlockNumber, final PeerConnection peerConnection) {
+    validatorNodes.get(peerConnection).updateEstimatedChainHeight(knownBlockNumber);
+  }
+
+  public void addNetworkPeers(final Collection<ValidatorPeer> nodes) {
```
CjHare: Is this an `add` operations or a `set` operation? ...what happens when its invoked twice?

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/IbftController.java
```diff
@@ -248,5 +254,9 @@ private void addMessageToFutureMessageBuffer(final long chainHeight, final Messa
       futureMessages.put(chainHeight, Lists.newArrayList());
     }
     futureMessages.get(chainHeight).add(rawMsg);
+
+    // Notify the synchroniser the transmitting peer must have the parent block to the received
```
jframe: Think this would make more sense being in processMessage after addMessageToFutureMessageBuffer call instead of being in the addMessageToFutureMessageBuffer method as notifying the synchroniser is not really part of updating the future message buffer.

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/IbftController.java
```diff
@@ -248,5 +254,9 @@ private void addMessageToFutureMessageBuffer(final long chainHeight, final Messa
...
+    // message's target height.
+    sychronizerUpdater.updatePeerChainState(chainHeight - 1L, rawMsg.getConnection());
```
jframe: Do we also need to call updatePeerChainState when we receive a new block event? I would hope that we don't need but not sure.

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/IbftController.java
rain-on: I had assumed that when the "smart" FutureMessageBuffer came to fruition then this would move across into that class, and it would be the FutureMessageHandler - meaning it would buffer and notify other people, thus the IbftController doesn't need to know about this stuff...

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/IbftController.java
rain-on: No we don't - that's already been handled by the BlockPropogation mechanism.

consensus/ibft/src/integration-test/java/tech/pegasys/pantheon/consensus/ibft/support/StubbedSynchronizerUpdater.java
CjHare: any thoughts? @rain-on

rain-on: just saw this - create the map instantiation, then putAll in here. Thus will continue to grow as you call "add" - given its called once in the integration test framework setup, should be acceptable (i.e. doesn't require a "clear()" call atm.

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/IbftController.java
jframe: fair enough

PR #857, opened 2019-02-14
Create SignedDataFactory

rain-on:
Split out the signing of data from the MessageFactory - not sure yet if this is useful...


rain-on:
This really did not help.

PR #858, opened 2019-02-14
Force fast sync chain to be canonical

ajsutton:
## PR description
When doing a fast sync, force re-org the chain to the chain we're fast syncing. We're only downloading the world state for the fast sync chain so need canonical head to be pointed there.


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/chain/DefaultMutableBlockchain.java
```diff
@@ -304,12 +332,14 @@ private BlockAddedEvent handleChainReorg(
...
-        forks.stream().filter(f -> f.equals(newChainHead.getHeader().getParentHash())).findAny();
-    parentFork.ifPresent(forks::remove);
+    // Old head is now a fork unless the common ancestor was the old head
+    // in which case we must be doing a forced re-org forward on the canonical chain.
+    if (!oldChain.getHash().equals(oldChainHead)) {
+      forks.add(oldChainHead);
+    }
+    // Remove new chain head and its parent if either was tracked as a fork
+    forks.remove(newChainHead.getHeader().getParentHash());
+    forks.remove(newChainHead.getHash());
```
mbaxter: So, previously all of this code was written assuming we would only ever be appending unknown / new blocks and the chainhead would always be the "best" known block.  But since we can now reorg to known blocks, I think there might be more modifications we need to make so that all of the data is consistent.  For example, `appendBlock` will return early if we already know about the block.  So, if we reorg to a known block and then append a known block, the various indexes won't get updated.  I think some of the fork tracking might be off too.  So, if you reorg to a fork (but not the head of the fork), and then keep appending blocks matching this fork, walking through the forkhead, that forkhead won't currently get removed from the tracked forks.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/chain/DefaultMutableBlockchain.java
```diff
@@ -155,6 +156,33 @@ public synchronized void appendBlock(final Block block, final List<TransactionRe
...
+    final Updater updater = blockchainStorage.updater();
+    try {
+      handleChainReorg(updater, block);
+    } catch (final NoSuchElementException e) {
+      // Any Optional.get() calls in this block should be present, missing data means data
+      // corruption or a bug.
+      updater.rollback();
+      throw new IllegalStateException("Blockchain is missing data that should be present.", e);
+    }
+    updater.commit();
```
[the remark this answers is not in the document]
ajsutton: We bail out with an exception if we rollback but it probably is clearer inside the try block.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/chain/DefaultMutableBlockchain.java
ajsutton: ugh true - this opens up a heap of extra complexity.  Three options that I can see:
 1. Delete the blocks we re-orged away from. Works for fast sync, not what you want as a general force-the-chain-head-to-this function
 2. Ditch this entirely. It really shouldn't be required for fast sync anyway and seems pretty dangerous.
 3. Address the corner cases, add a heap more tests and probably setup some fuzz testing.

I think my preference at this point is for option 2 - just don't do this (at least for now).

mbaxter: yeah, I think #2 makes sense for now

ajsutton:
Abandoning this approach for now.  Too much risk of it making things worse for a very unlikely corner case where fast sync manages to run up a big chain without the random verification noticing.

PR #862, opened 2019-02-14
Disconnect peers where the common ancestor is before our fast sync pivot

ajsutton:
## PR description
This is the simplest possible handling of missing world states when finding a sync target.  If we don't have a world state for common ancestor block, then it must be before the fast sync pivot block and we'll either have to start again from genesis (opening up significant potential for griefing) or refuse to sync to the peer.  So we refuse to sync to the peer and disconnect them.

This behaviour will need to be more nuanced when state pruning is adding as in that case we may want to search back from the common ancestor some number of blocks to find a world state we do have.  The details of how that behaviour would depend on the state pruning implementation though so going to keep it simple for now.


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/fullsync/FullSyncTargetManager.java
```diff
@@ -50,10 +53,28 @@
...
+  @Override
+  protected Optional<SyncTarget> finalizeSelectedSyncTarget(final SyncTarget syncTarget) {
+    final BlockHeader commonAncestor = syncTarget.commonAncestor();
+    if (protocolContext
+        .getWorldStateArchive()
+        .isWorldStateAvailable(commonAncestor.getStateRoot())) {
+      return Optional.of(syncTarget);
+    } else {
+      LOG.warn(
+          "Disconnecting {} because common ancestor at block {} is prior to our fast sync pivot block",
```
mbaxter: For future-proofing, might be better to have a more generic message.  Something like "Disconnecting {} because world state is unavailable at common ancestor block {}"

ajsutton: Yeah I went back and forth on how specific to be.  I suspect this area needs reworking as soon as we have a different way to wind up with blocks without state (e.g. pruning) anyway but probably better to be a bit generic.

PR #864, opened 2019-02-14
Break out RoundChangeCertificate validation

rain-on:
With upcoming changes to remove the NewRound message, and place the RoundChangeCertificate in the Proposal, it was decided to break out the RoundChangeCertiifcate validation into a separate file to minimise changes during message restructuring.


consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/validation/RoundChangeCertificateValidator.java
```diff
@@ -0,0 +1,129 @@
...
+
+    final Collection<SignedData<RoundChangePayload>> roundChangeMsgs =
+        roundChangeCert.getRoundChangePayloads();
+
+    if (roundChangeMsgs.size() < quorum) {
+      LOG.info("Invalid RoundChangeCertificate, insufficient RoundChange messages.");
+      return false;
+    }
+
+    if (!roundChangeCert.getRoundChangePayloads().stream()
```
CjHare: Would the RoundChangeCertificate benefit from providing this behaviour in a method, say `public boolean doesRoundEqual(ConsensRoundIdentifier expectedRound)`, as that would make the semantics nicer in this spot.

rain-on: Just the round check (i.e. not whole validation) - yeah could, the function is more like "doAllRoundChangesMatch(ConsensusRoundIdentifier expectedRound)"

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/validation/RoundChangeCertificateValidator.java
```diff
@@ -0,0 +1,129 @@
...
+        .allMatch(p -> p.getPayload().getRoundIdentifier().equals(expectedRound))) {
+      LOG.info(
+          "Invalid RoundChangeCertificate, not all embedded RoundChange messages have a "
+              + "matching target round.");
+      return false;
+    }
+
+    for (final SignedData<RoundChangePayload> roundChangeMsg :
```
jframe: The RoundChangePayloadValidator only needs to be created once. And this validation check could be replaced with allMatch on the stream so could do something like if (!roundChangeCert.getRoundChangePayloads().stream()
        .allMatch(roundChangeValidator::validateRoundChange)) { .. }

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/validation/RoundChangeCertificateValidator.java
```diff
@@ -0,0 +1,129 @@
...
+  public RoundChangeCertificateValidator(
+      final Collection<Address> validators,
+      final MessageValidatorForHeightFactory messageValidatorFactory,
+      final long chainHeight) {
+    this.validators = validators;
+    this.messageValidatorFactory = messageValidatorFactory;
+    this.quorum = IbftHelpers.calculateRequiredValidatorQuorum(validators.size());
+    this.chainHeight = chainHeight;
+  }
+
```
jframe: would if make sense to just have a top level validateRoundChangeCertificate method?

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/validation/RoundChangeCertificateValidator.java
rain-on: done.

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/validation/RoundChangeCertificateValidator.java
rain-on: At this stage we can't - there's 2 concepts - when we make Proposal, this may be an option.
1. Is the RoundChange self-consistent?
2. Does this Block align with the content in the RoundChange?

PR #867, opened 2019-02-14
[PAN-2196] Implement world state cancel

mbaxter:
## PR description
Implement world state download cancellation functionality.  On cancellation, the `WorldStateDownloader` will cancel any pending network tasks and clear its task queue.  Also made a few other small changes: 
- ~Removed the `status` tracking from the downloader in favor of using the future as a signal on the status~.  
- Modified the downloader to return an exceptional result if `run` is called on an already running instance.

### DIFF
--- a/ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/worldstate/WorldStateDownloader.java
+++ b/ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/worldstate/WorldStateDownloader.java
@@ -14,9 +14,11 @@ package tech.pegasys.pantheon.ethereum.eth.sync.worldstate;
 
 import tech.pegasys.pantheon.ethereum.core.BlockHeader;
 import tech.pegasys.pantheon.ethereum.core.Hash;
+import tech.pegasys.pantheon.ethereum.eth.manager.AbstractPeerTask;
 import tech.pegasys.pantheon.ethereum.eth.manager.AbstractPeerTask.PeerTaskResult;
 import tech.pegasys.pantheon.ethereum.eth.manager.EthContext;
 import tech.pegasys.pantheon.ethereum.eth.manager.EthPeer;
+import tech.pegasys.pantheon.ethereum.eth.manager.EthTask;
 import tech.pegasys.pantheon.ethereum.eth.sync.tasks.GetNodeDataFromPeerTask;
 import tech.pegasys.pantheon.ethereum.eth.sync.tasks.WaitForPeerTask;
 import tech.pegasys.pantheon.ethereum.worldstate.WorldStateStorage;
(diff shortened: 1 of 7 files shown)


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/worldstate/WorldStateDownloader.java
```diff
@@ -100,32 +98,35 @@ public WorldStateDownloader(
...
+      }
     }
 
+    requestNodeData(header);
     return future;
   }
 
   public void cancel() {
-    // TODO
+    getFuture().cancel(true);
```
ajsutton: Given that the use case for this is likely to be cancelling one download and immediately starting a new one for a later block, how do we avoid a race condition where outstanding requests and/or `handleCancellation` are still happening after this method returns and a new download starts?

services/queue/src/main/java/tech/pegasys/pantheon/services/queue/RocksDbTaskQueue.java
```diff
@@ -112,9 +112,25 @@ public synchronized long size() {
...
+  @Override
+  public synchronized void clear() {
+    assertNotClosed();
+    outstandingTasks.clear();
+    byte[] from = Longs.toByteArray(oldestKey.get());
+    byte[] to = Longs.toByteArray(lastEnqueuedKey.get() + 1);
+    try {
+      db.deleteRange(from, to);
+      lastDequeuedKey.set(0);
+      lastEnqueuedKey.set(0);
```
ajsutton: Do we need to set `oldestKey` to 0 as well?

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/worldstate/WorldStateDownloader.java
ajsutton: Just to think this through a bit more...

When you can `cancel`, that *same* thread will be used to run any follow actions added via `whenComplete`, `thenApply` and friends.  So if you call `WorldStateDonwloader.cancel` by the time that method returns we do actually know that `handleCancellation` has completed.  However, we don't know if `handleCancellation` runs before or after anything else that was chained to that future.

I think that will probably work out ok but we may want to have `createFuture` return the future returned from `whenComplete` instead of the original to fix it (just need to make sure it doesn't interfere with whether the returned future completes exceptionally or not).

services/queue/src/main/java/tech/pegasys/pantheon/services/queue/RocksDbTaskQueue.java
mbaxter: yes - good catch.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/worldstate/WorldStateDownloader.java
mbaxter: Another good catch.  I think I broke this when I removed the `status` var.  I reverted that change and now you shouldn't be able to execute `run()` before the cancellation has finished because the checks are now based on the `status` variable which gets updated inside of the synchronized `handleCancellation` method.

mbaxter: I went ahead and merged because I have another branch built on top of this one.  But if you see any issues with my follow-up updates, I can fix them in another PR.

PR #868, opened 2019-02-15
Introduce FutureUtils to reduce duplicated code around CompletableFuture

ajsutton:
## PR description
Introduce `FutureUtils` to provide some commonly required functionality around `CompletableFuture` such as creating a future that is completed exceptionally, handling exceptionally completed futures by composing with another future and propagating the result or exception from one future to another.

Applies these utilities in a few places to simplify existing code. There will certainly be more places they could be used, but this is a start.


util/src/main/java/tech/pegasys/pantheon/util/FutureUtils.java
```diff
@@ -0,0 +1,83 @@
...
+   * @return the CompletionStage
+   */
+  public static <T> CompletableFuture<T> exceptionallyCompose(
+      final CompletableFuture<T> future,
+      final Function<Throwable, CompletionStage<T>> errorHandler) {
+    final CompletableFuture<T> result = new CompletableFuture<>();
+    future.whenComplete(
+        (value, error) -> {
+          final CompletionStage<T> nextStep =
+              error != null ? errorHandler.apply(error) : completedFuture(value);
```
mbaxter: Should we add a try / catch around the `errorHandler` execution in case that function throws?

ajsutton: Yep seems like a good idea.

PR #873, opened 2019-02-15
Limit size of Ibft future message buffer

jframe:
## PR description
This add a limit to the number of ibft messages that are stored in the future buffer.

There is a total number of messages that are allowed. And also a max distance from the current height that messages are allowed for.

If the total message limit is reached then all the messages in the highest chain height are removed.

Additionally if there is only one height then we will remove the oldest message to make room instead.


consensus/ibft/src/integration-test/java/tech/pegasys/pantheon/consensus/ibft/support/TestContextBuilder.java
```diff
@@ -307,7 +309,9 @@ private static ControllerAndState createControllerAndFinalState(
                     messageValidatorFactory),
                 messageValidatorFactory),
             gossiper,
-            DUPLICATE_MESSAGE_LIMIT);
+            DUPLICATE_MESSAGE_LIMIT,
```
rain-on: would probably prefer to see the futureMessageBuffer get created here and passed in (rather than the raw values) - why? Can't answer that.

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/FutureMessageBuffer.java
```diff
@@ -0,0 +1,101 @@
...
+    this.buffer = buffer;
+  }
+
+  public void addMessage(final long msgChainHeight, final Message rawMsg) {
+    final long currentHeight = currentChainHeight.get();
+    if (!validMessageHeight(msgChainHeight, currentHeight)) {
+      return;
+    }
+
+    if (totalMessages >= futureMessagesLimit) {
```
rain-on: I had assumed evictMessages() MAY need to loop (i.e. what if you need to remove more than 1 height?) - should this conditional become a loop in evict?

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/FutureMessageBuffer.java
```diff
@@ -0,0 +1,101 @@
...
+      evictMessages();
+    }
+
+    if (!buffer.containsKey(msgChainHeight)) {
+      buffer.put(msgChainHeight, new ArrayList<>());
```
rain-on: what if this msgChainHeight is higher than the height removed in evictMessage? Eg. MsgChainHeight is 5, but we just removed chain-height 4 ...

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/FutureMessageBuffer.java
```diff
@@ -0,0 +1,101 @@
...
+    final boolean withinMaxChainHeight =
+        msgChainHeight <= currentHeight + futureMessagesMaxDistance;
+    return isFutureMsg && withinMaxChainHeight;
+  }
+
+  private void evictMessages() {
+    if (buffer.size() > 1) {
+      buffer.remove(buffer.lastKey());
+    } else {
+      List<Message> messages = buffer.firstEntry().getValue();
```
rain-on: Is it possible that buffer size is 0? I guess not.

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/FutureMessageBuffer.java
jframe: I don't think there is any need to remove any more than one height at a time. Each height must contain at least one message so if we remove a height we will have room for the message to be added.

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/FutureMessageBuffer.java
jframe: Assuming that msgChainHeight is still within the max distance we would add a message for chain height 5. Which is perhaps is a bit odd. Maybe it would be better not to add in that case, essentially evicting itself as the highest chain height? Need to think about this a bit more.

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/FutureMessageBuffer.java
jframe: I am making an assumption that it isn't 0 at this point but that only holds if futureMessagesLimit isn't 0. Be better if the evict messages handled that case, so i'll update evictMessages to handle that properly.

consensus/ibft/src/integration-test/java/tech/pegasys/pantheon/consensus/ibft/support/TestContextBuilder.java
CjHare: 👍 ^^^
It's cleaner, in terms of IoC and reduces number of constructors on the controller.
I'd suggest instead of passing in values, construct & pass in the objects: 
`final MessageTracker duplicateMessageTracker, final FutureMessageBuffer futureMessageBuffer`

jframe: That is cleaner and also reduces the need the seperate constructor in the IbftController.
Done.

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/FutureMessageBuffer.java
```diff
@@ -0,0 +1,99 @@
...
+  private final NavigableMap<Long, List<Message>> buffer;
+  private final long futureMessagesMaxDistance;
+  private final long futureMessagesLimit;
+
+  public FutureMessageBuffer(final long futureMessagesMaxDistance, final long futureMessagesLimit) {
+    this(futureMessagesMaxDistance, futureMessagesLimit, new TreeMap<>());
+  }
+
+  @VisibleForTesting
+  FutureMessageBuffer(
```
CjHare: Is this constructor really of use? ...looks like when it's called in the tests a `new TreeMap<>` is passed in

jframe: It is, I'm using that so I can check the state of the map is correct in the tests without exposing the map to everyone. A new TreeMap<> is passed in, but I'm checking in my tests messages are added to the map correctly.

CjHare: That's a tight coupling between the test and the specific implementation of the logic, I'm not convinced that's the best approach.

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/FutureMessageBuffer.java
rain-on: Yeah, I think I'd rather see it evict itself. MAYBE you could do something like add the new message, then run the eviction process.

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/FutureMessageBuffer.java
jframe: done. updated so don't need special constructor

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/FutureMessageBuffer.java
jframe: done. added handling of zero message limit.

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/FutureMessageBuffer.java
jframe: done. yeah I think it better for it to be evicted. changed the code to run the evict after after adding, avoids needing to special case this add case.

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/FutureMessageBuffer.java
```diff
@@ -0,0 +1,91 @@
...
+
+    addMessageToBuffer(msgChainHeight, rawMsg);
+
+    if (totalMessagesSize() > futureMessagesLimit) {
+      evictMessages();
+    }
+  }
+
+  private void addMessageToBuffer(final long msgChainHeight, final Message rawMsg) {
+    if (!buffer.containsKey(msgChainHeight)) {
```
rain-on: nit: could use computeIfAbsent/putIfAbsent

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/FutureMessageBuffer.java
```diff
@@ -0,0 +1,91 @@
...
+  private final NavigableMap<Long, List<Message>> buffer = new TreeMap<>();
+  private final long futureMessagesMaxDistance;
+  private final long futureMessagesLimit;
+
+  public FutureMessageBuffer(final long futureMessagesMaxDistance, final long futureMessagesLimit) {
+    this.futureMessagesMaxDistance = futureMessagesMaxDistance;
+    this.futureMessagesLimit = futureMessagesLimit;
+  }
+
+  public void addMessage(final long chainHeight, final long msgChainHeight, final Message rawMsg) {
```
rain-on: don't LOVE msgChainHeight coming in on this API .... rather this get reset in the Poll somehow.

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/FutureMessageBuffer.java
jframe: done

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/FutureMessageBuffer.java
```diff
@@ -0,0 +1,101 @@
...
+ * are removed. Otherwise if there is only one height the oldest inserted message is removed.
+ */
+public class FutureMessageBuffer {
+  private final NavigableMap<Long, List<Message>> buffer = new TreeMap<>();
+  private final long futureMessagesMaxDistance;
+  private final long futureMessagesLimit;
+  private long chainHeight;
+
+  public FutureMessageBuffer(
+      final long futureMessagesMaxDistance, final long futureMessagesLimit, final int chainHeight) {
```
CjHare: Drop the `chainHeight` and initialise it to zero, as the only place it's getting called in code is the IbftController that is passing in a hardcoded zero.

jframe: nice catch. that should not of been initialised to zero. fixed now.

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/FutureMessageBuffer.java
```diff
@@ -0,0 +1,103 @@
...
+  public FutureMessageBuffer(
+      final long futureMessagesMaxDistance,
+      final long futureMessagesLimit,
+      final long chainHeight) {
+    this.futureMessagesMaxDistance = futureMessagesMaxDistance;
+    this.futureMessagesLimit = futureMessagesLimit;
+    this.chainHeight = chainHeight;
+  }
+
+  public void updateChainHeight(final long chainHeight) {
```
rain-on: sorry to be difficult, are we worried that both this and poll take "chainHeight" - but they are different numbers?
I.e. this is the number of the most recent block on the chain - whereas 'poll' is really the "next" block ...
Wondering if renaming these functions to (something like):
* discardMessagesForHeightsBelow(x)
* extractMessagesWithHeight(y)

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/FutureMessageBuffer.java
jframe: done.changed so it is using been passed in as a arg.

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/FutureMessageBuffer.java
jframe: The height in the poll is really the message chain height, whereas the updateChainHeight is the current chain height we are working so they are different. I think renaming the arg in poll to msgChainHeight might that clearer.

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/statemachine/FutureMessageBuffer.java
```diff
@@ -0,0 +1,103 @@
...
+    if (buffer.size() > 1) {
+      buffer.remove(buffer.lastKey());
+    } else if (buffer.size() == 1) {
+      List<Message> messages = buffer.firstEntry().getValue();
+      messages.remove(0);
+    }
+  }
+
+  public List<Message> retrieveMessagesForHeight(final long height) {
+    this.chainHeight = height;
```
rain-on: nit: no need for the 'this'

jframe: done

PR #878, opened 2019-02-18
Add metrics for EthScheduler executors

ajsutton:
## PR description
Adds metrics to the executors used within EthScheduler so we have visibility in what they're doing, how long the work queues are etc.


ajsutton:
No Grafana pics, it was enough for me to take a quick look at the numbers and see we weren't rejecting tasks or building up huge backlogs (at least in the case I was interested in).

PR #879, opened 2019-02-18
[PAN-2270] Rename password hash command

lucassaldanha:
Renaming password-hash command to password hash --password=foo


docs/Reference/Pantheon-CLI-Syntax.md
```diff
@@ -926,14 +926,18 @@ $ pantheon public-key export --to=/home/me/me_project/not_precious_pub_key
...
 This command generates the hash of a given password.
 
 ```bash tab="Syntax"
-$ pantheon password-hash <my-password>
+$ pantheon password hash --password=<my-password>
 ```
 
 ```bash tab="Example"
-$ pantheon password-hash "password123"
+$ pantheon password hash --password="password123"
```
MadelineMurray: Should the quotes be removed from this example?  With the old `password-hash` using quotes gave you the wrong hash (ie, then the login didn't work)

PR #887, opened 2019-02-18
Unit tests of BlockBroadcaster

s-matthew-english:
## PR description

includes two unit tests: 

--> one the confirms the `propagateBlock` method on `EthPeer` is invoked when all criteria are satisfied. 

--> one that confirms the `propagateBlock` method on `EthPeer` is not invoked when all criteria are not satisfied. 

**NOTE**
when 808 lands the change set of this code will be reduced to only the file `BlockBroadcasterTest.java`


ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockPropagationManagerTest.java
```diff
@@ -561,4 +566,25 @@ public void shouldNotImportBlocksThatAreAlreadyBeingImported() {
 
     verify(ethScheduler, times(1)).scheduleSyncWorkerTask(any(Supplier.class));
   }
+
```
ekellstrand: Idea: Would it test any addition workflow (I'm not sure offhand) if you added a third node?  ie, create the block on first node like currently.  have node 2 process everything with native (non-test code), and wait for the message to arrive or process on 3rd node?

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockBroadcasterTest.java
```diff
@@ -0,0 +1,104 @@
...
+            .difficulty(UInt256.ZERO)
+            .number(0L)
+            .gasLimit(1L)
+            .gasUsed(1L)
+            .timestamp(Instant.now().truncatedTo(ChronoUnit.SECONDS).getEpochSecond())
+            .extraData(Bytes32.wrap(bytes(Bytes32.SIZE)))
+            .mixHash(Hash.ZERO)
+            .nonce(0L)
+            .blockHashFunction(MainnetBlockHashFunction::createHash)
+            .buildBlockHeader();
```
ajsutton: You can just use `BlockHeaderTestFixure` to create a simple test block.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockBroadcasterTest.java
```diff
@@ -0,0 +1,97 @@
...
+    final BlockBroadcaster blockBroadcaster = new BlockBroadcaster(ethContext);
+    final Block block = generateBlock();
+    blockBroadcaster.propagate(block, UInt256.ZERO);
+
+    verify(ethPeer0, never()).propagateBlock(any(), any());
+    verify(ethPeer1, times(1)).propagateBlock(any(), any());
+  }
+
+  @Test
+  public void blockPropagationRejectOnHasSeenBlock() {
```
ajsutton: We probably don't need this one given we have the test above to check we send only to peers that haven't seen the block.

PR #889, opened 2019-02-18
RPC Pub/Sub clarification from feedback 

MadelineMurray:
Update to address feedback from @hmijail


docs/Reference/Using-JSON-RPC-API.md
```diff
@@ -78,7 +78,8 @@ Send individual requests as a JSON data package at each prompt:
 > {"jsonrpc":"2.0","method":"web3_clientVersion","params":[],"id":53}
 ```
 
-The [RPC Pub/Sub methods](../Using-Pantheon/RPC-PubSub.md) can also be used over WebSockets.
+As well at the [JSON-RPC API methods](../Reference/JSON-RPC-API-Methods.md), the [RPC Pub/Sub methods](../Using-Pantheon/RPC-PubSub.md)
```
lucassaldanha: I think this can be misleading. Pub-Sub methods can ONLY be used with WebSockets. All other JSON-RPC API methods can be used either via HTTP or WebSockets.

MadelineMurray: This line is in a section on WebSockets within the JSON-RPC content - does that make difference?  https://docs.pantheon.pegasys.tech/en/stable/Reference/Using-JSON-RPC-API/#http-and-websocket-requests

PR #896, opened 2019-02-19
[MINOR] Move metric instantiation

mbaxter:
## PR description
Instantiate individual metrics where they're used, rather than passing them into constructors.  This helps us stop metric construction from migrating all the way up to the `PantheonController` and should make constructor parameters more stable.


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/task/AbstractEthTask.java
```diff
@@ -29,19 +31,27 @@
 
 public abstract class AbstractEthTask<T> implements EthTask<T> {
 
+  static volatile LabelledMetric<OperationTimer> ethTasksTimer = null;
```
ajsutton: This shouldn't be static - we don't want the timer to be shared across separate instances of Pantheon.  To avoid creating multiple metrics with the same name (which isn't allowed) you'd have to cache them in the MetricsSystem.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/task/AbstractEthTask.java
```diff
@@ -29,19 +31,27 @@
...
+  protected AbstractEthTask(final MetricsSystem metricsSystem) {
+    setupTaskTimer(metricsSystem);
     taskTimer = ethTasksTimer.labels(getClass().getSimpleName());
   }
 
+  private synchronized void setupTaskTimer(final MetricsSystem metricsSystem) {
+    if (ethTasksTimer == null) {
+      ethTasksTimer =
+          metricsSystem.createLabelledTimer(
+              MetricCategory.SYNCHRONIZER, "task", "Internal processing tasks", "taskName");
```
ajsutton: This has a race condition where separate instances could both wind up creating the timer at the same time with one of them likely failing because of the duplicate name check.  The `synchronized` only restricts it to a single thread per instance because the method isn't static.

mbaxter:
Closing this pending rework related to metrics caching

mbaxter:
Updated `PrometheusMetricsSystem` to cache metrics

PR #899, opened 2019-02-19
When picking fast sync pivot block, use the peer with the best total difficulty

ajsutton:
## PR description
When picking the fast sync pivot block, ensure the peer we consider best is evaluated on total difficulty, instead of chain height. Otherwise on MainNet we tend to pick an ETC node and pivot block because it has a greater height even though it has lower total difficulty.


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/EthPeers.java
```diff
@@ -33,6 +33,9 @@
   public static final Comparator<EthPeer> CHAIN_HEIGHT =
       Comparator.comparing(((final EthPeer p) -> p.chainState().getEstimatedHeight()));
 
+  private static final Comparator<EthPeer> HIGHEST_TOTAL_DIFFICULTY_PEER =
+      TOTAL_DIFFICULTY.thenComparing(CHAIN_HEIGHT);
```
shemnon: Apart from total difficulty (which is the only thing the yellow paper defines) do we know what other clients do to distinguish heavier chains?  Chain height is a quick win but Ghost also originally defined ommers as adding to the difficulty (but that's not in the yellow paper).

ajsutton: The only thing that matters is total difficulty - if two chains have the same difficulty then clients will generally stick with whichever one they had first (ie equal difficulty does not trigger a re-org) although I have a vague memory of Geth including randomness in there for some reason.

Since Byzantium the difficulty of a block is increased if it contains any ommers.

PR #903, opened 2019-02-19
Periodically try to connect to some peers already in our peer table

ajsutton:
## PR description
Periodically select some peers already in our peer table and attempt to connect to them.  Previously we only attempted to connect to newly discovered peers, but if the pending connections exceeded our max peer limit we wouldn't make any attempt to connect.

Apart from needing some tests, before merging we should take some measurements to see if this actually finds more peers or not.  Initial tests are inconclusive.

### DIFF
--- a/ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/netty/NettyP2PNetwork.java
+++ b/ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/netty/NettyP2PNetwork.java
@@ -24,6 +24,7 @@ import tech.pegasys.pantheon.ethereum.p2p.discovery.DiscoveryPeer;
 import tech.pegasys.pantheon.ethereum.p2p.discovery.PeerDiscoveryAgent;
 import tech.pegasys.pantheon.ethereum.p2p.discovery.PeerDiscoveryEvent.PeerBondedEvent;
 import tech.pegasys.pantheon.ethereum.p2p.discovery.PeerDiscoveryEvent.PeerDroppedEvent;
+import tech.pegasys.pantheon.ethereum.p2p.discovery.PeerDiscoveryStatus;
 import tech.pegasys.pantheon.ethereum.p2p.discovery.VertxPeerDiscoveryAgent;
 import tech.pegasys.pantheon.ethereum.p2p.discovery.internal.PeerRequirement;
 import tech.pegasys.pantheon.ethereum.p2p.peers.DefaultPeer;
@@ -43,6 +44,7 @@ import tech.pegasys.pantheon.util.Subscribers;
 
 import java.net.InetSocketAddress;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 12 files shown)


ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/netty/NettyP2PNetwork.java
```diff
@@ -335,6 +339,32 @@ public void checkMaintainedConnectionPeers() {
...
+      Collections.shuffle(peers);
+      LOG.info("Initiating connection to {} peers from the peer table", availablePeerSlots);
+      peers.stream().limit(availablePeerSlots).forEach(this::connect);
+      LOG.info(
+          "Connection count {} (pending: {}, connected: {})",
+          this.connectionCount(),
+          pendingConnections.size(),
+          connections.size());
+    }
+  }
```
ajsutton: This is the key part of the change.  Pretty much all of the rest is just removing an unnecessary conversion from Stream to list when in every production case we then convert back to a stream anyway.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/netty/NettyP2PNetwork.java
```diff
@@ -335,6 +339,32 @@ public void checkMaintainedConnectionPeers() {
...
+          peerDiscoveryAgent
+              .getPeers()
+              .filter(peer -> peer.getEndpoint().getTcpPort().isPresent())
+              .filter(peer -> peer.getStatus() == PeerDiscoveryStatus.BONDED)
+              .filter(peer -> !isConnected(peer) && !isConnecting(peer))
+              .collect(Collectors.toList());
+      Collections.shuffle(peers);
+      LOG.info("Initiating connection to {} peers from the peer table", availablePeerSlots);
+      peers.stream().limit(availablePeerSlots).forEach(this::connect);
+      LOG.info(
```
shemnon: This should be debug level.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/netty/NettyP2PNetwork.java
```diff
@@ -335,6 +339,32 @@ public void checkMaintainedConnectionPeers() {
...
+    if (availablePeerSlots > 0) {
+      final List<DiscoveryPeer> peers =
+          peerDiscoveryAgent
+              .getPeers()
+              .filter(peer -> peer.getEndpoint().getTcpPort().isPresent())
+              .filter(peer -> peer.getStatus() == PeerDiscoveryStatus.BONDED)
+              .filter(peer -> !isConnected(peer) && !isConnecting(peer))
+              .collect(Collectors.toList());
+      Collections.shuffle(peers);
+      LOG.info("Initiating connection to {} peers from the peer table", availablePeerSlots);
```
shemnon: This should be debug level.

ajsutton: Oh definitely, pretty sure both of those logs should be trace at best. :)

ajsutton:
Meredith picked this up and applied changes in a separate PR.

PR #907, opened 2019-02-19
IBFT to use VoteTallyCache

rain-on:
It was identified during a demonstration that Pantheon, when running
in IBFT would show a "Bad Block Import" when a validator was added 
removed from the validator pool.

It was determined this was due to IBFT maintaining a single, 'global'
copy of the curent list of validators, which was updated when a block
was imported - thus when a block which had been imported via IBFT
was then received via Eth block propogation, the validator list would
not align with the global list (as it had been updated in the IBFT
import).

The solution has been to utilise the VoteTallyCache as used in the
Clique implementation.


pantheon/src/main/java/tech/pegasys/pantheon/util/BlockImporter.java
```diff
@@ -162,7 +162,7 @@ private void extractSignatures(final Block block) {
     final BlockHeaderValidator<C> blockHeaderValidator = protocolSpec.getBlockHeaderValidator();
     final boolean validHeader =
         blockHeaderValidator.validateHeader(
-            header, previousHeader, context, HeaderValidationMode.FULL);
+            header, previousHeader, context, HeaderValidationMode.DETACHED_ONLY);
```
jframe: why did the validation mode need to be changed?

rain-on: This was a bug in Pantheon which was hidden up until this change - basically they do async block validation (thus block 2 may be validated before block 1) - this is illegal in IBFT, as we must validate a block fully with respect to its parent.

pantheon/src/main/java/tech/pegasys/pantheon/util/BlockImporter.java
```diff
@@ -81,7 +81,7 @@
         if (header.getNumber() == BlockHeader.GENESIS_BLOCK_NUMBER) {
           continue;
         }
-        if (header.getNumber() % 100 == 0) {
+        if (header.getNumber() % 1 == 0) {
```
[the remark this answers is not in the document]
rain-on: this was not meant to change, was used for debug, reverted.

pantheon/src/main/java/tech/pegasys/pantheon/util/BlockImporter.java
jframe: Checked and this was changed to fix a bug "the validation rule was a hidden bug in Pantheon - the BlockImporter validates blocks ASYNC - meaning, out of order :disappointed: Which IBFT can no longer do!
So when doing ASYNC, you MUST do only detached rules as the initial validation, then do the rest at the end"

PR #909, opened 2019-02-19
[PAN-2316] check permissions CLI dependencies

macfarla:
## PR description
if permissions config file is set, check that either nodes or accounts whitelisting is enabled.

## Fixed Issue(s)
fixes #PAN-2316


pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -552,6 +552,13 @@ public void run() {
               + "or specify the beneficiary of mining (via --miner-coinbase <Address>)");
     }
 
+    if (permissionsConfigFile != null) {
+      if (!permissionsAccountsEnabled && !permissionsNodesEnabled) {
+        throw new ParameterException(
```
Errorific: Do we want this to be an error? Convention here seems to be options that aren’t going to have an effect is a warning like on line 636.

macfarla: have posed the question for @jakehaugen

macfarla: changed to warning

PR #911, opened 2019-02-19
Incremental Optimization(s) on BlockBroadcaster

s-matthew-english:
## PR description

--> adapt `BlockBroadcaster` to use `send` method on `EthPeer`

--> adapt `EthProtocolMamnager` to utilize `BlockBroadcaster` to disseminate newly mined block


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/EthProtocolManager.java
```diff
@@ -271,24 +273,12 @@ private void handleStatusMessage(final EthPeer peer, final MessageData data) {
   @Override
   public void blockMined(final Block block) {
     // This assumes the block has already been included in the chain
-
-    final Optional<UInt256> totalDifficulty = blockchain.getTotalDifficultyByHash(block.getHash());
-    if (!totalDifficulty.isPresent()) {
+    final Optional<UInt256> maybeTotalDifficulty =
+        blockchain.getTotalDifficultyByHash(block.getHash());
+    if (!maybeTotalDifficulty.isPresent()) {
       throw new IllegalStateException(
```
ajsutton: I know this was pretty like this to start with but while you're here: `blockchain.getTotalDifficultyByHash(block.getHash()).orElseThrow(() -> new IllegalStateException(...))` would be simpler.

s-matthew-english: Yeah, for sure. I'll try to keep that pattern in mind in the future, it's not the first time you've given me this feedback

PR #931, opened 2019-02-20
[PAN-2198] Update GetNodeDataFromPeerTask to return a map

s-matthew-english:
## PR description

 Update `GetNodeDataFromPeerTask` to return map to node data by hash.


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/task/GetNodeDataFromPeerTask.java
```diff
@@ -83,6 +87,13 @@ protected ResponseStream sendRequest(final EthPeer peer) throws PeerNotConnected
       // Message contains unrequested data, must not be the response to our request.
```
mbaxter: The corresponding ticket for this task was badly written and didn't really mention this, but the main focus for this update should be to minimize the number of hash calculations we do to improve efficiency.  

So, I suggest combining this check on line 86 (which checks that the data in the response corresponds to the data requested) with the construction of the map in `mapNodeDataByHash`.  You could do this by writing a for loop that iterates over `nodeData`.  For each element in the list, you can calculate the hash and check if the hash is in `hashes`.  If not, you can return early without calculating all of the hashes.  If it is in the list, you can add it to your map, which will be returned only if you make it all the way through the loop.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/manager/task/GetNodeDataFromPeerTaskTest.java
```diff
@@ -23,42 +23,49 @@
...
   @Override
-  protected List<BytesValue> generateDataToBeRequested() {
-    final List<BytesValue> requestedData = new ArrayList<>();
+  protected Map<Hash, BytesValue> generateDataToBeRequested() {
+    final Map<Hash, BytesValue> requestedData = new HashMap<>();
     for (int i = 0; i < 3; i++) {
       final BlockHeader blockHeader = blockchain.getBlockHeader(10 + i).get();
-      requestedData.add(
+      requestedData.put(
+          Hash.fromHexStringLenient(Integer.toHexString(i)),
```
mbaxter: This should be the hash of the node data you're requesting rather than a random hash.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/manager/task/GetNodeDataFromPeerTaskTest.java
```diff
@@ -23,42 +23,49 @@
...
     return requestedData;
   }
 
   @Override
-  protected EthTask<PeerTaskResult<List<BytesValue>>> createTask(
-      final List<BytesValue> requestedData) {
-    final List<Hash> hashes = requestedData.stream().map(Hash::hash).collect(toList());
+  protected EthTask<PeerTaskResult<Map<Hash, BytesValue>>> createTask(
+      final Map<Hash, BytesValue> requestedData) {
+    final List<Hash> hashes = requestedData.values().stream().map(Hash::hash).collect(toList());
```
mbaxter: If you update line 38 above, you can then just use: `hashes = requestedData.keySet()`

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/manager/task/GetNodeDataFromPeerTaskTest.java
```diff
@@ -23,42 +23,49 @@
...
   @Override
   protected void assertPartialResultMatchesExpectation(
-      final List<BytesValue> requestedData, final List<BytesValue> partialResponse) {
+      final Map<Hash, BytesValue> requestedData, final Map<Hash, BytesValue> partialResponse) {
     assertThat(partialResponse.size()).isLessThanOrEqualTo(requestedData.size());
     assertThat(partialResponse.size()).isGreaterThan(0);
-    assertThat(requestedData).containsAll(partialResponse);
+    final List<BytesValue> requestData = new ArrayList<>(requestedData.values());
+    final List<BytesValue> resultData = new ArrayList<>(partialResponse.values());
+    assertThat(requestData).containsAll(resultData);
```
mbaxter: You're only checking the values here, but I think its worth checking that the values are mapped with the correct keys.  Pending updates to line 38, you should be over to iterate over the partial response keyValue pairs and then look up each value in `requestedData` and check that it matches.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/manager/task/GetNodeDataFromPeerTaskTest.java
```diff
@@ -23,42 +23,49 @@
...
   protected void assertResultMatchesExpectation(
-      final List<BytesValue> requestedData,
-      final PeerTaskResult<List<BytesValue>> response,
+      final Map<Hash, BytesValue> requestedData,
+      final PeerTaskResult<Map<Hash, BytesValue>> response,
       final EthPeer respondingPeer) {
-    assertThat(response.getResult()).containsExactlyInAnyOrderElementsOf(requestedData);
+    final List<BytesValue> requestData = new ArrayList<>(requestedData.values());
+    final List<BytesValue> resultData = new ArrayList<>(response.getResult().values());
+    assertThat(resultData).containsExactlyInAnyOrderElementsOf(requestData);
```
mbaxter: Same thing here as above: you should be able to check keys and values are as expected.  Should just be an equality check here ...

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/task/GetNodeDataFromPeerTask.java
```diff
@@ -63,26 +65,27 @@ protected ResponseStream sendRequest(final EthPeer peer) throws PeerNotConnected
...
+      return Optional.of(emptyMap());
     }
     final NodeDataMessage nodeDataMessage = NodeDataMessage.readFrom(message);
     final List<BytesValue> nodeData = nodeDataMessage.nodeData();
-    if (nodeData.isEmpty()) {
-      return Optional.empty();
-    } else if (nodeData.size() > hashes.size()) {
-      // Can't be the response to our request
-      return Optional.empty();
-    }
```
mbaxter: You want to keep this check here:
```
    if (nodeData.size() > hashes.size()) {
      // Can't be the response to our request
      return Optional.empty();
    }
```

PR #938, opened 2019-02-21
[PRIV] Complete Private Transaction Processor

iikirilov:
## PR description

Complete wire-up and implementation of private transaction processor. Includes acceptance test for private transactions.

## Fixed Issue(s)
PRIV-16
PRIV-17
PRIV-21


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/PrivacyParameters.java
```diff
@@ -14,26 +14,41 @@
...
 import java.io.IOException;
 import java.net.URI;
+import java.nio.file.Path;
 
 import com.google.common.io.Files;
 
 public class PrivacyParameters {
   private static final String ENCLAVE_URL = "http://localhost:8888";
   public static final URI DEFAULT_ENCLAVE_URL = URI.create(ENCLAVE_URL);
+  String PRIVATE_DATABASE_PATH = "private";
```
lucassaldanha: Is this field used or updated anywhere else that would prevent it being private and final?

PR #939, opened 2019-02-21
[PAN-2312] Validate DAO block

mbaxter:
## PR description
This PR adds a `DaoForkPeerValidator` that will run for every connected eth peer when pantheon is run against mainnet.  The validator checks if the peer is ready to be validated by looking at the peer's estimated chain height.  If the height is past the DAO block, it will request and check the block from the peer at the DAO fork.  If this block is on the wrong side of the fork, the peer will be disconnected. 

Some secondary changes in this PR include:
- Updating `BlockHeadersMessage.getHeaders` to return a list
- Updating `AbstractGetHeadersFromPeerTask` to perform stricter matching on responses:  if a response contains more headers than were requested, we don't consider this response a match to our request
- Making the timeout in `PeerRequestTask` configurable
- Updating some test utilities to provide more control over the execution of executor tasks within tests


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/task/AbstractGetHeadersFromPeerTask.java
```diff
@@ -73,13 +72,17 @@ protected AbstractGetHeadersFromPeerTask(
     }
 
     final BlockHeadersMessage headersMessage = BlockHeadersMessage.readFrom(message);
-    final Iterator<BlockHeader> headers = headersMessage.getHeaders(protocolSchedule);
-    if (!headers.hasNext()) {
+    final List<BlockHeader> headers = headersMessage.getHeaders(protocolSchedule);
+    if (headers.size() == 0) {
```
ajsutton: nit:
```suggestion
    if (headers.isEmpty()) {
```

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/peervalidation/PeerValidatorRunner.java
```diff
@@ -0,0 +1,67 @@
...
+      peerValidator
+          .validatePeer(ethPeer)
+          .whenComplete(
+              (validated, err) -> {
+                if (err != null || !validated) {
+                  // Disconnect invalid peer
+                  disconnectPeer(ethPeer);
+                }
+              });
+    } else {
```
ajsutton: Do we need this to be `} else if (!peer.isDisconnected()) {`?

pantheon/src/main/java/tech/pegasys/pantheon/controller/MainnetPantheonController.java
```diff
@@ -116,6 +123,9 @@ public MainnetPantheonController(
             metricsSystem);
     final SyncState syncState =
         new SyncState(blockchain, ethProtocolManager.ethContext().getEthPeers());
+    final LabelledMetric<OperationTimer> ethTasksTimer =
+        metricsSystem.createLabelledTimer(
+            MetricCategory.SYNCHRONIZER, "task", "Internal processing tasks", "taskName");
```
ajsutton: What was the reason for pushing this up to here?  Seems odd to have only done it for `MainnetPantheonController` and not clique & IBFT, plus we're generally trying to push common construction code out of the controller.

mbaxter: Because we need it to construct any `EthTask`'s and `DaoForkPeerValidator` uses an `EthTask`.  I think the real fix is to keep working on `MetricsSystem` to allow metrics to be created idempotently so we can then push this timer creation into `AbstractEthTask` where it's used.  I have some in-progress work on that which I can come back to.

ajsutton: ahhh, yeah that's very unfortunate.

ajsutton: Just remembered, I added some caching for metrics back when I did the EVM metrics but had to revert it, so https://github.com/PegaSysEng/pantheon/commit/525aa19d5075fa376341715c19a15152e1554642#diff-b95472676d731050caa643c3ebe592b1 may be useful.

PR #940, opened 2019-02-21
RocksDbQueue Threading Tweaks

ajsutton:
## PR description
Since `RocksDbQueue` uses `synchronized` for its methods, it doesn't need to also use `AtomicLong` and `AtomicBoolean`.  The `close` method wasn't synchronized but needs to be since it has to access the `dequeueIterator` field to close it.


services/queue/src/main/java/tech/pegasys/pantheon/services/queue/RocksDbTaskQueue.java
```diff
@@ -37,13 +36,13 @@
...
-  private final AtomicLong lastDequeuedKey = new AtomicLong(0);
+  private long lastEnqueuedKey = 0;
+  private long lastDequeuedKey = 0;
   private RocksIterator dequeueIterator;
-  private final AtomicLong oldestKey = new AtomicLong(0);
+  private long oldestKey = 0;
   private final Set<RocksDbTask<T>> outstandingTasks = new HashSet<>();
 
-  private final AtomicBoolean closed = new AtomicBoolean(false);
+  private boolean closed = false;
```
mbaxter: Do these vars need to be `volatile`?

ajsutton: No because every access to them is within a synchronized section.

mbaxter: Still getting up to speed on threading obviously :) . I didn't realize `synchronized` causes all the vars to be updated from main memory.

ajsutton: Yep, it has a read-fence on entry to the `synchronized` block and a write one on exit.  So as long as both the writing thread and the reading thread used synchronized there's no need for volatile.

PR #943, opened 2019-02-22
[MINOR] Eliminate redundant header validation

s-matthew-english:
## PR description

wrt block propagation -- currently we're validating headers twice: once to decide if we should propagate and once when we actually import.  

changes here enforce that we validate just once, then propagate and import in parallel.

### DIFF
--- a/ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockPropagationManager.java
+++ b/ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockPropagationManager.java
@@ -237,19 +237,6 @@ public class BlockPropagationManager<C> {
     return getBlockTask.run().thenCompose((r) -> importOrSavePendingBlock(r.getResult()));
   }
 
-  Boolean validateHeader(final Block block, final BlockHeader parent) throws Exception {
-    final ProtocolSpec<C> protocolSpec =
-        protocolSchedule.getByBlockNumber(block.getHeader().getNumber());
-    final BlockHeaderValidator<C> blockHeaderValidator = protocolSpec.getBlockHeaderValidator();
-    return ethContext
-        .getScheduler()
-        .scheduleComputationTask(
-            () ->
-                blockHeaderValidator.validateHeader(
-                    block.getHeader(), parent, protocolContext, HeaderValidationMode.FULL))
-        .get();
-  }
-
   private void broadcastBlock(final Block block, final BlockHeader parent) {
     final UInt256 totalDifficulty =
         protocolContext
@@ -260,7 +247,6 @@ public class BlockPropagationManager<C> {
     blockBroadcaster.propagate(block, totalDifficulty);
   }
 
-  @SuppressWarnings("LocalCanBeFinal")
   @VisibleForTesting
   CompletableFuture<Block> importOrSavePendingBlock(final Block block) {
     // Synchronize to avoid race condition where block import event fires after the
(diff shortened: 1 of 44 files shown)


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockPropagationManager.java
```diff
@@ -292,29 +276,61 @@ private void handleNewBlockHashesFromNetwork(final EthMessage message) {
...
+
+    final ProtocolSpec<C> protocolSpec =
+        protocolSchedule.getByBlockNumber(block.getHeader().getNumber());
+    final BlockHeaderValidator<C> blockHeaderValidator = protocolSpec.getBlockHeaderValidator();
 
     // Import block
     final PersistBlockTask<C> importTask =
         PersistBlockTask.create(
-            protocolSchedule, protocolContext, block, HeaderValidationMode.FULL, metricsSystem);
+            protocolSchedule, protocolContext, block, HeaderValidationMode.NONE, metricsSystem);
```
ajsutton: Can we create this inside the if(validate) block so it's only created if it will actually be used? Mostly just so that the first thing you come across isn't `HeaderValidationMode.NODE` which is a bit scary. :)

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockPropagationManager.java
```diff
@@ -292,29 +276,61 @@ private void handleNewBlockHashesFromNetwork(final EthMessage message) {
...
+
     return ethContext
         .getScheduler()
-        .scheduleSyncWorkerTask(importTask::run)
-        .whenComplete(
-            (r, t) -> {
-              importingBlocks.remove(block.getHash());
-              if (t != null) {
+        .scheduleSyncWorkerTask(
+            () -> {
```
ajsutton: There's probably a few methods that should be extracted here to make it easier to read.  I'd probably have the whole of this worker task as a private method (maybe `validateAndProcessPendingBlock`?)

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockPropagationManager.java
```diff
@@ -292,29 +276,61 @@ private void handleNewBlockHashesFromNetwork(final EthMessage message) {
...
+              if (blockHeaderValidator.validateHeader(
+                  block.getHeader(), parent, protocolContext, HeaderValidationMode.FULL)) {
+                ethContext
+                    .getScheduler()
+                    .scheduleSyncWorkerTask(() -> broadcastBlock(block, parent));
+                return ethContext
```
ajsutton: I'd be tempted to extract this blob that does the actual import into a separate method as well.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockPropagationManager.java
```diff
@@ -292,29 +276,61 @@ private void handleNewBlockHashesFromNetwork(final EthMessage message) {
...
+                    .getScheduler()
+                    .scheduleSyncWorkerTask(importTask::run)
```
ajsutton: We're already in a worker thread so can remove `scheduleSyncWorkerTask` and just call `run` directly.  It still returns a `CompletableFuture<Block>` so we'd still chain the `whenComplete` in the same way.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockPropagationManager.java
```diff
@@ -292,19 +276,54 @@ private void handleNewBlockHashesFromNetwork(final EthMessage message) {
...
+  }
+
+  private CompletableFuture<Block> validateAndProcessPendingBlock(
+      final BlockHeaderValidator<C> blockHeaderValidator,
+      final Block block,
+      final BlockHeader parent) {
+    if (blockHeaderValidator.validateHeader(
+        block.getHeader(), parent, protocolContext, HeaderValidationMode.FULL)) {
+      ethContext.getScheduler().scheduleSyncWorkerTask(() -> broadcastBlock(block, parent));
+      return ethContext.getScheduler().scheduleSyncWorkerTask(() -> runImportTask(block));
```
ajsutton: You don't need to defer this to a separate worker task since you're already in a worker thread.
```suggestion
      return runImportTask(block);
```

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockPropagationManager.java
```diff
@@ -292,19 +276,54 @@ private void handleNewBlockHashesFromNetwork(final EthMessage message) {
...
+
+  private CompletableFuture<Block> runImportTask(final Block block) {
+    final PersistBlockTask<C> importTask =
+        PersistBlockTask.create(
+            protocolSchedule, protocolContext, block, HeaderValidationMode.NONE, metricsSystem);
+    return importTask
+        .run()
         .whenComplete(
-            (r, t) -> {
+            (result, throwable) -> {
```
ajsutton: +1 to the better naming here. :)

PR #944, opened 2019-02-22
RLP encode subcommand

NicolasMassart:
## PR description

Add RLP encoding subcommand for at least ibftExtraData to be inserted in genesis file.

There's only one encodable type for the moment, however the `--type` option is already there with a default value on this type. We'll add the more possible types later if needed and the current type will remain the default one.

The file output is optional and it writes to standard output by default, same for
standard input or a file.

Tests and doc changes are added.

## Fixed Issue(s)

fixes PAN-2318


pantheon/src/main/java/tech/pegasys/pantheon/cli/RLPSubCommand.java
```diff
@@ -0,0 +1,180 @@
...
+        names = "--from",
+        paramLabel = MANDATORY_FILE_FORMAT_HELP,
+        description = "File containing JSON object to encode",
+        arity = "1..1")
+    private File jsonSourceFile = null;
+
+    @Option(
+        names = "--to",
+        paramLabel = MANDATORY_FILE_FORMAT_HELP,
+        description = "File to write encoded RPL string to.",
```
rain-on: RLP not RPL

pantheon/src/main/java/tech/pegasys/pantheon/cli/RLPSubCommand.java
```diff
@@ -0,0 +1,180 @@
...
+    private void encode(final String jsonInput) {
+      // map the json to the object matching the type option
+      // the object must be an RLPEncodable object
+      if (jsonInput == null || jsonInput.isEmpty()) {
+        throw new ExecutionException(
+            spec.commandLine(), "An error occurred while trying to read the JSON data.");
+      } else {
+        try {
+          JsonObject jsonObject = new JsonObject(jsonInput);
+          RLPEncodable objectToEncode = jsonObject.mapTo(IbftExtraData.class);
```
rain-on: does this mean people's json text must match the field names of the IbftExtraData class?
Are there good error messages showing what the content should be (i.e. mismatching field names etc?)

rain-on: Do we make them put in the whole of extra data, or just the bits that are of interest? I.e. for genesis file the round should be 0, the seals should be empty, ideally the validators are ordered.

pantheon/src/main/java/tech/pegasys/pantheon/cli/RLPSubCommand.java
```diff
@@ -0,0 +1,180 @@
...
+              spec.commandLine(), "Unable to load the JSON data. Please check JSON input format.");
+        } catch (IllegalArgumentException e) {
+          throw new ParameterException(
+              spec.commandLine(),
+              "Unable to map the JSON data with IbftExtraData type. Please check JSON input format.");
+        }
+      }
+    }
+
+    private void writeOutput(final BytesValue rlpEncodedOutput) {
```
rain-on: should we also produce the bootnode information? Or is there another ticket to create that?

NicolasMassart: It's just the genesys extra data part here. What's the bootnode information ?

pantheon/src/main/java/tech/pegasys/pantheon/cli/RLPSubCommand.java
NicolasMassart: I will change that with the adapter, being able to just use an array.

NicolasMassart: for the moment yes, but will change with the adapter.

pantheon/src/main/java/tech/pegasys/pantheon/cli/RLPSubCommand.java
NicolasMassart: oups

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/IbftExtraData.java
```diff
@@ -120,15 +125,15 @@ private BytesValue encode(final EncodingType encodingType) {
     return encoder.encoded();
   }
 
+  @JsonCreator
+  public static IbftExtraData createGenesisExtraData(
```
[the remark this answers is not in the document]
NicolasMassart: will be extracted to an adapter

consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/IbftExtraData.java
```diff
@@ -27,11 +28,14 @@
...
 
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
 /**
  * Represents the data structure stored in the extraData field of the BlockHeader used when
  * operating under an IBFT 2.0 consensus mechanism.
  */
-public class IbftExtraData {
+public class IbftExtraData implements RLPEncodable {
```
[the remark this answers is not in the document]
NicolasMassart: as @mbaxter also suggested it's what I will try to do.

pantheon/src/main/java/tech/pegasys/pantheon/cli/rlp/IbftExtraDataCLIAdapter.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.stream.Collectors;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ * Adapter to convert a typed JSON to an IbftExtraData object This adapter handles the JSON to RLP
+ * encoding
+ */
+public class IbftExtraDataCLIAdapter implements JSONToRLP {
```
shemnon: Instead of an extra interface with an adapter, I think it would be more concise and direct to move the method body into the enum type and add an "encode" method on that enum:

https://github.com/shemnon/pantheon/commit/e178bafac4786084cbeebdb5ded146fb952b47ed

NicolasMassart: The thing is that for now we only have one type in the enum but we may end up with more and I wanted to keep the enum as simple as possible by just referencing an adapter and have one adapter for each data type, well isolated into each class file. But why not. What would you think @mbaxter and @rain-on ?

shemnon: You can use the same brace-initialized-enum-value pattern for future encoders, so this doesn't preclude more values in the enum.

NicolasMassart: No it doesn't but the more values we'll have in the enum the more code we'll have and I'm happier with small files when possible. But I admit your proposal removes a class and an interface so why not. Waiting for a second advice.

s-matthew-english: keeping the enum simple makes sense to me

NicolasMassart: Indeed I would like to try to start splitting things to prevent mega classes like PantheonCommand. CLI is usually the place where you enter Pantheon so having it very easy to navigate seems important to me. So yes it requires a bit more classes and interfaces but it also makes it easier for someone new to understand and add a feature in my opinion.

PR #950, opened 2019-02-24
Single threaded world state persistence

ajsutton:
## PR description
Previously the world state downloader used multiple different threads to persist the data into RocksDB. This often led to timeouts acquiring the write lock from RocksDB and the data would be lost. Now a single service thread is used to do all persistence, avoiding lock contention.

Also includes a fix for `RocksDbTaskQueue` to ensure it handles data that may have existed on disk before it was created correctly (by ignoring it consistently).


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/worldstate/WorldStateDownloader.java
```diff
@@ -373,4 +403,58 @@ private synchronized void markDone() {
...
+              markDone();
+              result.get().complete(null);
+            } else {
+              ethContext.getScheduler().scheduleSyncWorkerTask(() -> requestNodeData(header));
+            }
+          }
+        } catch (final InterruptedException ignore) {
+          Thread.currentThread().interrupt();
+        } catch (final RuntimeException e) {
+          LOG.error("Unexpected error while persisting world state", e);
```
ekellstrand: Consider logging some context for this exception so that it is easier to track down (eg header, etc)

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/worldstate/TrieNodeDataRequest.java
```diff
@@ -58,7 +59,11 @@
   private Stream<NodeDataRequest> getRequestsFromChildTrieNode(final Node<BytesValue> trieNode) {
     if (trieNode.isReferencedByHash()) {
       // If child nodes are reference by hash, we need to download them
-      NodeDataRequest req = createChildNodeDataRequest(Hash.wrap(trieNode.getHash()));
+      final Hash hash = Hash.wrap(trieNode.getHash());
+      if (hash.equals(MerklePatriciaTrie.EMPTY_TRIE_NODE_HASH) || hash.equals(BytesValue.EMPTY)) {
```
ekellstrand: consider inverting the ".equals(...) statements so that you never have to worry about the hash being null.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/worldstate/WorldStateDownloader.java
```diff
@@ -178,28 +205,21 @@ private void requestDataFromPeer(final BlockHeader header, final EthPeer peer) {
...
-                final Updater updater = worldStateStorage.updater();
-                updater.putAccountStateTrieNode(header.getStateRoot(), rootNode);
-                updater.commit();
-                markDone();
-              } else {
-                // Send out additional requests
-                requestNodeData(header);
+              if (error != null
+                  && !(ExceptionUtils.rootCause(error) instanceof RejectedExecutionException)) {
+                LOG.error("World state data request failed", error);
```
ekellstrand: consider adding some context to the log message to make debugging easier (eg peer, toRequest, header, etc)

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/worldstate/TrieNodeDataRequest.java
ajsutton: We literally just created the hash on the line above. It can't possibly be null.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/worldstate/WorldStateDownloader.java
ajsutton: Again there's no useful context we can include here beyond the exception.  This is something completely unexpected that happened somewhere through the entire create, send and process request.  It shouldn't ever happen but if it does we want to see the stack trace. Most errors are already handled in the details - this is the equivalent of the uncaught exception handler.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/worldstate/TrieNodeDataRequest.java
shemnon: for constants it is a solid pattern.  even if in this case it is impossible (until hash.wrap starts returning nulls)

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/worldstate/WorldStateDownloader.java
shemnon: If we had what eric asked for at debug or trace it would help to re-create a reproducible state.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/worldstate/WorldStateDownloader.java
ajsutton: We don't really have any context here - there's only one world state download so the header is useless and we're attempting to persist about 394 data points - far too much to include in a log.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/worldstate/WorldStateDownloader.java
ajsutton: What log message are you actually expecting to see here? I've been working in this code a fair bit and can't imagine what we could add here that would actually be useful.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/worldstate/TrieNodeDataRequest.java
ajsutton: If we're requesting data by hash with a null hash we're screwed *long* before here.

shemnon: `CONSTANT.equals(variable)` is a pattern that keeps us from thinking "have I initialized `variable`" While in this one case it may be true we shouldn't have to think about it when we use the constant first pattern.  It's also a rule for Google's internal Java readability process (separate form the code formatting guides).

ajsutton: I'm aware of the pattern but not having to speak like Yoda is really quite nice when there's absolutely no way it can possibly be null.  But just to get out of this discussion I've flipped it round. :)

PR #951, opened 2019-02-24
Cleanup Ibft executors

rain-on:
Explicitly clean up executors used for IBFT timers.


pantheon/src/main/java/tech/pegasys/pantheon/controller/IbftPantheonController.java
```diff
@@ -214,21 +215,20 @@
 
     final IbftGossip gossiper = new IbftGossip(uniqueMessageMulticaster);
 
+    final ScheduledExecutorService timerExecutor = Executors.newSingleThreadScheduledExecutor();
```
jframe: Could be good idea to use the MonitoredExecutors.newFixedThreadPool so we can give our thread pool a name and get some free metrics as well.

pantheon/src/main/java/tech/pegasys/pantheon/controller/IbftPantheonController.java
```diff
@@ -286,6 +286,12 @@
           } catch (final InterruptedException e) {
             LOG.error("Failed to shutdown ibft processor executor");
           }
+          timerExecutor.shutdownNow();
+          try {
+            timerExecutor.awaitTermination(5, TimeUnit.SECONDS);
+          } catch (final InterruptedException e) {
+            LOG.error("Failed to shutdown timerexecutor");
```
jframe: nit: thinking this should be "Failed to shutdown timer executor"

rain-on: done.

pantheon/src/main/java/tech/pegasys/pantheon/controller/IbftPantheonController.java
rain-on: done.

PR #955, opened 2019-02-25
Create peer discovery packets on a worker thread

ajsutton:
## PR description

Creating peer discovery packets involves cryptographically signing the contents which is a fairly expensive operation.  When we do that for a bunch of messages at once, such as when bonding to newly discovered peers, it can cause the Vertx event thread to be blocked for multiple seconds, causing errors and slow processing of incoming messages.

This change moves creation and signing of packets onto worker threads so that the event queue isn't blocked but ensures that the created packet is then delivered back on the event queue to maintain the single-threaded behaviour the code requires.


shemnon:
LGTM if those who are vertx familiar like it.

PR #957, opened 2019-02-25
Fix bootnode docs

shemnon:
## PR description

fix bootnodes option, it needs an explicitly empty value and not just a single flag style option.


shemnon:
nevermind, didn't see the `arity = "0..*",` option on bootnodes before this.

PR #960, opened 2019-02-25
CLI stack traces when debugging

shemnon:
## PR description

Sometimes we get very short and unhelpful exceptions from deep in
pantheon ('null' for example).  This change makes it so that if
--logging is at DEBUG, TRACE, or ALL then the stack trace that caused
the exit will also be printed.

This is done by providing our own impl of defaultExceptionHandler that
is aware of the logLevel in PantheonCommand.


shemnon:
*not* 1.0 critical.

PR #966, opened 2019-02-26
Gradle dependency for JMH annotation, for IDEs that aren't IntelliJ

CjHare:
## Fixed Issue(s)
EclipseIDE dependency issues for the `queue` project.


CjHare:
Building Pantheon with a headless EcliseIDE could be added as part of the smoke-test phase (idea from @Errorific )

PR #980, opened 2019-02-26
Fix benchmark compilation

ajsutton:
## PR description
Add compileJmh to the build process to ensure that benchmark code at least compiles.  Fix the compilation error.


Jenkinsfile
```diff
@@ -54,7 +54,7 @@ try {
                 docker.image(build_image).inside("--link ${d.id}:docker") {
                     try {
                         stage(stage_name + 'Prepare') {
-                            sh './gradlew --no-daemon --parallel clean compileJava compileTestJava assemble'
+                            sh './gradlew --no-daemon --parallel clean compileJava compileTestJava compileJmh assemble'
```
Errorific: Change also needs to happen on lines 81, 105, & 135

ajsutton: I considered that but we actually only need one task to do the compilation - the results aren't used so no point in having them all do it.

ajsutton: Actually may be better to make it its own stage like javadoc is...

PR #984, opened 2019-02-26
Fix for bug in the validation of Round-Change-Certificates

saltiniroberto:
## PR description
The Round-Change-Certificate must contain at least Quorum(n) Round-Change messages sent by **different** validators
The current code does not verify this condition, but only that the Round-Change-Certificate contains at least Quorum(n) Round-Change messages sent by validators.
This PR fixes the issue.
**Note: Tests must be added**

### DIFF
--- a/consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/validation/RoundChangeCertificateValidator.java
+++ b/consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/validation/RoundChangeCertificateValidator.java
@@ -29,6 +29,7 @@ import tech.pegasys.pantheon.ethereum.core.Block;
 
 import java.util.Collection;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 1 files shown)


consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/validation/RoundChangeCertificateValidator.java
```diff
@@ -58,7 +59,8 @@ public boolean validateRoundChangeMessagesAndEnsureTargetRoundMatchesRoot(
     final Collection<SignedData<RoundChangePayload>> roundChangeMsgs =
         roundChangeCert.getRoundChangePayloads();
 
-    if (roundChangeMsgs.size() < quorum) {
+    if (roundChangeMsgs.stream().map(SignedData::getAuthor).collect(Collectors.toSet()).size()
```
jframe: This can be written a bit more simply: roundChangeMsgs.stream().map(SignedData::getAuthor).distinct().count()

rain-on: we should create a test which validates the fix (which appears necessary!)

saltiniroberto: Nice...I didn't know about `distinct()`

saltiniroberto:
Replaced by #989

PR #988, opened 2019-02-26
Add jmh compile to jenkins checkins tests

shemnon:
## PR description

JMH compilation should be part of the PR checkin process (and break the build if needed).

Fix existing JMH errors.


shemnon:
Dup of #980

PR #989, opened 2019-02-26
RoundChangeCertificateValidator requires unique authors

rain-on:


consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/validation/RoundChangeCertificateValidator.java
```diff
@@ -84,6 +88,18 @@ public boolean validateRoundChangeMessagesAndEnsureTargetRoundMatchesRoot(
     return true;
   }
 
+  private boolean validateDistinctAuthors(
```
CjHare: You can name this with the negation in the title (negate the need for the bang in the if).

e.g. hasDuplicateAuthors()

PR #996, opened 2019-02-27
[BUG] remove 127.0.0.1 from default hosts configs

Errorific:
in PR #974 it added 127.0.0.1 to the host whitelist to make tests work under more environments. Adding this to the whitelist opens up a security vulnerability. This reverts that then changes the default setup for hosts to use the localhost loopback name.


Errorific:
after discussion 127.0.0.1 is safe as a whitelisted host

PR #1002, opened 2019-02-27
[PAN-2367] Fix race condition in WebSocket AT

lucassaldanha:


acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/pubsub/Subscription.java
```diff
@@ -44,8 +45,9 @@ public void verifyEventReceived(final Hash expectedTransaction) {
   }
 
   public void verifyEventReceived(final Hash expectedTransaction, final int expectedOccurrences) {
+    waitFor(() -> assertThat(connection.getSubscriptionEvents()).isNotEmpty());
```
ajsutton: I'd have expected the whole method to be inside the waitFor.  So we might have received an event we weren't waiting for but need to keep waiting for the one we do want.

PR #1004, opened 2019-02-27
[PRIV] Generate Private Contract Address

vinistevam:
## PR description
Create a new method to use SenderAddress, nonce and PrivacyGroup to generate the private address.
Remove the resolveAccountFromPublicState.

## Fixed Issue(s)

[PRIV-25](https://pegasys1.atlassian.net/browse/PRIV-25)


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/Address.java
```diff
@@ -129,6 +129,30 @@ public static Address contractAddress(final Address senderAddress, final long no
...
+   *
+   * <p>This implement equation (86) in Section 7 of the Yellow Paper (rev. a91c29c).
+   *
+   * @param senderAddress the address of the transaction sender.
+   * @param nonce the nonce of this transaction.
+   * @param privacyGroup hash of participants list ordered from Enclave response.
+   * @return The generated address of the created private contract.
+   */
+  public static Address privateContractAddress(
+      final Address senderAddress, final long nonce, final int privacyGroup) {
```
lucassaldanha: As per our Slack discussion, the privacyGroupID should be calculated as:
`keccak256(RLP(address1, address2, ..., addressN))`

Should we change this method signature now or is this something that will be refactored later?

iikirilov: we will change it now

PR #1005, opened 2019-02-27
[PAN-2369] RunnerTest fail on Windows due to network startup timing issue

ekellstrand:
## PR description
[PAN-2369] RunnerTest fail on Windows due to network startup timing issue

- Forces RunnerTest to wait for the P2PNetwork to start and actually configure it's ports before asking for them.
- Re-Enables RunnerTest.fullSyncFromGenesis()
- Re-Enables RunnerTest.fastSyncFromGenesis()

## Fixed Issue(s)
[PAN-2369]

[PAN-2369]: https://pegasys1.atlassian.net/browse/PAN-2369
[PAN-2369]: https://pegasys1.atlassian.net/browse/PAN-2369

### DIFF
--- a/pantheon/src/test/java/tech/pegasys/pantheon/RunnerTest.java
+++ b/pantheon/src/test/java/tech/pegasys/pantheon/RunnerTest.java
@@ -157,11 +157,11 @@ public final class RunnerTest {
 
       // Wait for network to initialize to get the P2P UDP port
       Awaitility.await()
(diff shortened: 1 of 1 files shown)


ekellstrand:
In this case, slow is fine.  The timing issue only seems to happen on fast boxes with lots of cores.  And in my case, it was reliably only a couple of ms required.  However, I'm not opposed to 20 sec, to that'll be coming right up.

PR #1011, opened 2019-02-28
Remove full hashimoto implementation as its never used

ajsutton:
## PR description
EthHasher.Full is never used except for a single ignored test.  Remove it.


shemnon:
Is there a reason we would ever use it?  And if there is what's the difficulty of going back into the git commits, other than the possibility we all forgot about it or we've had a full turnover of staff?

ajsutton:
The lack of testing is probably the biggest reason to remove it but it's also nice to just get rid of code that's not used, as Danno said it's easy to pull back out of git history should we need it in the future.

PR #1012, opened 2019-02-28
[PAN-2373] admin_nodeInfo rpc-api

shemnon:
## PR description

Implement admin_nodeInfo RPC call.  This involved bringing data from the
ethNetworkConfig and genesis json into the rpc call, so it had a lot of
testing side effects.

## Fixed Issue(s)

PAN-2373


ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/JsonRpcMethodsFactory.java
```diff
@@ -139,6 +146,8 @@
 
   public Map<String, JsonRpcMethod> methods(
       final String clientVersion,
+      final int networkId,
+      final JsonNode genesisConfigConfig,
```
ajsutton: genesisConfigConfig?

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/AdminNodeInfo.java
```diff
@@ -0,0 +1,113 @@
...
+      response.put("ports", ImmutableMap.of("discovery", port, "listener", port /*??*/));
+
+      final ChainHead chainHead = blockchainQueries.getBlockchain().getChainHead();
+      response.put(
+          "protocols",
+          ImmutableMap.of(
+              "eth",
+              ImmutableMap.of(
+                  "config",
+                  genesisConfigConfig,
```
ajsutton: Not entirely sure what the right answer is here but I don't particularly like passing the "raw" json around so would be nice to use either the parsed `GenesisConfigFile` or `GenesisConfigOptions` here.  That would then also ensure we return Pantheon's understanding of the genesis config, not just the raw text which avoids confusion when some setting isn't taking effect because it's misspelt as it wouldn't appear here at all. 

I'm not sure if that justifies the extra complexity of having to reserialize the JSON or if there would be some reason to want to have unsupported options funnelled through to here.

pantheon/src/main/java/tech/pegasys/pantheon/RunnerBuilder.java
```diff
@@ -360,10 +364,19 @@ private FilterManager createFilterManager(
       final FilterManager filterManager,
       final Optional<AccountWhitelistController> accountWhitelistController,
       final PrivacyParameters privacyParameters) {
+    JsonNode genesisConfigConfig;
+    try {
+      genesisConfigConfig =
+          new ObjectMapper().readTree(ethNetworkConfig.getGenesisConfig()).get("config");
+    } catch (final IOException ioe) {
+      genesisConfigConfig = null;
+    }
```
ajsutton: Definitely not keen on having to parse the JSON a second time here (as opposed to just in `GenesisConfigFile.fromConfig`).

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/AdminNodeInfo.java
shemnon: The main question is do we want the literal config elements in the json text or do we want Pantheon's understanding of the options?  Since we are doing live nodeInfo I think we want the understanding.  Wiring up json de/serialization for the relevant interfaces should do the trick.

pantheon/src/main/java/tech/pegasys/pantheon/RunnerBuilder.java
shemnon: The json de/serialization will address this.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/JsonRpcMethodsFactory.java
shemnon: The config element of the genesisConfig.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/AdminNodeInfo.java
shemnon: Another option is to ask JsonGenesisConfigOptions for its configRoot, which is already the type we need.  But that is the unparsed and uninterpreted values.

ajsutton: Yeah I was wondering that.  The big downside is that it's not quite the raw values (we normalise all keys to lower case) and not just what pantheon understands either.

shemnon: Roger.  I think I can add an interface method to PantheonController to return the GenesisConfigOptions.

config/src/main/java/tech/pegasys/pantheon/config/CliqueConfigOptions.java
```diff
@@ -34,4 +37,15 @@ public long getEpochLength() {
...
   }
+
+  Map<String, Object> asMap() {
+    final ImmutableMap.Builder<String, Object> builder = ImmutableMap.builder();
+    if (cliqueConfigRoot.containsKey("epochlength")) {
+      builder.put("epochLength", getEpochLength());
+    }
+    if (cliqueConfigRoot.containsKey("blockperiodseconds")) {
+      builder.put("blockPeriodSeconds", getBlockPeriodSeconds());
+    }
```
ajsutton: Getting pedantic now but I wonder if we should always include this with the default value if none is set?  Would make this a really useful API for what values are actually in effect.

shemnon: All of our existing json files have them, and a default is not called out in the spec.  Consensus specs are a bad place for "this default that we have all agreed to by convention" values, so I'll always report them.  Also, it makes the code cleaner. :)

PR #1016, opened 2019-02-28
[PAN-2302] Resume world state download from existing queue

mbaxter:
## PR description
This PR updates `RocksDbTaskQueue` so that on construction, any existing data on disk will be read and accounted for.  Previously, the queue would always be empty on initialization.  Now that the queue can be resumed, the `WorldStateDownloader` modifies its behavior when started with a non-empty queue.  Rather than requesting the root node when the download begins, this node will only be requested when all other nodes have been processed.  This prevents us from queueing the root node multiple times each time the download is resumed.

Additionally, some trie node utility methods were made more generic and moved from the tests into `TrieNodeDecoder`, which lead to the deduplication of some logic in `TrieNodeDataRequest`.


ethereum/trie/src/test/java/tech/pegasys/pantheon/ethereum/trie/TrieNodeDecoderTest.java
```diff
@@ -0,0 +1,224 @@
...
+    trie.commit(tx::put);
+    tx.commit();
+
+    List<Node<BytesValue>> result =
+        TrieNodeDecoder.breadthFirstDecoder(storage::get, trie.getRootHash())
+            .collect(Collectors.toList());
+    assertThat(result.size()).isEqualTo(1);
+    assertThat(result.get(0).getValue()).contains(BytesValue.of(1));
+    // TODO: fix path representation
+    // assertThat(result.get(0).getPath()).isEqualTo(BytesValue.fromHexString("0x100000"));
```
ajsutton: Does this need looking at still?

mbaxter: Yeah - I'll make a ticket for this.  It's an existing issue.

mbaxter: Looks like this isn't actually a "bug" - the path value is just handled in an idiosyncratic way.  Updated the test.

PR #1019, opened 2019-03-01
[PAN-2386] Report the correct tcp port in PING packets when it differs from the UDP port

ajsutton:
## PR description
When Pantheon is started with `--p2p-port=0` an available port is selected for the UDP peer discovery and TCP peer connections but there is no guarantee this port will be the same.  Previously Pantheon would assume the UDP and TCP ports were the same so send the UDP port in place of the TCP port in PING messages, making it impossible for other peers to connect to it as they would use the wrong port.


ethereum/p2p/src/test/java/tech/pegasys/pantheon/ethereum/p2p/discovery/PeerDiscoveryTestHelper.java
```diff
@@ -145,7 +145,7 @@ public MockPeerDiscoveryAgent startDiscoveryAgent(
 
   public MockPeerDiscoveryAgent startDiscoveryAgent(final AgentBuilder agentBuilder) {
     final MockPeerDiscoveryAgent agent = createDiscoveryAgent(agentBuilder);
-    agent.start();
+    agent.start(65500).join();
```
mbaxter: You could use `PeerDiscoveryTestHelper.nextAvailablePort` here.  Or you could try grabbing the discovery port from the agent's config and reusing that for the tcp port.  Would be nice if we don't have to start scattering random numbers throughout the tests.

ajsutton: Done.

mbaxter: Is it possible to clean up the rest of the magic numbers?

ajsutton: I'll have a look.  They are just arbitrary numbers since we don't assert on them and don't open a port for them anywhere.

ajsutton: I've pulled them out as constants - one of the tests already had a constant it had used for this previously.

PR #1020, opened 2019-03-01
Ensure devp2p ports are written to ports file correctly

ajsutton:
## PR description
Start P2P network synchronously so the ports are guaranteed to be known before we write the ports file.
Include the P2P TCP port in ports file even when peer discovery is disabled.
Load information from the advertised peer rather than the discovery listening socket so we get the right IP instead of 0.0.0.0.
Fix admin_nodeInfo to include the ?discport param in the enode URI when the discovery port differs from the P2P port and not fail with internal server error when p2p is enabled but discovery is disabled.


ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/NetworkRunner.java
```diff
@@ -89,7 +84,7 @@ public void start() {
     if (started.compareAndSet(false, true)) {
       LOG.info("Starting Network.");
       setupHandlers();
-      networkExecutor.submit(network);
```
ajsutton: Not entirely sure why this ever had its own dedicated executor - starting the network is actually very quick as it basically just opens a port and returns.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/AdminNodeInfo.java
```diff
@@ -67,25 +66,23 @@ public JsonRpcResponse response(final JsonRpcRequest req) {
...
-          "enode://"
-              + nodeId.toString().substring(2)
-              + "@"
-              + InetAddresses.toUriString(inetAddress)
-              + ":"
-              + port);
+      final PeerInfo peerInfo = peerNetwork.getLocalPeerInfo();
+      final BytesValue nodeId = peerInfo.getNodeId();
+      final Peer advertisedPeer = peerNetwork.getAdvertisedPeer();
+      if (advertisedPeer != null) {
```
mbaxter: Should we update `getAdvertisedPeer` to return an `Optional`?

ajsutton: Done.  Good idea.

PR #1022, opened 2019-03-01
[PRIV] Get Internal logs and output

Puneetha17:
## PR description

Return the logs and output of internal transaction of a PrivateTx to the eea_PrivateTxReceipt. 

## Fixed Issue(s)

https://pegasys1.atlassian.net/browse/PRIV-23


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/precompiles/privacy/PrivacyPrecompiledContract.java
```diff
@@ -89,18 +105,40 @@ public BytesValue compute(final BytesValue input, final MessageFrame messageFram
...
               publicWorldState,
-              privateWorldState,
+              privateWorldStateUpdater,
               messageFrame.getBlockHeader(),
               privateTransaction,
               messageFrame.getMiningBeneficiary(),
               OperationTracer.NO_TRACING,
               messageFrame.getBlockHashLookup());
 
+      privateWorldStateUpdater.commit();
```
lucassaldanha: Don't we need to check the result from `privateTransactionProcessor.processTransaction()` to know if the transaction has been processed successfully?

ethereum/core/src/test-support/java/tech/pegasys/pantheon/ethereum/core/InMemoryStorageProvider.java
```diff
@@ -57,6 +59,16 @@ public WorldStateStorage createWorldStateStorage() {
     return new KeyValueStorageWorldStateStorage(new InMemoryKeyValueStorage());
   }
 
+  @Override
+  public PrivateTransactionStorage createPrivateTransactionStorage() {
+    return null;
```
lucassaldanha: Returning null here might lead to errors in the future. Wouldn't it be better to return a stub implementation of the `PrivateTransactionStorage`?

ethereum/core/src/test-support/java/tech/pegasys/pantheon/ethereum/core/InMemoryStorageProvider.java
```diff
@@ -57,6 +59,16 @@ public WorldStateStorage createWorldStateStorage() {
...
+  }
+
+  @Override
+  public PrivateStateStorage createPrivateStateStorage() {
```
lucassaldanha: Returning null here might lead to errors in the future. Wouldn't it be better to return a stub implementation of the `PrivateStateStorage`?

lucassaldanha:
Waiting for feedback on the AT failure.

PR #1025, opened 2019-03-03
Avoid port conflicts in acceptance tests

ajsutton:
## PR description
Acceptance tests were attempting to find an available port by starting a socket server with port 0, stopping it then assuming that port would remain unused which led to a lot of test failures when Pantheon failed to start up because of a port conflict.

This PR changes the behaviour so that Pantheon is always started with port 0 as the p2p port and then the port is read out of the `pantheon.ports` file. As a result, we need to start a cluster by first starting one node as the bootnode, getting it's enode URL (which requires the port) and then starting the others using that bootnode.  The peer discovery process then ensures all peers in the cluster wind up connected to each other without requiring them all to be in the bootnode list.

Note this builds on the fixes in #1021, #1020, #1019 and #1018. The changeset will be much easier to review once they've been merged.


acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/PantheonNode.java
```diff
@@ -67,14 +67,13 @@
 import org.web3j.protocol.websocket.WebSocketService;
 import org.web3j.utils.Async;
 
-public class PantheonNode implements Node, NodeConfiguration, RunnableNode, AutoCloseable {
```
CjHare: Is any class still using `Node` interface, once removed from `PantheonNode`?

ajsutton: Yes, `PantheonNode`.  `RunnableNode` extends `Node` so no point in declaring it twice (it really messes up class hierarchy views).

CjHare: 👍

PR #1028, opened 2019-03-03
[PAN-2354] Handle metrics request closing before response is generated

ajsutton:
## PR description
Log the reason a metric request failed and check if the result has been closed before writing to it to avoid printing spurious errors to the console.


metrics/src/main/java/tech/pegasys/pantheon/metrics/prometheus/MetricsHttpService.java
```diff
@@ -205,7 +205,10 @@ private void metricsRequest(final RoutingContext routingContext) {
         false,
         (res) -> {
           if (res.failed()) {
+            LOG.debug("Request for metrics failed", res.cause());
```
mbaxter: Should this be a `warn` or `error`?

ajsutton: Normally I wouldn't make it an error since the error is reported back out to the user already but actually in this case there's no user input to the request so any failures are almost certainly caused by our code being broken.  I've increased it to error.

PR #1029, opened 2019-03-04
[PAN-2352] Make alloc optional and provide nicer error messages when genesis config is invalid

ajsutton:
## PR description
Allow the `alloc` element in genesis configs to be omitted, defaulting to no allocations.
Provide better error messages when items in the genesis config are invalid.


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/chain/GenesisState.java
```diff
@@ -153,38 +153,58 @@ private static Hash calculateGenesisStateHash(final List<GenesisAccount> genesis
...
-    return Hash.wrap(Bytes32.fromHexString(genesis.getMixHash()));
-  }
-
-  private static long parseNonce(final GenesisConfigFile genesis) {
-    String nonce = genesis.getNonce().toLowerCase(Locale.US);
-    if (nonce.startsWith("0x")) {
-      nonce = nonce.substring(2);
-    }
-    return Long.parseUnsignedLong(nonce, 16);
+    return withNiceErrorMessage("mixHash", genesis.getMixHash(), Hash::fromHexStringLenient);
```
mbaxter: I assume you meant to make the mixHash parsing more lenient here^ ?  Wonder if we should accept any zero-valued hash, and otherwise be strict ...

ajsutton: Yes this was deliberate (as with parentHash).  It's unlikely that it will make any difference but since the mix hash for the genesis file really doesn't matter (as long as it's the same for all nodes) it seems nice to not have to add all the digits for no real reason.  It doesn't seem worth the extra code to specifically allow it only for 0x0.

PR #1036, opened 2019-03-04
[minor] add rpc_modules json-rpc

shemnon:
## PR description

rpc_module enumerates the modules exposed to RPC.  This allows the
`geth attach` geth javascript console to attach to Pantheon and use the
json-rpc APIs that pantheon and geth share.

### DIFF
--- a/ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/RpcModules.java
+++ b/ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/RpcModules.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2018 ConsenSys AG.
+ *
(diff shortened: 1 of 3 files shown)


ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/RpcModules.java
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  @Override
+  public String getName() {
+    return "rpc_modules";
+  }
+
+  @Override
+  public JsonRpcResponse response(final JsonRpcRequest req) {
+    // For now, just return an empty list.
```
mbaxter: Looks like you can cut this comment.

shemnon: Done

PR #1039, opened 2019-03-04
[PAN-2342] Update discovery logic to trust bootnodes only when out of sync

lucassaldanha:
When using smart contract based permissioning, the node will trust only its bootnodes until it reaches a sync'd state for the first time. Peer discovery will be disabled until we reach a sync'd state.

Summary:
* Created *SyncStatusNodePermissioningProvider* class that subscribes to the Synchronizer SyncStatus updates to check if the node has ever reached a sync'd state.
* Added a temporary implementation of NodePermissioningController that may or may not have a SyncStatusNodePermissioningProvider. For now, the default behaviour is to allow anything and never delay peer discovery.
* Updated PeerDiscoveryController with the NodePermissioningController temporary implementation. It won't change the current behaviour of discovery.


ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
```diff
@@ -189,7 +193,12 @@ public void start() {
         bootstrapNodes.stream()
             .filter(this::whitelistIfPresentIsNodePermitted)
             .collect(Collectors.toList());
-    recursivePeerRefreshState.start(initialDiscoveryPeers, localPeer.getId());
+    if (nodePermissioningController.delayPeerDiscovery()) {
+      nodePermissioningController.peerDiscoveryCallback(
```
lucassaldanha: When we dealy peer discovery, we should set the callback that will be triggered after we reach the sync status. In this case, the callback will be enabling peer discovery through `recursivePeerRefreshState.start`.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
```diff
@@ -135,6 +136,9 @@
 
   private RecursivePeerRefreshState recursivePeerRefreshState;
 
+  private final NodePermissioningController nodePermissioningController =
```
lucassaldanha: In this implementation, we aren't using the NodePermissioningController for any permissioning.

ethereum/permissioning/src/main/java/tech/pegasys/pantheon/ethereum/permissioning/node/NodePermissioningController.java
```diff
@@ -0,0 +1,61 @@
...
+
+  public boolean isPermitted(final EnodeURL sourceEnode, final EnodeURL destinationEnode) {
+    LOG.trace("Checking node permission: {} -> {}", sourceEnode, destinationEnode);
+
+    return syncStatusNodePermissioningProvider
+        .map((provider) -> provider.isPermitted(sourceEnode, destinationEnode))
+        .orElse(true);
+  }
+
+  public boolean delayPeerDiscovery() {
```
Errorific: squeeze this and peerDiscoveryCallback together into 1 function that either calls the runnable immediately or delays it based on whats necessary.

Also be aware of threading issues between setting the discovery callback and onReachingSync getting called by the observable thread.

Errorific: And probably empty the callback optional before running it so you get a latch that stops anything from racing to call it twice

ethereum/permissioning/src/main/java/tech/pegasys/pantheon/ethereum/permissioning/node/NodePermissioningProvider.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package tech.pegasys.pantheon.ethereum.permissioning.node;
+
+import tech.pegasys.pantheon.util.enode.EnodeURL;
+
+public interface NodePermissioningProvider {
```
Errorific: \@FunctionalInterface just cause we can

ethereum/permissioning/src/test/java/tech/pegasys/pantheon/ethereum/permissioning/node/NodePermissioningControllerTest.java
```diff
@@ -0,0 +1,104 @@
...
+  @Test
+  public void shouldExecutePeerDiscoveryCallbackWhenHasReachedSyncCallbackIsTriggered() {
+    final Runnable peerDiscoveryCallback = mock(Runnable.class);
+    when(syncStatusNodePermissioningProvider.hasReachedSync()).thenReturn(false);
+
+    controller.startPeerDiscoveryCallback(peerDiscoveryCallback);
+    hasReachedSyncCallback.run();
+
+    verify(peerDiscoveryCallback).run();
+    // should set the callback to empty after executing it once
```
macfarla: this test has the same setup as the previous one `ifSyncHasNotBeenReachedShouldSetPeerDiscoveryCallback` but the outcome is different?

lucassaldanha: The outcome is different 'cause after running the callback we set it to empty. In the previous test we should not run the callback, just check if it has been set. I'll fix it! It is probably a CopyAndPasteException! ;)

macfarla: 👍

ethereum/permissioning/src/main/java/tech/pegasys/pantheon/ethereum/permissioning/node/provider/SyncStatusNodePermissioningProvider.java
```diff
@@ -0,0 +1,82 @@
...
+    this.synchronizer = synchronizer;
+    this.syncStatusObserverId = this.synchronizer.observeSyncStatus(this::handleSyncStatusUpdate);
+    this.bootnodes.addAll(bootnodes);
+  }
+
+  private void handleSyncStatusUpdate(final SyncStatus syncStatus) {
+    if (syncStatus != null && !hasReachedSync) {
+      long blocksBehind = syncStatus.getHighestBlock() - syncStatus.getCurrentBlock();
+      if (blocksBehind <= 0) {
+        synchronized (this) {
```
Errorific: this synchronized needs to start before L43 where hasReachedSync is checked

ethereum/permissioning/src/main/java/tech/pegasys/pantheon/ethereum/permissioning/node/provider/SyncStatusNodePermissioningProvider.java
```diff
@@ -0,0 +1,82 @@
...
+      }
+    }
+  }
+
+  public boolean hasReachedSync() {
+    return hasReachedSync;
+  }
+
+  public void setHasReachedSyncCallback(final Runnable runnable) {
+    this.hasReachedSyncCallback = Optional.of(runnable);
```
Errorific: Feels like this needs to be mindful of hasReachedSync and not take the runnable if we've already reached sync or execute it at the point or something. 
Maybe as per our discussions on `NodePermissioningController` the `hasReachedSync` could be an `AtomicBoolean` and we use that to synchronise whether we're still waiting or whether we are still waiting? mild problem there where isPermitted can be blocked behind the callback getting completed. 
Maybe treat `hasReachedSyncCallback` same way as `NodePermissioningController.peerDiscoveryCallback` to prevent that runnable from getting run twice? 
Or maybe this ones just better off with more `synchronized`.

PR #1043, opened 2019-03-05
773 Add PendingTransactions JSON-RPC

### LINKED ISSUE #773: Add JSON RPC endpoint to look up the pending transactions
Create a JSON RPC endpoint to query the transaction pool, looking up the list of pending transactions.

The initial version will form the basis for some more advanced querying support.

**method name:** pantheon_txpool_pendingTransactions

**Inputs:** none

**Outputs:** Return an array of transaction info objects, where a transaction info is based off of the private TransactionInfo class in PendingTransactions.

the transaction info should specifically expose:
receivedFromLocalSource and the underlying transaction.

It should specifically not expose the internal sequence number.

This issue/feature is the basis for some fuller query functionality.

In the future it is expected that the transaction info would be expanded to include more information, and that richer query function would be added.

EdwinLeeGreene:
## Fixed Issue(s)
fixes #773


ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/JsonRpcMethodsFactory.java
```diff
@@ -248,7 +249,8 @@
           new MinerStart(miningCoordinator),
           new MinerStop(miningCoordinator),
           minerSetCoinbase,
-          new MinerSetEtherbase(minerSetCoinbase));
+          new MinerSetEtherbase(minerSetCoinbase),
+          new TxPoolPendingTransactions(transactionPool.getPendingTransactions()));
```
ewlg: Not sure if this should be in the mining section; placed it there for now.

EdwinLeeGreene: Not sure if this is the right place for this. I placed it in the mining section for now.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/TxPoolPendingTransactions.java
```diff
@@ -0,0 +1,42 @@
...
+    return "pantheon_txpool_pendingTransactions";
+  }
+
+  @Override
+  public JsonRpcResponse response(final JsonRpcRequest request) {
+    final JsonRpcSuccessResponse jsonRpcSuccessResponse =
+        new JsonRpcSuccessResponse(
+            request.getId(),
+            new PendingTransactionsResult(pendingTransactions.getTransactionInfo()));
+    return jsonRpcSuccessResponse;
```
macfarla: simpler to just return new JsonRpcSuccessResponse(...)

PR #1047, opened 2019-03-05
Fix ImportBlocksTask to only request from peers that claim to have the blocks

ajsutton:
## PR description
ImportBlocksTask was requesting data from any idle peer without checking that the peer had a chain height above the starting block it was requesting.  As a result, peers earlier on the chain would respond with an empty response and risk being disconnected with USELESS_PEER.

Also includes the type of request we made when recording a useless response to make debugging easier.


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/tasks/ImportBlocksTask.java
```diff
@@ -96,6 +97,11 @@ protected void executeTaskWithPeer(final EthPeer peer) throws PeerNotConnected {
             });
   }
 
+  @Override
+  protected Optional<EthPeer> findSuitablePeer() {
```
mbaxter: Wonder if its worth it to require subclasses to implement this...

ajsutton: Interesting.  Will take a look at that and follow up with a PR if it works out.

PR #1048, opened 2019-03-05
Track added at in txpool

rojotek:
Store the instant that a pending transaction is added to the txpool, and expose this out through the api.


consensus/clique/src/test/java/tech/pegasys/pantheon/consensus/clique/blockcreation/CliqueBlockCreatorTest.java
```diff
@@ -109,7 +110,7 @@ public void proposerAddressCanBeExtractFromAConstructedBlock() {
         new CliqueBlockCreator(
             coinbase,
             parent -> extraData.encode(),
-            new PendingTransactions(5),
+            new PendingTransactions(5, Clock.systemUTC()),
```
ajsutton: Could we use the fixed clock in tests please?  Real-time in tests comes back to bite you.  Might need a util method somewhere to make it easy since `Clock.fixed` requires specifying a time and we generally don't care what time is used.

pantheon/src/main/java/tech/pegasys/pantheon/controller/MainnetPantheonController.java
```diff
@@ -147,7 +147,7 @@ private MainnetPantheonController(
 
     final TransactionPool transactionPool =
         TransactionPoolFactory.createTransactionPool(
-            protocolSchedule, protocolContext, ethProtocolManager.ethContext());
+            protocolSchedule, protocolContext, ethProtocolManager.ethContext(), Clock.systemUTC());
```
ajsutton: Would be good if we could only create one clock instance in the controller (there's a second one being passed to `DefaultBlockScheduler`.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/results/TransactionInfoResult.java
```diff
@@ -22,17 +24,24 @@
...
   }
 
   @JsonGetter(value = "hash")
   public String getHash() {
     return hash;
   }
 
+  @JsonGetter(value = "addedToPoolAt")
+  public long getAddedToPoolAt() {
+    return addedToPoolAt.toEpochMilli();
```
ajsutton: As discussed this should be ISO8601 format rather than epoch millis.

consensus/clique/src/test/java/tech/pegasys/pantheon/consensus/clique/blockcreation/CliqueBlockCreatorTest.java
rojotek: Added a TestClock with a fixed() static method.

pantheon/src/main/java/tech/pegasys/pantheon/controller/MainnetPantheonController.java
rojotek: done -- we now have a single call to `Clock.systemUTC()` in the codebase.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/results/TransactionInfoResult.java
rojotek: cool -- have used toString on the instant which returns an ISO8601 string.

PR #1049, opened 2019-03-05
Remove DB init need to use public key subcommand

NicolasMassart:
## PR description
Remove the need to initialise the database when running public key subcommands.

A functional interface is introduced for the keypar loader from file to enable testing of the subcommand by injecting a predefined keypair. The injection takes place in the PantheonCommand class as a method that can be overridden in the test extension of the class.
Unit tests are updated.

## Fixed Issue(s)

fixes PAN-2259


pantheon/src/main/java/tech/pegasys/pantheon/cli/PublicKeySubCommand.java
```diff
@@ -59,16 +60,32 @@
...
+    this.keyLoader = keyLoader;
   }
 
   @Override
   public void run() {
     spec.commandLine().usage(out);
   }
 
+  @FunctionalInterface
+  public interface KeyLoader {
```
mbaxter: (style) Usually internal classes / interfaces go at the bottom of the class

PR #1067, opened 2019-03-07
Fix race condition in RunnerTest

usmansaleem:
## PR description
While attempting to build pantheon on a Windows machine, the `RunnerTest.fastSyncFromGenesis` was constantly failing at ` final Peer advertisedPeer = runnerAhead.getAdvertisedPeer().get();` because Runner isn't initialized completely. It is fixed by using same logic as `tech.pegasys.pantheon.tests.acceptance.dsl.node.PantheonNodeRunner#waitForPortsFile`
## Fixed Issue(s)
Wait for Runner to start before attempt to access advertise peer.


ajsutton:
Thanks for reporting this and the contributed fix. While this fix would indeed work, we've recently changed the network startup to be more synchronous to try and avoid this kind of problem but didn't take the final step of exposing a synchronous call to start `Runner` without waiting for it to exit.  While more invasive I believe it can prevent us needing to think about waiting for ports to open in embedded scenarios like this.

Would you be able to try the changes in https://github.com/ajsutton/pantheon/tree/runner-start and let me know if that fixes the issue for you please?

usmansaleem:
@ajsutton Thank you for the feedback.

Following would indeed do the trick, this was the alternate solution I had, but then I found existing code for waiting for "port files" so I copied that. I'll try it your branch today and then close my pull request.

```
Awaitility.await()
          .atMost(20, TimeUnit.SECONDS)
          .ignoreExceptions()
          .untilAsserted(() -> assertThat(runnerAhead.getAdvertisedPeer()).isNotEmpty());
```

ajsutton:
It shouldn't even need that extra wait now because `start()` won't return until the ports are all open but I messed up and didn't remove it from my first attempt at investigating. I've removed it on my branch now so would be great to check it on your setup since we know it fails there (and it passes quite consistently for me and in our CI).

usmansaleem:
@ajsutton your branch (https://github.com/ajsutton/pantheon/tree/runner-start) build successfully in my setup.

ajsutton:
Awesome thank you. I'll get a PR up and hopefully get that merged soon.

usmansaleem:
Closing as changes are/will be fixed from another contribution.

PR #1069, opened 2019-03-07
workaround for 404 error pages assets and sitemap links

NicolasMassart:
## PR description
This is a temporary fix for 404 and sitemap urls until we have the ability to match readthedocs real urls.

**Preview the 404 at https://pantheontest.readthedocs.io/en/fix-pan-2311_404_links/notexistingpage
and the sitemap at https://pantheontest.readthedocs.io/en/fix-pan-2311_404_links/sitemap.xml**

Some work is in progress about that on the RTD site for Google Summer of Code.
See https://docs.readthedocs.io/en/latest/gsoc.html#better-mkdocs-integration
This was discussed on RTD Gitter and I'm going to add some more details on the requirements for these changes on RTD page with a PR.

Once MkDocs integration will be better in RTD we'll be able to revert these hacks.
The hack is to make 404 error pages and sitemap.xml urls point to the /en/stable/ version instead of having a link relative to root. The mkdocs.yml config file site_url value was set to https://docs.pantheon.pegasys.tech/en/stable/ and some changes were made on 404 templates.

Also side changes : updated footer to remove hardcoded values and updated minimal version of mkdocs-material (we always use the latest anyway but it prevent using an old one)
## Fixed Issue(s)

fixes PAN-2311
fixes PAN-2384


docs/custom_theme/404.html
```diff
@@ -0,0 +1,18 @@
+{% extends "main.html" %}
+
+<!-- Content block -->
+{% block content %}
+
+<p style="text-align: center;font-size: 3em"><i class="md-icon">account_balance</i></p>
+<h1 id="404-page-not-found" style="text-align: center">404</h1>
+<p style="text-align: center"><strong>Deity not found in the Pantheon</strong></p>
+<p style="text-align: center">
```
MadelineMurray: _Deity not found in the Pantheon_ doesn't really make sense to me? I think it relies on the user knowing both whay deity and Pantheon mean and then making the leap that we're using Pantheon with a different meaning to our product. Maybe I'm too literal but would _Page not found in the Pantheon documentation_ be clearer?

docs/custom_theme/404.html
```diff
@@ -0,0 +1,18 @@
...
+  Try the <a href="/">homepage</a>, or use the search field on the top of this page.
+</p>
+<p style="text-align: center">
+  If you think we made a mistake and deleted a page that should be here, then please tell us on
+  <a href="{{config.extra.support.gitter}}">{{config.site_name}} Gitter channel</a> or create an issue in
```
MadelineMurray: We're about to move the issues to Jira and remove the ability to create GitHub issues so it might be worth just directing this to the GitHub issues tab (they'll be a redirect on there to Jira).  Otherwise, this will need updating again in a week.

docs/custom_theme/404.html
NicolasMassart: As discussed, we can afford some humour here.

PR #1071, opened 2019-03-07
[PAN-2348] Changes in chain head should trigger new permissioning check for active peers

lucassaldanha:
## PR description
Every time that the chain head changes, we must check if all active peers are still permitted.
* NettyP2PNetwork is listening to BlockAdded events
* For each active peer connection, check with NodePermissioningController if the peer is still permitted
* If the peer is not permitted, disconnect


ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
```diff
@@ -231,6 +282,30 @@ private boolean whitelistIfPresentIsNodePermitted(final DiscoveryPeer sender) {
...
+
+  private boolean isPeerPermitted(final Peer peer) {
+    final EnodeURL localPeerEnodeURL = new EnodeURL(localPeer.getEnodeURI());
+    final EnodeURL remotePeerEnodeURL = new EnodeURL(peer.getEnodeURI());
+
+    return nodePermissioningController
+        .map(
+            c ->
+                c.isPermitted(localPeerEnodeURL, remotePeerEnodeURL)
+                    || c.isPermitted(remotePeerEnodeURL, localPeerEnodeURL))
```
lucassaldanha: Do we want to say that the peer is permitted only when communication in both ways is permitted?

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/netty/NettyP2PNetwork.java
```diff
@@ -497,6 +546,42 @@ public void start() {
...
+                peerConnection.disconnect(DisconnectReason.REQUESTED);
+              }
+            });
+  }
+
+  private boolean isPeerConnectionAllowed(final PeerConnection peerConnection) {
+    LOG.trace(
+        "Checking if connection with peer {} is permitted", peerConnection.getPeer().getNodeId());
+
+    final String localNodeId = ourPeerInfo.getNodeId().toString().substring(2);
```
Errorific: refactor the local & remote enode bits to a seperate function like `peerToEnode(PeerInfo peer, InetSocketAddress address)`

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/netty/NettyP2PNetwork.java
```diff
@@ -497,6 +546,42 @@ public void start() {
...
+    final String remoteNodeId = peerConnection.getPeer().getNodeId().toString().substring(2);
+    final InetSocketAddress remoteAddress = (InetSocketAddress) peerConnection.getRemoteAddress();
+    final String remoteHostAddress = remoteAddress.getHostString();
+    final int remotePort = remoteAddress.getPort();
+    final EnodeURL remotePeerEnodeURL = new EnodeURL(remoteNodeId, remoteHostAddress, remotePort);
+
+    return nodePermissioningController
+        .map(
+            c ->
+                c.isPermitted(localPeerEnodeURL, remotePeerEnodeURL)
```
Errorific: Something feels wrong about this check. Don't think it should be `||`. Wondering if the directional part of the rules makes sense or not, or whether we need to track the direction a connection was established in?

PR #1075, opened 2019-03-08
[PAN-2345] node permissioning controller

macfarla:
## PR description
Refactor NodeWhitelistController to NodeLocalConfigPermissioningController
Refactored NodePermissioningController to accept multiple providers and check all of them
Added to LocalConfigNodePermissioningController the notion of its own enode URL. The node will not add its own enode to the whitelist. 

## Fixed Issue(s)
fixes #PAN-2345


ethereum/permissioning/src/main/java/tech/pegasys/pantheon/ethereum/permissioning/node/provider/LocalConfigNodePermissioningProvider.java
```diff
@@ -0,0 +1,41 @@
...
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package tech.pegasys.pantheon.ethereum.permissioning.node.provider;
+
+import tech.pegasys.pantheon.ethereum.permissioning.NodeLocalConfigPermissioningController;
+import tech.pegasys.pantheon.ethereum.permissioning.node.NodePermissioningProvider;
+import tech.pegasys.pantheon.util.enode.EnodeURL;
+
+public class LocalConfigNodePermissioningProvider implements NodePermissioningProvider {
```
lucassaldanha: Do we need this class? Can't we make `NodeLocalConfigPermissioningController` implement `NodePermissioningProvider` and implement `isPermitted(final EnodeURL sourceEnode, final EnodeURL destinationEnode)` there?

ethereum/permissioning/src/main/java/tech/pegasys/pantheon/ethereum/permissioning/node/NodePermissioningController.java
```diff
@@ -25,22 +25,28 @@
...
     LOG.trace("Checking node permission: {} -> {}", sourceEnode, destinationEnode);
 
-    return syncStatusNodePermissioningProvider
-        .map((provider) -> provider.isPermitted(sourceEnode, destinationEnode))
-        .orElse(true);
+    if (syncStatusNodePermissioningProvider.isPresent()
+        && !syncStatusNodePermissioningProvider.get().isPermitted(sourceEnode, destinationEnode)) {
+      return false;
+    } else {
+      for (int i = 0; i < providers.length; i++)
```
lucassaldanha: If the array has the wrong order of providers (smart contract first, and then local config), we might end up with the wrong logic. Either we force the user to specify the providers by type and ensure we set them in the proper order, or we need to check that they are in the proper order.

Also, we need to think about the case when only one of the two types of permissioning is being used. E.g. only local config vs. only smart contract permissioning.

ethereum/permissioning/src/main/java/tech/pegasys/pantheon/ethereum/permissioning/node/NodePermissioningController.java
```diff
@@ -25,22 +25,28 @@
   private static final Logger LOG = LogManager.getLogger();
 
   private final Optional<SyncStatusNodePermissioningProvider> syncStatusNodePermissioningProvider;
+  NodePermissioningProvider[] providers;
 
   public NodePermissioningController(
```
lucassaldanha: The `NodePermissiningController` should also be able to read the permissioning config flags and use the `NodePermissioningProviders` accordingly.

It feels like we could use a factory that from the PermissioningConfiguration gives you an instance of `NodePermissioningController`.

ethereum/permissioning/src/main/java/tech/pegasys/pantheon/ethereum/permissioning/node/NodePermissioningController.java
Errorific: Order doesn't matter from current thinking. In discussions (if memory serves) we decided we wanted to make sure we pass all permissioning filters to allow a connection rather than have one override the other, allowing the intersection of the 2 permissioning lists. The use case this currently misses is if a node is banned on the smart contract but allowed on the local contract which we discussed not doing for the moment as it allows the addition of nodes the network agreed shouldn't be there and we would wait for a real user to ask for it then consider it again.

ethereum/permissioning/src/main/java/tech/pegasys/pantheon/ethereum/permissioning/node/provider/LocalConfigNodePermissioningProvider.java
macfarla: done

ethereum/permissioning/src/main/java/tech/pegasys/pantheon/ethereum/permissioning/node/NodePermissioningController.java
macfarla: there is a certain dependency on the CLI options for smartContractAddress and nodeEnabled/accountEnabled

macfarla: I've added a factory - but not sure how to test it without exposing the providers[] directly

macfarla: also added a separate SmartContractPermissioningConfiguration object - made more sense than adding to PermissioningConfiguration which encapsulates the TOML config (and gets reloaded from file) and I think PermissioningConfiguration should be refactored to LocalPermissioningConfiguration but wanted to validate this approach (two separate configs) first

lucassaldanha: I think we should have one PermissioningConfiguration object. From what I remember, the PermissioningConfigurationBuilder is the one that does all the TOML magic.

macfarla: ok I made one PermissioningConfiguration class that has two optionals - LocalPermissioningConfiguration and SmartContractPermissioningConfiguratoin

macfarla: sorry about all the refactoring...

ethereum/permissioning/src/main/java/tech/pegasys/pantheon/ethereum/permissioning/NodePermissioningControllerFactory.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.List;
+
+public class NodePermissioningControllerFactory {
+
+  public NodePermissioningController create(
+      final PermissioningConfiguration permissioningConfiguration,
+      final Synchronizer synchronizer,
+      final Collection<EnodeURL> bootnodes,
+      final TransactionSimulator transactionSimulator) {
+    final SyncStatusNodePermissioningProvider syncStatusProvider =
```
lucassaldanha: We should only use `SyncStatusNodePermissioningProvider` if the smart contract based permissioning is enabled.

ethereum/permissioning/src/test/java/tech/pegasys/pantheon/ethereum/permissioning/node/NodePermissioningControllerFactoryTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+    List<NodePermissioningProvider> providers = controller.getProviders();
+    assertThat(providers.size()).isEqualTo(1);
+
+    NodePermissioningProvider p1 = providers.get(0);
+    assertThat(p1).isNotNull();
+  }
+
+  @Test
+  public void testCreateWithLocalNodePermissioningEnabledOnly() {
```
lucassaldanha: We need to validate that when we are using local config only, the `SyncStatusNodePermissioningProvider` is not present.

ethereum/permissioning/src/test/java/tech/pegasys/pantheon/ethereum/permissioning/node/NodePermissioningControllerFactoryTest.java
```diff
@@ -0,0 +1,100 @@
...
+    NodePermissioningControllerFactory factory = new NodePermissioningControllerFactory();
+    NodePermissioningController controller =
+        factory.create(config, synchronizer, bootnodes, transactionSimulator);
+
+    List<NodePermissioningProvider> providers = controller.getProviders();
+    assertThat(providers.size()).isEqualTo(0);
+  }
+
+  @Test
+  public void testCreateWithSmartContractNodePermissioningEnabledOnly() {
```
lucassaldanha: We need to validate that when we are using smart contract perm, the `SyncStatusNodePermissioningProvider` is present.

ethereum/permissioning/src/test/java/tech/pegasys/pantheon/ethereum/permissioning/node/NodePermissioningControllerFactoryTest.java
```diff
@@ -0,0 +1,100 @@
...
+    NodePermissioningController controller =
+        factory.create(config, synchronizer, bootnodes, transactionSimulator);
+
+    List<NodePermissioningProvider> providers = controller.getProviders();
+    assertThat(providers.size()).isEqualTo(1);
+
+    NodePermissioningProvider p1 = providers.get(0);
+    assertThat(p1).isNotNull();
+  }
+}
```
lucassaldanha: I believe we could have a test that will validate the behaviour when we have both permissioning sources enabled.

ethereum/permissioning/src/test/java/tech/pegasys/pantheon/ethereum/permissioning/node/NodePermissioningControllerFactoryTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+    NodePermissioningControllerFactory factory = new NodePermissioningControllerFactory();
+    NodePermissioningController controller =
+        factory.create(config, synchronizer, bootnodes, transactionSimulator);
+
+    List<NodePermissioningProvider> providers = controller.getProviders();
+    assertThat(providers.size()).isEqualTo(1);
+
+    NodePermissioningProvider p1 = providers.get(0);
+    assertThat(p1).isNotNull();
```
lucassaldanha: Instead of checking that it is not null, we should probably check that we have the right provider type:
```
assertThat(p1).isInstanceOf(NodeLocalConfigPermissioningController.class);
```

ethereum/permissioning/src/test/java/tech/pegasys/pantheon/ethereum/permissioning/node/NodePermissioningControllerFactoryTest.java
```diff
@@ -0,0 +1,100 @@
...
```
lucassaldanha: Instead of checking that it is not null, we should probably check that we have the right type:
```
assertThat(p1).isInstanceOf(NodeSmartContractPermissioningController.class);
```

ethereum/permissioning/src/test/java/tech/pegasys/pantheon/ethereum/permissioning/node/NodePermissioningControllerTest.java
```diff
@@ -57,4 +67,62 @@ public void peerDiscoveryCallbackShouldBeDelegatedToSyncStatusNodePermissioningP
 
     verify(syncStatusNodePermissioningProvider).setHasReachedSyncCallback(any(Runnable.class));
   }
+
+  @Test
+  public void whenInSyncWeShouldDelegateToLocalConfigNodePermissioningProvider() {
```
lucassaldanha: This test is wrong. When using the local config permissioning only, the SyncStatusPermissioningProvider should not be used.

lucassaldanha:
For the `NodePermissioningControllerTest` test, we need to ensure we have the different possible scenarios covered:
1. When using only local config permissioning
    - Should NOT check the `SyncStatusPermissioningProvider`
    - Should only delegate to the `NodeLocalConfigPermissioningController` (only one provider will be enabled)
2. When using SmartContract based permissioning (either standalone or in conjuction with local config)
    - If `SyncStatusPermissioningProvider.isPermitted()` returns false, communication is not permitted
    - If `SyncStatusPermissioningProvider.isPermitted()` returns true:
        - communication is permitted if every other provider returns true
        - communication is forbidden if any provider returns false

The current name of the test methods imply the the `SyncStatusPermissioningProvider` is always used. This shouldn't happen.

ethereum/permissioning/src/test/java/tech/pegasys/pantheon/ethereum/permissioning/node/NodePermissioningControllerTest.java
macfarla: renamed and corrected the test - check behaviour when no SyncStatusProvider

ethereum/permissioning/src/test/java/tech/pegasys/pantheon/ethereum/permissioning/node/NodePermissioningControllerFactoryTest.java
macfarla: done

ethereum/permissioning/src/test/java/tech/pegasys/pantheon/ethereum/permissioning/node/NodePermissioningControllerFactoryTest.java
macfarla: done

ethereum/permissioning/src/test/java/tech/pegasys/pantheon/ethereum/permissioning/node/NodePermissioningControllerFactoryTest.java
macfarla: done

ethereum/permissioning/src/test/java/tech/pegasys/pantheon/ethereum/permissioning/node/NodePermissioningControllerFactoryTest.java
macfarla: done

ethereum/permissioning/src/test/java/tech/pegasys/pantheon/ethereum/permissioning/node/NodePermissioningControllerFactoryTest.java
macfarla: done

ethereum/permissioning/src/main/java/tech/pegasys/pantheon/ethereum/permissioning/NodePermissioningControllerFactory.java
macfarla: done

ethereum/permissioning/src/main/java/tech/pegasys/pantheon/ethereum/permissioning/NodeLocalConfigPermissioningController.java
```diff
@@ -292,4 +294,9 @@ public WhitelistOperationResult result() {
       return message;
     }
   }
+
+  @Override
+  public boolean isPermitted(final EnodeURL sourceEnode, final EnodeURL destinationEnode) {
+    return isPermitted(sourceEnode) && isPermitted(destinationEnode);
```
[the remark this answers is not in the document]
macfarla: changed this and added selfEnode to this class

ethereum/permissioning/src/test/java/tech/pegasys/pantheon/ethereum/permissioning/node/NodePermissioningControllerFactoryTest.java
```diff
@@ -0,0 +1,138 @@
...
+    NodePermissioningControllerFactory factory = new NodePermissioningControllerFactory();
+    NodePermissioningController controller =
+        factory.create(config, synchronizer, bootnodes, selfEnode, transactionSimulator);
+
+    List<NodePermissioningProvider> providers = controller.getProviders();
+    assertThat(providers.size()).isEqualTo(1);
+
+    NodePermissioningProvider p1 = providers.get(0);
+    assertThat(p1).isInstanceOf(SmartContractPermissioningController.class);
+    assertThat(controller.getSyncStatusNodePermissioningProvider().isPresent()).isTrue();
```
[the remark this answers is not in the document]
macfarla: done

ethereum/permissioning/src/test/java/tech/pegasys/pantheon/ethereum/permissioning/node/NodePermissioningControllerFactoryTest.java
```diff
@@ -0,0 +1,138 @@
...
+    NodePermissioningControllerFactory factory = new NodePermissioningControllerFactory();
+    NodePermissioningController controller =
+        factory.create(config, synchronizer, bootnodes, selfEnode, transactionSimulator);
+
+    List<NodePermissioningProvider> providers = controller.getProviders();
+    assertThat(providers.size()).isEqualTo(1);
+
+    NodePermissioningProvider p1 = providers.get(0);
+    assertThat(p1).isInstanceOf(NodeLocalConfigPermissioningController.class);
+    assertThat(controller.getSyncStatusNodePermissioningProvider().isPresent()).isFalse();
```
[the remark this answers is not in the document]
macfarla: done

PR #1076, opened 2019-03-08
Added section on using Geth attach with Pantheon 

MadelineMurray:
Added section on using Geth attach with Pantheon


docs/JSON-RPC-API/Using-JSON-RPC-API.md
```diff
@@ -25,14 +13,40 @@ Set the host to `0.0.0.0` to allow remote connections.
...
 
+## Postman
+
+Use the button to import our collection of examples to [Postman](https://www.getpostman.com/). 
+
+[![Run in Postman](https://run.pstmn.io/button.svg)](https://app.getpostman.com/run-collection/cffe1bc034b3ab139fa7)
+
+## Geth Console 
+
+To use the geth console with Pantheon:
```
NicolasMassart: May we explain why we want to use Geth console ?
Possible answers (on or more):
- we don't provide one in Pantheon as we think Geth one is working well
- it enables you to use JS directly in the console to create small script around calling the rpc method
- it makes Pantheon able to be managed with non interactive JS scripts.
- ...

docs/Reference/JSON-RPC-API-Methods.md
```diff
@@ -2482,4 +2482,39 @@ None
         "id": 1,
         "result": "Success"
     }
-    ```
\ No newline at end of file
+    ```
+    
+## Miscellaneous Methods 
+
+### rpc_modules
```
NicolasMassart: Perhaps explain the use of this method a bit more like it's used by Geth Console to know what it can call. SOme other projects have the same question about what it's used for : https://github.com/ethereum/web3.py/issues/254

docs/Reference/JSON-RPC-API-Methods.md
```diff
@@ -79,7 +79,7 @@ Properties of the node object are:
 
 !!! example
     ```bash tab="curl HTTP request"
-    $ curl -X POST --data '{"jsonrpc":"2.0","method":"admin_nodeInfo","params":[],"id":1}' <JSON-RPC-http-endpoint:port>
+    curl -X POST --data '{"jsonrpc":"2.0","method":"admin_nodeInfo","params":[],"id":1}' <JSON-RPC-http-endpoint:port>
```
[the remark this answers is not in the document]
MadelineMurray: I've been removing the when I see them.  My thought being that unless we have a 'copy command' button it makes it harder to copy and paste examples. Happy to discuss if you disagree though.

docs/Reference/JSON-RPC-API-Methods.md
MadelineMurray: Per Danno the fact this is used by geth console has no bearing on it's implementation so he thought we should leave that out.  It's unlikely to be called by anyone but after discussion with the POs I'm going to leave it in the doc for completeness.

PR #1078, opened 2019-03-08
Added tx-pool-max-size command line parameter

rojotek:
Added tx-pool-max-size command line parameter to config the max size of the tx pool.


pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -458,6 +459,14 @@ protected KeyLoader getKeyLoader() {
           "The address to which the privacy pre-compiled contract will be mapped to (default: ${DEFAULT-VALUE})")
   private final Integer privacyPrecompiledAddress = Address.PRIVACY;
 
+  @Option(
+      names = {"--tx-pool-max-size"},
+      paramLabel = MANDATORY_INTEGER_FORMAT_HELP,
+      description =
+          "Maximum number of pending transactions that will be kept in the transaction pool (default: ${DEFAULT-VALUE})",
+      arity = "1")
+  private final Integer txPoolMaxSize = PendingTransactions.MAX_PENDING_TRANSACTIONS;
```
NicolasMassart: Perhaps you could move the constant to DefaultCommandValues.MAX_PENDING_TRANSACTIONS where all the other default values are. Also PendingTransactions.MAX_PENDING_TRANSACTIONS was used only in the TransactionPoolFactory so it won't brake anything.

lucassaldanha: I'd argue the opposite. I'd rather have the default values where they belong. If it is a default for the number of pending transactions, why it wouldn't be in the `PendingTransactions` class?

PR #1080, opened 2019-03-11
Lowercase addresses in dev.json so they're compatible with MetaMask

ajsutton:
## PR description
Recent versions of MetaMask reject addresses that use uppercase letters, so ensure the letters in our dev.json are all lowercase to make it easy to copy them and use in MetaMask.


ajsutton:
Closing. The confusion comes from a MetaMask bug but these addresses appear to be valid according to EIP55 so may as well keep them that way.

PR #1081, opened 2019-03-11
Added IBFT 1.0

MadelineMurray:
Added IBFT 1.0


docs/Consensus-Protocols/QuorumIBFT.md
```diff
@@ -0,0 +1,10 @@
+description: Pantheon consensus protocols
+<!--- END of page meta data -->
+
+# IBFT 1.0 
+
+Pantheon nodes can participate (submit transactions and receive blocks) in a Quorum [IBFT 1.0](https://github.com/ethereum/EIPs/issues/650) 
+network but cannot be a validator. 
+
+To connect to a Quorum IBFT 1.0 network, use the [`--genesis-file`](../Reference/Pantheon-CLI-Syntax.md#genesis-file)
```
rain-on: its worth noting that Quorum uses "Istanbul", whereas we use "ibft" - thus there needs to be a manual modification of Quorum's genesis file in order for us to connect (this is kinda dumb, and maybe we should spend 2 Story Points just to make this the same).
Also - I am not sure if our config params line up with Quorums, so that may require further manual mods.

mkdocs.yml
```diff
@@ -60,6 +60,7 @@ nav:
         - Comparing PoA Consensus Protocols: Consensus-Protocols/Comparing-PoA.md
         - Clique: Consensus-Protocols/Clique.md
         - IBFT 2.0: Consensus-Protocols/IBFT.md
+        - IBFT 1.0: Consensus-Protocols/QuorumIBFT.md
```
NicolasMassart: Perhaps showing Quorum in the page title for nav.
```suggestion
        - Quorum IBFT 1.0: Consensus-Protocols/QuorumIBFT.md
```

docs/Consensus-Protocols/QuorumIBFT.md
NicolasMassart: Is that a way to say that it's not going to work @rain-on ? Because if it's broken, we can't publish it as is.

MadelineMurray: @NicolasMassart - have tested this and the only modification needed is the 'istanbul' -> 'ibft' one. Included in docs.

mkdocs.yml
MadelineMurray: Fixed

PR #1082, opened 2019-03-11
Add timeout when waiting for JSON-RPC, WebSocket RPC and Metrics services to stop

ajsutton:
## PR description

Avoids the risk of `Runner.close` hanging indefinitely if something causes one of the services it stops to hang.  We go on to close all of vertx anyway so it's unlikely to matter and hanging indefinitely with no messages definitely doesn't help.


pantheon/src/main/java/tech/pegasys/pantheon/Runner.java
```diff
@@ -98,20 +97,47 @@ public void close() throws Exception {
...
+
+  private void waitForServiceToStart(
+      final String serviceName, final CompletableFuture<?> startFuture) {
+    while (!startFuture.isDone()) {
+      try {
+        startFuture.get(60, TimeUnit.SECONDS);
+      } catch (final InterruptedException e) {
+        LOG.debug("Interrupted while waiting for service to start", e);
+        Thread.currentThread().interrupt();
+        return;
```
shemnon: Why return and not continue (retry)?  That would imply if the thread gets interrupted we don't care if it succeeds or fails to start up

ajsutton: If we mark the thread as interrupted and retry it will immediately throw interrupted exception again when we call `get` and if we don't re-mark the thread as interrupted then we're just ignoring the interrupt.

I'm going to go with throwing an exception so that startup explicitly fails and for an obvious reason.  Must admit, I'm not entirely against just ignoring the interrupt in this case either...

PR #1085, opened 2019-03-12
Improve imported block log line

shemnon:
Improve the import log line to include relevant metrics like number
of transacitons, ommers, and gas usage.
Also, this fixes a bug in that we get task times when metrics are not
turned on.

Sample line:
```
2019-03-11 21:03:50.245+00:00 | EthScheduler-Workers-3 | INFO  | BlockPropagationManager | Imported #36 / 20 tx / 0 om / 420,000 (4.0%) gas / (0xecb366469d88d9d75b20721581797c1eb1118557b93db9a460dc0e000b9ec7fe) in 0.143s.
```


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/task/AbstractEthTask.java
```diff
@@ -123,22 +126,16 @@ public final boolean isCancelled() {
...
         subTaskFutures.add(subTaskFuture);
-        subTaskFuture.whenComplete(
-            (r, t) -> {
-              subTaskFutures.remove(subTaskFuture);
-            });
-        return subTaskFuture;
+        subTaskFuture.whenComplete((r, t) -> subTaskFutures.remove(subTaskFuture));
       } else {
-        return completedExceptionally(new CancellationException());
+        completedExceptionally(new CancellationException());
```
ajsutton: This creates a new future that is then immediately thrown away so I think this whole else block can just be removed.

shemnon: Done.

PR #1086, opened 2019-03-12
Trim default metrics

shemnon:
## PR description

Testing has shown that RocksDB and BigQueue metrics has a negative
impact on performance under load.  Disable them by default.


shemnon:
opened histogram spike as PAN-2440

PR #1087, opened 2019-03-12
[PAN-2422] Add in-memory cache over world state download queue

mbaxter:
## PR description
Add `CachingTaskCollection` which provides an in-memory cache on top of existing task queue implementations.  Migrate from a `TaskQueue` interface to the more generic `TaskCollection` interface and update package name from `queue` to `tasks`.  Integrate new `CachingTaskCollection` into `WorldStateDownloader`.


services/tasks/src/test/java/tech/pegasys/pantheon/services/tasks/CachingTaskCollectionTest.java
```diff
@@ -0,0 +1,241 @@
...
+      final String testName, final TaskCollectionSupplier wrappedTaskCollection) throws Exception {
+    this.wrappedTaskCollection = wrappedTaskCollection.get();
+  }
+
+  private CachingTaskCollection<BytesValue> createCachingCollection(final int cacheSize) {
+    return new CachingTaskCollection<>(wrappedTaskCollection, cacheSize);
+  }
+
+  @Parameters(name = "Wrap {0}")
+  public static Collection<Object[]> getTestParametersForConfig() throws IOException {
```
ajsutton: Why do we need to test this wrapping two different implementations? If we have tests to ensure `InMemoryTaskQueue` and `RocksDbTaskQueue` individually implement `TaskCollection` correctly this we should be able to just test the caching later with either one of them (probably `InMemoryTaskQueue` for simplicity and faster tests).

mbaxter: The thought was it would give extra coverage and potentially reveal inconsistencies, but you're right - it slows down the tests and probably isn't worth it.  Updated.

Issue #1055, opened 2019-03-05
Implement DebugTraceBlock JSON RPC APIs


timbeiko:
### Description
Implement the DebugTraceBlock JSON-RPC call in Pantheon, which will return a stack trace containing the opcodes called by transactions in the block.  Three variants of the API should be implemented: 

* DebugTraceBlock, which takes in a valid RLP encoding of a block as input; 
  * **Note**, the parent of this block must be present or it will fail. 
* DebugTraceBlockByHash, which takes in the block hash as input; 
* DebugTraceBlockByNumber, which will take the block number as input. 

### Acceptance Criteria
* All three APIs listed above are implemented in Pantheon;
  * Inputs are as defined above;
  * Return is an RLP output, for example: 
```
{
  gas: 999,
  returnValue: "",
  structLogs: [{
      depth: 1,
      error: "",
      gas: 99999,
      gasCost: 2,
      memory: null,
      op: "PUSH1",
      pc: 0,
      stack: [],
      storage: {}
  }
```
* Implementation includes testing;
* Optional: add documentation for this feature, see the [Contribution Guidelines](https://github.com/PegaSysEng/pantheon/blob/master/CONTRIBUTING.md#documentation-style-guide).

### Additional information
For a similar API implemented in Pantheon, see [DebugTraceTransaction](https://github.com/PegaSysEng/pantheon/blob/5f8059d0559509042944488a611bcf554058d41a/ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/DebugTraceTransaction.java) and its [DebugTraceTransactionTest](https://github.com/PegaSysEng/pantheon/blob/5f8059d0559509042944488a611bcf554058d41a/ethereum/jsonrpc/src/test/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/DebugTraceTransactionTest.java).

kziemianek:
Hey, I would love to tackle this issue, can I ?

ajsutton:
@kziemianek That would be fantastic.  Feel free to ask questions here or on Gitter (http://gitter.im/PegaSysEng/pantheon) if you need any pointers.  And feel free to put up draft PRs to get early feedback on the work too.

arash009:
Closing this issue and moving the ticket to Jira for tracking -- [PAN-2506](https://pegasys1.atlassian.net/browse/PAN-2506).

[...]

PR #1088, opened 2019-03-12
[WIP] Add block trace RPC methods (#1055)

kziemianek:
## PR description

Adds rpc methods to debug block by providing block number, block hash or block itself encoded by RLP.

## Fixed Issue(s)
fixes: #1055


ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/results/DebugTraceBlockResult.java
```diff
@@ -0,0 +1,67 @@
...
+import tech.pegasys.pantheon.ethereum.jsonrpc.internal.processor.TransactionTrace;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+import com.fasterxml.jackson.annotation.JsonGetter;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+
+@JsonPropertyOrder({"gas", "structLogs"})
+public class DebugTraceBlockResult {
```
kziemianek: It's missing `value` property (it's listed in issue), any hints how block's value can be computed ?

ajsutton: The issue description is missing an important element here unforunately - the result should be an array of transaction trace results.  So:

```
[
{
  gas: 999,
  returnValue: "",
  structLogs: [{
      depth: 1,
      error: "",
      gas: 99999,
      gasCost: 2,
      memory: null,
      op: "PUSH1",
      pc: 0,
      stack: [],
      storage: {}
  },
  ... (more transactions)
]
```

The `returnValue` is then the return value of the transaction the same way `DebugTraceTransactionResult` extracts it.  In fact, I suspect `DebugTraceBlockResult` isn't required and the success response is just `return new JsonRpcSuccessResponse(request.getId(), listOfDebugTraceTransactionResult)`

kziemianek: Thanks, i'll change it.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/DebugTraceBlock.java
```diff
@@ -0,0 +1,51 @@
...
+  public String getName() {
+    return "debug_traceBlock";
+  }
+
+  @Override
+  public JsonRpcResponse response(final JsonRpcRequest request) {
+    final String input = parameters.required(request.getParams(), 0, String.class);
+    final Block block =
+        Block.readFrom(
+            RLP.input(BytesValue.fromHexString(input)), MainnetBlockHashFunction::createHash);
```
ajsutton: We probably want to avoid hard coding `MainnetBlockHashFunction` here and instead use a `ScheduleBasedBlockHashFunction` so that it automatically selects the right hash function based on the protocol schedule.  My suggestion would be to add a `BlockHashFunction` parameter to the constructor and create the `ScheduleBasedBlockHashFunction` back up in `JsonRpcMethodsFactory` so `DebugTraceBlock` avoids having a direct dependency on `ProtocolSchedule`.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/DebugTraceBlock.java
```diff
@@ -0,0 +1,51 @@
...
+    DebugTraceBlockResult result =
+        blockTracer.trace(block).map(DebugTraceBlockResult::new).orElse(null);
+    return new JsonRpcSuccessResponse(request.getId(), result);
```
ajsutton: We probably should return an error response with a message if the block parent isn't on the block chain rather than just success with a `null` result.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/processor/BlockReplay.java
```diff
@@ -41,13 +44,107 @@ public BlockReplay(
...
+                              action.performAction(
+                                  transaction,
+                                  header,
+                                  blockchain,
+                                  mutableWorldState,
+                                  transactionProcessor))
+                      .collect(Collectors.toList());
+              return Optional.of(new BlockTrace(transactionTraces));
+            })
+        .orElse(null);
```
ajsutton: I suspect we could avoid duplication through this class by setting up the versions of methods that accept a `Hash` to just lookup the block by hash and if it's present delegate to the version that accepts a `Block` rather than duplicating code.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/processor/BlockTracer.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Optional;
+
+/** Used to produce debug traces of blocks */
+public class BlockTracer {
+
+  private final BlockReplay blockReplay;
+
+  private final BlockReplay.Action<TransactionTrace> replayAction =
+      (transaction, header, blockchain, mutableWorldState, transactionProcessor) -> {
+        DebugOperationTracer tracer = new DebugOperationTracer(TraceOptions.DEFAULT);
```
ajsutton: Would be good to have the `TraceOptions` come from an optional extra param to the JSON-RPC calls rather than hard coding to default.

kziemianek:
@ajsutton thanks, I really apprecieate your comments and I'll follow Your suggestions, new commits comming soon.

kziemianek:
@ajsutton thanks for feedback, i'll follow Your suggestion :). I'm still exploring codebase.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/processor/BlockReplay.java
kziemianek: The block is provided as input to `DebugTraceBlock` method so that's why there is a signature with `Block` as an argument,  do we assume that provided block is already present ?

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/DebugTraceBlock.java
kziemianek: Absolutely, I wasn't aware of `ScheduleBasedBlockHashFunction` and had to pass something.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/DebugTraceBlock.java
```diff
@@ -0,0 +1,76 @@
...
+  @Override
+  public String getName() {
+    return "debug_traceBlock";
+  }
+
+  @Override
+  public JsonRpcResponse response(final JsonRpcRequest request) {
+    final String input = parameters.required(request.getParams(), 0, String.class);
+    final Block block =
+        Block.readFrom(RLP.input(BytesValue.fromHexString(input)), this.blockHashFunction);
```
ajsutton: This may throw `RLPException` if the input isn't valid RLP which we should probably catch and return a meaningful error. Currently it would result in an internal server error response.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/DebugTraceJsonRpcMethod.java
```diff
@@ -0,0 +1,39 @@
...
+import tech.pegasys.pantheon.ethereum.debug.TraceOptions;
+import tech.pegasys.pantheon.ethereum.jsonrpc.internal.JsonRpcRequest;
+import tech.pegasys.pantheon.ethereum.jsonrpc.internal.parameters.JsonRpcParameter;
+import tech.pegasys.pantheon.ethereum.jsonrpc.internal.processor.TransactionTraceParams;
+
+import java.util.Optional;
+
+public interface DebugTraceJsonRpcMethod {
+
+  default TraceOptions getTraceOptions(
```
ajsutton: This would probably make more sense as a static method rather than a default method. It's probably clearer to pass in the `transactionTraceparamsParameterIndex` as a param to the method than using the second default method.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/processor/BlockReplay.java
```diff
@@ -41,13 +44,88 @@ public BlockReplay(
...
+  }
+
+  private <T> Optional<T> performActionPerBlockTransaction(
+      final Hash blockHash, final BlockTransactionAction<T> action) {
+    return getBlock(blockHash)
+        .map(block -> performActionPerBlockTransaction(block.getHeader(), block.getBody(), action))
+        .orElse(Optional.empty());
+  }
+
+  private <T> Optional<T> performActionPerBlockTransaction(
```
ajsutton: This method name is a bit misleading.  It sounds like it's looping through calling `action` for each transaction in the block but actually appears to just call the action once. So maybe `performActionWithBlock` would be a better name?

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/processor/BlockReplay.java
```diff
@@ -62,47 +140,32 @@ public BlockReplay(
...
-              worldState.updater(),
-              blockHeader,
-              transaction,
-              spec.getMiningBeneficiaryCalculator().calculateBeneficiary(blockHeader),
-              new BlockHashLookup(blockHeader, blockchain));
-          return action.performAction(
-              transaction, blockHeader, blockchain, worldState, transactionProcessor);
-        });
+  @FunctionalInterface
+  private interface BlockTransactionAction<T> {
```
ajsutton: Maybe this should be `BlockAction`?

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/processor/BlockReplay.java
```diff
@@ -62,47 +140,32 @@ public BlockReplay(
...
+    Optional<T> perform(
+        BlockBody body,
+        BlockHeader blockHeader,
+        Blockchain blockchain,
+        MutableWorldState worldState,
+        TransactionProcessor transactionProcessor);
   }
 
+  @FunctionalInterface
   public interface Action<T> {
```
ajsutton: And maybe this is now `TransactionAction`?

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/processor/BlockTracer.java
```diff
@@ -0,0 +1,55 @@
...
+  public BlockTracer(final BlockReplay blockReplay) {
+    this.blockReplay = blockReplay;
+  }
+
+  public Optional<BlockTrace> trace(final Hash blockHash, final DebugOperationTracer tracer) {
+    return Optional.of(blockReplay.block(blockHash, prepareReplayAction(tracer)));
+  }
+
+  public Optional<BlockTrace> trace(final Block block, final DebugOperationTracer tracer) {
+    return Optional.of(blockReplay.block(block, this.prepareReplayAction(tracer)));
```
ajsutton: nit: the `this.` prefix isn't required here.

ethereum/jsonrpc/src/test/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/DebugTraceBlockTest.java
```diff
@@ -0,0 +1,139 @@
...
+                    parentBlock.calculateSize())));
+
+    final JsonRpcSuccessResponse response =
+        (JsonRpcSuccessResponse) debugTraceBlock.response(request);
+    final Collection<?> result = (Collection<?>) response.getResult();
+    assertEquals(2, result.size());
+  }
+
+  @Test
+  public void shouldReturnErrorResponseWhenParenBlockMissing() {
```
ajsutton: nit: missing `t` at the end of `Paren`.

kziemianek:
Sorry @ajsutton , I'm very busy lately. Thanks for commit and merge.

ajsutton:
@kziemianek Thank you for doing all the hard work. It's great to have this APIs available.

PR #1089, opened 2019-03-12
Renamed txpool transaction method in the api.

rojotek:
Renamed txpool transaction method in the api.


rojotek:
Closed this as it is superseeded by #1095

PR #1090, opened 2019-03-13
Updated examples to use default host and port

MadelineMurray:
Updated examples to use default host and port


docs/Reference/JSON-RPC-API-Methods.md
```diff
@@ -3,6 +3,13 @@ description: Pantheon JSON-RPC API methods reference
 
 # JSON-RPC API Methods
 
+## Endpoint Host and Port
```
NicolasMassart: Maybe make that more visible using a !!!attention

MadelineMurray: I've updated it.  That is better.

PR #1092, opened 2019-03-13
Better bootnodes option error message

NicolasMassart:
## PR description

change --bootnodes option to a method to be able to handle the check and error

Handling this as a method enables to keep the parameter as a string and not a
strongly typed value. It then is always accepted by the command line so it's
always considered as a param and doesn't raise an unknown param error but also
immediately checked against the enode pattern and intelligible matching error
is displayed.

Converter is removed and only a simple lambda is used in a try block
leaving the handling of the error message close to the option thus making code
easier to understand. Anyway, as we use a Collection of Strings as input no
converter is needed anymore.

Update and add tests to check non regression and handle the new behaviour.
Also this change can change the order in which options are handled in the spec.
So tests are updated to check for the presence of options but not specifying a
static order. verifyOptionsConstraintLoggerCall now loops trough the options
to check they are contained in the error string but not in a specific order.

## Fixed Issue(s)

fixes PAN-2387 and PAN-2015


pantheon/src/test/java/tech/pegasys/pantheon/cli/PantheonCommandTest.java
```diff
@@ -756,22 +746,33 @@ public void callingWithBootnodesOptionButNoValueMustPassEmptyBootnodeList() {
     assertThat(commandErrorOutput.toString()).isEmpty();
   }
 
-  @Ignore(
-      "NC-2015 - Temporarily enabling zero-arg --bootnodes to permit 'bootnode' configuration, which changes the error.")
   @Test
-  public void callingWithInvalidBootnodesMustDisplayErrorAndUsage() {
+  public void callingWithValidBootnodeMustSuccess() {
```
lucassaldanha: Maybe rename to `callingWithValidBootnodeMustSucceed`?

PR #1097, opened 2019-03-14
Add pipe stage name to thread while executing

ajsutton:
## PR description
To make it easier to identify which threads are performing which tasks, pipeline will add the stage name to the end of the thread name while the stage is executing.  This makes it much easier to use tools like VisualVM to explore performance.


services/pipeline/src/main/java/tech/pegasys/pantheon/services/pipeline/Pipeline.java
```diff
@@ -49,7 +49,7 @@
 
   Pipeline(
       final Pipe<?> inputPipe,
-      final Collection<Runnable> stages,
+      final Collection<Stage> stages,
```
shemnon: Why Collection?  Does the order of the stages matter?  That would imply List.

ajsutton: It doesn't matter at this point - we're going to execute them all in their own threads anyway.  The order of stages is really enforced by the way the pipes connect them.

PR #1098, opened 2019-03-14
Support flat mapping with multiple threads

ajsutton:
## PR description
Adds a flatMap equivalent of `thenProcessInParallel` so that 1-to-many (including 1-to-none) operations can be performed with multiple concurrent threads.


services/pipeline/src/test/java/tech/pegasys/pantheon/services/pipeline/PipelineBuilderTest.java
```diff
@@ -232,6 +232,42 @@ public void shouldProcessInParallel() throws Exception {
...
+                "stageName",
+                value -> {
+                  if (value == 2) {
+                    try {
+                      latch.await();
+                    } catch (InterruptedException e) {
+                      throw new RuntimeException(e);
+                    }
+                  }
+                  return Stream.of(value.toString());
```
mbaxter: (optional) The functionality might be a little clearer if you output multiple values here

PR #1109, opened 2019-03-15
Adding JSON-RPC permissions to Pantheon TOML example file

CjHare:
## PR description
When updating my local Pantheons to run, I needed these settings to get Pantheon JSON-RPC running again,


docs/Configuring-Pantheon/Using-Configuration-File.md
```diff
@@ -40,10 +40,13 @@ Specific differences between the command line and the TOML file format are:
     
     rpc-http-host="5.6.7.8"
     rpc-http-port=5678
+    rpc-http-apis=["ETH","NET","WEB3"]
+    rpc-http-cors-origins=["*"]
+    host-whitelist=["127.0.0.1","localhost"]
```
ajsutton: This 127.0.0.1 and localhost should be the default for host-whitelist so shouldn't be required here.

rpc-http-apis should default to ETH, NET and WEB3 as well.

So you should only need the rpc-http-cors-origins and then only if you want to open your local node to every website you visit.

CjHare: true, the `rpc-http-apis` I was using was actually a long list of non-default apis.

CjHare:
Don't actually need this PR after all, the default options work for the default use cases

PR #1117, opened 2019-03-18
Pipeline Improvements

ajsutton:
## PR description
When creating batches as part of a pipeline, instead of having a dedicated thread that creates the batches, do it inline via a new `BatchingReadPipe` which wraps a normal `ReadPipe`.

Also improve the metrics we capture for pipelines so we capture number of tasks added and removed from each pipeline making it possible to calculate how many tasks are in each pipe.


services/pipeline/src/test/java/tech/pegasys/pantheon/services/pipeline/BatchingReadPipeTest.java
```diff
@@ -0,0 +1,134 @@
...
+
+    assertThat(batchingPipe.hasMore()).isFalse();
+
+    assertThat(batchingPipe.get()).isNull();
+    assertThat(batchingPipe.poll()).isNull();
+  }
+
+  @Test
+  public void shouldAddAtMostOneItemWhenDraining() {
+    // Collecting a batch of batches is pretty silly so only ever drain one batch at a time.
```
mbaxter: 🤔

ajsutton: For the record, nothing actually ever calls this but it seems better to give it a vaguely sane implementation than to throw an exception and not implement the underlying interface properly.

PR #1119, opened 2019-03-18
#626 add net enode json rpc endpoint

### LINKED ISSUE #626: Add net_enode json rpc endpoint
A simple endpoint with no params that will return the current enode in the result.

eg
```
{
  "id":1,
  "jsonrpc": "2.0",
  "result": "enode://6f8a80d14311c39f35f516fa664deaaaa13e85b2f7493f37f6144d86991ec012937307647bd3b9a82abe2974e1407241d54947bbb39763a4cac9f77166ad92a0@10.3.58.6:30303?discport=30301"
}
```

mbergstrand:
## PR description
Add net_enode to json RPC. 

## Fixed Issue(s)
fixes #626


ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/NetEnode.java
```diff
@@ -0,0 +1,59 @@
...
+
+  @Override
+  public String getName() {
+    return "net_enode";
+  }
+
+  @Override
+  public JsonRpcResponse response(final JsonRpcRequest req) {
+    try {
+      final StringBuilder response = new StringBuilder();
```
lucassaldanha: I think it would be simpler to use a structure like:
```
String enodeURI = p2pNetwork.getAdvertisedPeer().map(Peer::getEnodeURI).orElse("");
```

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/NetEnode.java
```diff
@@ -0,0 +1,59 @@
...
+      p2pNetwork
+          .getAdvertisedPeer()
+          .ifPresent(
+              advertisedPeer -> {
+                response.append(advertisedPeer.getEnodeURI());
+              });
+      return new JsonRpcSuccessResponse(req.getId(), response.toString());
```
lucassaldanha: I'll discuss with the team what we should do when peer discovery is disabled (and getAdvertisedPeer() returns empty). I'm not sure if returning an empty string is the right approach.
Anyway, this shouldn't block us from merging this PR after addressing the other bits.

lucassaldanha: Take a look at `EthHashrate`. When someone requests the hash rate that the node is mining. When we aren't mining, we return an error json rpc response with a message explaining why. We should do something similar here.

ethereum/jsonrpc/src/test/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/NetEnodeTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+    final JsonRpcRequest request = netEnodeRequest();
+    final JsonRpcResponse expectedResponse =
+        new JsonRpcSuccessResponse(request.getId(), advertisedPeer.get().getEnodeURI());
+
+    assertThat(method.response(request)).isEqualToComparingFieldByField(expectedResponse);
+  }
+
+  private JsonRpcRequest netEnodeRequest() {
+    return new JsonRpcRequest("2.0", "admin_enode", new Object[] {});
```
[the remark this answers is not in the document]
mbergstrand: Corrected. Don't know how I missed that.

ethereum/jsonrpc/src/test/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/NetEnodeTest.java
```diff
@@ -0,0 +1,67 @@
...
+  private final DefaultPeer defaultPeer = new DefaultPeer(nodeId, "1.2.3.4", 7890, 30303);
+  private final Optional<Peer> advertisedPeer = Optional.of(defaultPeer);
+
+  @Mock private P2PNetwork p2PNetwork;
+
+  @Before
+  public void before() {
+    this.method = new NetEnode(p2PNetwork);
+  }
+
```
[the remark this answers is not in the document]
mbergstrand: I added that, and also a few more tests for better coverage.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/NetEnode.java
mbergstrand: Thank you, that is much cleaner.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/NetEnode.java
mbergstrand: I have refactored to check isP2pEnabled and return an appropriate  error response. 
Still have a question if P2P is enabled but the enode cannot be found.
For now I am simply returning P2P disabled there as well, but I left it separate so that can easily be changed if needed.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/NetEnode.java
```diff
@@ -0,0 +1,66 @@
...
+  @Override
+  public String getName() {
+    return "net_enode";
+  }
+
+  @Override
+  public JsonRpcResponse response(final JsonRpcRequest req) {
+    try {
+      if (p2pNetwork.isP2pEnabled()) {
+        String enodeURI = p2pNetwork.getAdvertisedPeer().map(Peer::getEnodeURI).orElse("");
```
lucassaldanha: the `getEnodeURI()` method that returns a String was replaced with `getEnodeURLString()`. That's why the compiler is complaining about your method now.

PR #1128, opened 2019-03-19
[MINOR] added fromHexStringStrict to check for exactly 20 byte addresses

macfarla:
## PR description
Added method to create Address from exactly 20 bytes to be used by CLI (tests remain using existing method which pads shorter values)


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/Address.java
```diff
@@ -96,6 +96,21 @@ public static Address fromHexString(final String str) {
     return new Address(BytesValue.fromHexStringLenient(str, SIZE));
   }
 
+  /**
+   * Parse an hexadecimal string representing an account address.
+   *
+   * @param str An hexadecimal string representing a valid account address (strictly 20 bytes).
+   * @return The parsed address.
+   * @throws NullPointerException if the provided string is {@code null}.
```
Errorific: Looks like you return null rather than throw a null pointer exception when the hexstring is null

Errorific: and thinking further probably better to throw an illegal argument exception from checkArgument rather than a null pointer

macfarla: this is also true for the existing fromHexString method. Do I leave that one as is?

PR #1130, opened 2019-03-19
[PAN-2394] Synchronizer returns false if it is in sync

macfarla:
## PR description
eth_syncing now returns false if the node is in sync

## Fixed Issue(s)
fixes #PAN-2394


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/DefaultSynchronizer.java
```diff
@@ -132,7 +132,7 @@ private void startFullSync() {
 
   @Override
   public Optional<SyncStatus> getSyncStatus() {
-    if (!started.get()) {
+    if (!started.get() || syncState.isInSync()) {
```
Errorific: isInSync is true while we're within 5 blocks of head. Does that behaviour match this call or do we want to return something while we're closer to head?

lucassaldanha: I believe what we need it to get the SyncResult and check if the distance to the head is equal to zero.

macfarla: modified to use if (syncState.syncStatus().getCurrentBlock() == syncState.syncStatus().getHighestBlock())

PR #1132, opened 2019-03-19
[PAN-2347] nodepermissioningcontroller used for devp2p connection filtering

Errorific:
Need to make sure that incoming and outgoing devp2p connections are abiding by the nodePermissioningControllers rules. Also some associated quality of life refactors.


ethereum/permissioning/src/main/java/tech/pegasys/pantheon/ethereum/permissioning/node/NodePermissioningController.java
```diff
@@ -65,8 +65,7 @@ public void startPeerDiscoveryCallback(final Runnable peerDiscoveryCallback) {
     return syncStatusNodePermissioningProvider;
   }
 
-  @VisibleForTesting
-  List<NodePermissioningProvider> getProviders() {
+  public List<NodePermissioningProvider> getProviders() {
```
lucassaldanha: Given that we are exposing our list of providers, would it make sense to change it to an unmodifiable list?

Errorific: Will go private again in PAN-2449

PR #1133, opened 2019-03-19
PAN-2286 - quieter exceptions when network is unreachable

shemnon:
## PR description

Lower the log level of discovery connections to unreachable networks,
such as a box without IPv6 enabled connecting to an IPv6 peer.

## Fixed Issue(s)

PAN-2286

### DIFF
--- a/ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/PeerDiscoveryAgent.java
+++ b/ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/PeerDiscoveryAgent.java
@@ -41,6 +41,7 @@ import tech.pegasys.pantheon.util.Subscribers;
 import tech.pegasys.pantheon.util.bytes.BytesValue;
 
 import java.net.InetSocketAddress;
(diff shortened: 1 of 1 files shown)


ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/PeerDiscoveryAgent.java
```diff
@@ -204,11 +205,16 @@ protected void handleOutgoingPacket(final DiscoveryPeer peer, final Packet packe
         .whenComplete(
             (res, err) -> {
               if (err != null) {
-                LOG.warn(
-                    "Sending to peer {} failed, packet: {}",
-                    peer,
-                    wrapBuffer(packet.encode()),
-                    err);
+                if (err instanceof SocketException && err.getMessage().contains("unreachable")) {
```
mbaxter:
```suggestion
                if (err instanceof SocketException && err.getMessage().equals("Network is unreachable")) {
```

shemnon:
The text varies between platforms.  "Network unreachable" also shows up in
other jdk source repos.

PR #1139, opened 2019-03-20
Disconnect peers before the pivot block while fast syncing

ajsutton:
## PR description
Automatically configure the trailing peers limit during a fast sync to consider any peer with a chain height below the pivot block to be a trailing peer and having a trailing peer limit of 0.  ie disconnect any peers that haven't reached the pivot block because they're useless to us.

After fast sync completes we revert back to the current unlimited trailing peer configuration.


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/fastsync/FastSyncDownloader.java
```diff
@@ -62,6 +66,12 @@ public FastSyncDownloader(
 
   private FastSyncState storeState(final FastSyncState state) {
     fastSyncStateStorage.storeState(state);
+    if (state.getPivotBlockNumber().isPresent()) {
+      trailingPeerRequirements =
```
mbaxter: This is a little unexpected in a "storeState" method

ajsutton: I was thinking along the lines of storing the new trailing peer requirements was storing state but probably is clearer to just have its own method.

PR #1141, opened 2019-03-20
Added static nodes doc

MadelineMurray:
Added static nodes doc and reorganised networking content.


docs/Configuring-Pantheon/Networking/Managing-Peers.md
```diff
@@ -0,0 +1,66 @@
...
+can be used to disable P2P peer discovery.
+Set this option to `false` if you are running a test node or a network with fixed nodes.
+
+Use the [`admin_addPeer`](../../Reference/JSON-RPC-API-Methods.md#admin_addpeer) JSON-RPC method to add peers
+or configure [static nodes](#static-nodes) when [`--discovery-enabled`](../../Reference/Pantheon-CLI-Syntax.md#discovery-enabled) 
+is set to false. 
+
+## Static Nodes
+
+Use a `static-nodes.json` file to configure static nodes. The node connects to the listed nodes on startup.
```
NicolasMassart: what is a static node ? What is the goal of having some, what's the use case ? In what case is it mandatory ?

NicolasMassart:
Two PRs, one for the reorganisation and another for the static node addition would have been less confusing.

docs/Configuring-Pantheon/Networking/Managing-Peers.md
rain-on: Its a Geth thing, and is wanted by Kaleido - what is it used for 🤷‍♂️ It effectively says "on startup, attempt to connect to nodes X-Y-Z" (and continue to do so if you lose connection).
This is SLIGHTLY different to permissioning, which says "I am allowed to talk to Nodes X-Y-Z", but does not explicitly try to contact X-Y-Z (I don't think).

docs/Configuring-Pantheon/Networking/Managing-Peers.md
```diff
@@ -0,0 +1,66 @@
...
+    [
+    "enode://cea71cb65a471037e01508cebcc178f176f9d5267bf29507ea1f6431eb6a5dc67d086dc8dc54358a72299dab1161febc5d7af49d1609c69b42b5e54544145d4f@127.0.0.1:30303",
+    "enode://ca05e940488614402705a6b6836288ea902169ecc67a89e1bd5ef94bc0d1933f20be16bc881ffb4be59f521afa8718fc26eec2b0e90f2cd0f44f99bc8103e60f@127.0.0.1:30304"    
+    ]
+    ``` 
+
+!!! note
+    Each node has a `static-nodes.json` file. Ensure each node in the network has the same `static-nodes.json` file. 
+
+Nodes added by [`admin_addPeer`](../../Reference/JSON-RPC-API-Methods.md#admin_addpeer) are not added to the
```
rain-on: Is it worth saying - if the NodeID in the enode (passed to AdminPeer, or staticNodes) does not match the ACTUAL node ID of the remote peer - the connection will fail (appear to timeout) - but remote node will show a failure regarding "Invalid MAC".
A better error log may be forthcoming, but maybe not soon.

docs/Configuring-Pantheon/Networking/Managing-Peers.md
```diff
@@ -0,0 +1,66 @@
...
+!!! example 
+    ```json
+    [
+    "enode://cea71cb65a471037e01508cebcc178f176f9d5267bf29507ea1f6431eb6a5dc67d086dc8dc54358a72299dab1161febc5d7af49d1609c69b42b5e54544145d4f@127.0.0.1:30303",
+    "enode://ca05e940488614402705a6b6836288ea902169ecc67a89e1bd5ef94bc0d1933f20be16bc881ffb4be59f521afa8718fc26eec2b0e90f2cd0f44f99bc8103e60f@127.0.0.1:30304"    
+    ]
+    ``` 
+
+!!! note
+    Each node has a `static-nodes.json` file. Ensure each node in the network has the same `static-nodes.json` file.
```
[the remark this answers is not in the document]
MadelineMurray: Changed to this to we recommend

docs/Configuring-Pantheon/Networking/Managing-Peers.md
```diff
@@ -0,0 +1,83 @@
...
+
+To configure a network of static nodes: 
+
+1. List [enode URLs](../Node-Keys.md#enode-url) of the nodes in the [`static-nodes.json` file](#static-nodesjson-file).
+
+1. Save the `static-nodes.json` file in the data directory of each node. 
+
+1. Start Pantheon with discovery disabled using [`--discovery-enabled=false`](../../Reference/Pantheon-CLI-Syntax.md#discovery-enabled).   
+
+To update the static peers, use the [`admin_addPeer`](../../Reference/JSON-RPC-API-Methods.md#admin_addpeer)
```
rain-on: Maybe "To Modify the static peers at run time...."
''' Note-  runtime modifications of static peers are not persisted between runs

[yeah, being a pedant]

docs/Configuring-Pantheon/Networking/Managing-Peers.md
```diff
@@ -0,0 +1,83 @@
...
+
+!!! note
+    Updates to static nodes are not persisted through node restarts. The `static-nodes.json` file
+    is not updated by `admin_addPeer` and `admin_removePeer` methods. 
+    
+    Nodes outside of the static nodes are not prevented from connecting.  To prevent nodes from connecting,
+    use [Permissioning](../../Permissions/Permissioning.md). 
+    
+!!! caution 
+    If connections are timing out, ensure the node ID in the [enode URL](../Node-Keys.md#enode-url) is correct.
```
rain-on: This one is hard :/ As far as the user knows, AddPeer returns True - however, the peer never gets connected (so won't turn up in Admin_getPeers)... in retrospect, the timeout is a furphy - its more like:
If the added peer does not appear in the reported peer list, check the supplied enode (including NodeID and endpoint) is correctly defined, the node is running, is listening for TCP connections on the endpoint and has not reached maxpeers.

docs/Configuring-Pantheon/Networking/Managing-Peers.md
MadelineMurray: Updated

docs/Configuring-Pantheon/Networking/Managing-Peers.md
MadelineMurray: Updated

PR #1153, opened 2019-03-21
Simplifying node permissioning ATs

lucassaldanha:
## PR description
* Removed JSON-RPC specific node permissioning tests
* Improved NodePermissioning ATs to use available JSON-RPC APIs and assert on the expected outcome of invoking them


acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/permissioning/NodesWhitelistAcceptanceTest.java
```diff
@@ -59,12 +60,25 @@ public void permissionedNodeShouldDiscoverOnlyAllowedNodes() {
...
 
+  @Test
+  public void forbiddenNodeAddedToWhitelistCanConnectToPermissionedNode() {
+    permissionedNode.verify(net.awaitPeerCount(2));
+
+    // add node to the whitelist
+    permissionedNode.verify(perm.addNodesToWhitelist(forbiddenNode));
+    permissionedNode.verify(perm.getNodesWhitelist(3));
+
+    permissionedNode.verify(admin.addPeer(((RunnableNode) forbiddenNode).enodeUrl()));
```
ajsutton: Can we change `admin.addPeer` to take a `Node` instead of the node URL and then it can do the cast and extract inside the DSL.

PR #1156, opened 2019-03-22
[PAN-2327] Notify of dropped messages

rojotek:
Added ability to subscribe to dropped transactions from the transaction pending pool.
Implemented subscription webservice to support this.
Added metrics to the pending transactions, tracking the number of local and remote transactions in the pool.


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/PendingTransactions.java
```diff
@@ -85,6 +125,11 @@ public void removeTransaction(final Transaction transaction) {
                     transactionsBySender.remove(transaction.getSender());
                   }
                 });
+        if (removedTransactionInfo.isReceivedFromLocalSource()) {
+          localTransactionCounter.inc(-1);
+        } else {
+          remoteTransactionCounter.inc(-1);
```
ajsutton: Counters should never be decremented, they only go up.  We should have a separate "transactions_dropped_total" counter that we increment here (probably with a "source" label for local & remote).

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/websocket/subscription/pending/PendingTransactionDroppedSubscriptionService.java
```diff
@@ -0,0 +1,52 @@
...
+  public void onTransactionDropped(final Transaction transaction) {
+    notifySubscribers(transaction.hash());
+  }
+
+  private void notifySubscribers(final Hash pendingTransaction) {
+    final List<Subscription> subscriptions = pendingTransactionSubscriptions();
+
+    for (final Subscription subscription : subscriptions) {
+      subscriptionManager.sendMessage(
+          subscription.getId(), new PendingTransactionResult(pendingTransaction));
```
ajsutton: Should we create the `PendingTransactionResult` once and reuse it to send to each subscription?

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/PendingTransactions.java
```diff
@@ -54,25 +58,61 @@
   private final Collection<PendingTransactionListener> listeners =
       newSetFromMap(new ConcurrentHashMap<>());
 
+  private final Collection<PendingTransactionDroppedListener> transactionDroppedListeners =
+      newSetFromMap(new ConcurrentHashMap<>());
```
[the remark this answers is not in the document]
rojotek: 👍

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/websocket/subscription/pending/PendingTransactionDroppedSubscriptionService.java
rojotek: Yes

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/PendingTransactions.java
```diff
@@ -51,28 +55,71 @@
...
+    addRemovedCounter(transactionCounter, "remote", "dropped");
+  }
+
+  private void addRemovedCounter(
+      final LabelledMetric<Counter> transactionCounter,
+      final String source,
+      final String addedToBlock) {
+    transactionRemovedCounters.put(
+        new DecrementTransactionKey("local".equals(source), "addedToBlock".equals(addedToBlock)),
+        transactionCounter.labels(source, "removed", addedToBlock));
```
ajsutton: I think you've probably over-optimised this (probably because of my previous suggestions, sorry). It's ok to call `.labels` multiple times. It's very slightly faster if you can just store the result like with `localTransactionAddedCounter` and `removeTransactionAddedCounter`.  Given there's both a local/remote and addedToBlock/dropped axis, I'd just store the `LabelledMetric<Counter>` and then when you need to increment call `.labels(...).inc()`. The indirection of looking up in a map likely nullifies any benefit to the caching.

PR #1169, opened 2019-03-26
RocksDB Statistics in Metrics

shemnon:
## PR description

Expose all the RocksDB statistics as metrics


services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/RocksDbKeyValueStorage.java
```diff
@@ -81,6 +232,22 @@ private RocksDbKeyValueStorage(
...
+            "RocksDB reported statistics for " + ticker.name(),
+            () -> stats.getTickerCount(ticker));
+      }
+
+      if (metricsSystem instanceof PrometheusMetricsSystem) {
+        for (final HistogramType histogram : HISTOGRAMS) {
+          ((PrometheusMetricsSystem) metricsSystem)
+              .addCollector(ROCKSDB_STATS, histogramToCollector(histogram));
+        }
+      }
```
mbaxter: Can you wrap all of this new logic up in a utility class in the metrics package?  That way we can reuse this for other RocksDB instances as needed.  It looks like you could create a `RocksDBStats` class with methods `getStatistics()` and `registerMetrics(MetricsSystem metrics, MetricsCategory category)`.

services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/RocksDbKeyValueStorage.java
```diff
@@ -35,6 +46,144 @@
...
+    TickerType.READ_AMP_ESTIMATE_USEFUL_BYTES, // COUNT: 0
+    TickerType.READ_AMP_TOTAL_READ_BYTES, // COUNT: 0
+    TickerType.NUMBER_RATE_LIMITER_DRAINS, // COUNT: 0
+    TickerType.NUMBER_ITER_SKIP, // COUNT: 0
+    TickerType.NUMBER_MULTIGET_KEYS_FOUND, // COUNT: 0
+  };
+
+  // Histograms - treated as prometheus summaries
+  static final HistogramType[] HISTOGRAMS = {
+    HistogramType.DB_GET, // COUNT : 4716170 SUM : 29514961
```
mbaxter: What are all of these comments for?

metrics/src/main/java/tech/pegasys/pantheon/metrics/MetricCategory.java
```diff
@@ -24,13 +24,14 @@
   PEERS("peers"),
   PROCESS("process", false),
   ROCKSDB("rocksdb"),
+  ROCKSDB_STATS("rocksdb", false),
```
mbaxter: Also - might be worth making these category names more accurate while you're in here.  Perhaps: `ROCKSDB_KVSTORE` / `ROCKSDB_KVSTORE_STATS`?

services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/RocksDbKeyValueStorage.java
shemnon: Previous measurements.  I'm was trying to see if there were stats we should get rid of.  But the performance latency impact of keeping all of them is almost unmeasurable.

services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/RocksDbKeyValueStorage.java
shemnon: If we keeps the statistics object in rocksdb and treat it like a part of the options then we can have `RocksDBStats` have just static methods.

metrics/src/main/java/tech/pegasys/pantheon/metrics/MetricCategory.java
shemnon: I went with `KVSTORE_ROCKSDB` in case we ever switch DBs it has good hierarchy: `KVSTORE_MEMDB`, `KVSTORE_JDBC`, etc.

services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/RocksDbKeyValueStorage.java
shemnon: Deleted

metrics/src/main/java/tech/pegasys/pantheon/metrics/MetricCategory.java
mbaxter: yeah - thats better

services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/RocksDBStats.java
```diff
@@ -0,0 +1,198 @@
...
+    TickerType.NUMBER_DB_SEEK_FOUND,
+    TickerType.NUMBER_DB_NEXT_FOUND,
+    TickerType.NUMBER_DB_PREV_FOUND,
+    TickerType.ITER_BYTES_READ,
+    TickerType.NO_FILE_CLOSES,
+    TickerType.NO_FILE_OPENS,
+    TickerType.NO_FILE_ERRORS,
+    // TickerType.STALL_L0_SLOWDOWN_MICROS,
+    // TickerType.STALL_MEMTABLE_COMPACTION_MICROS,
+    // TickerType.STALL_L0_NUM_FILES_MICROS,
```
mbaxter: I assume we want to keep these here for visibility of all of the options?

services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/RocksDBStats.java
```diff
@@ -0,0 +1,198 @@
+package tech.pegasys.pantheon.services.kvstore;
```
mbaxter: What do you think about moving this to: `tech.pegasys.pantheon.metrics.rocksdb`.  We have a similar paradigm for vertx utilities in `tech.pegasys.pantheon.metrics.vertx`.

services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/RocksDBStats.java
```diff
@@ -0,0 +1,198 @@
...
+    HistogramType.BYTES_PER_MULTIGET,
+    HistogramType.BYTES_COMPRESSED,
+    HistogramType.BYTES_DECOMPRESSED,
+    HistogramType.COMPRESSION_TIMES_NANOS,
+    HistogramType.DECOMPRESSION_TIMES_NANOS,
+    HistogramType.READ_NUM_MERGE_OPERANDS,
+  };
+
+  static void registerRocksDBMetrics(
+      final Statistics stats, final PrometheusMetricsSystem metricsSystem) {
```
mbaxter: If you move this to the `metrics` package and inject `MetricsCategory`, we can reuse this elsewhere.  For example, in the `RocksDbTaskQueue`.

shemnon: That would give the metrics package a dependency on RocksDB, and then every package that uses metrics would drag that dependency along as deadweight.

services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/RocksDBStats.java
shemnon: yes, to make it clear their absence is intentional.

mbaxter: (optional) Might be worth an explanatory comment :)

services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/RocksDBStats.java
mbaxter: Yeah, thats a good point.  And that's the case now with vertx.  I guess the "right" thing to do would be to have independent `vertxmetrics` and `rocksdbmetrics` packages.

shemnon:
I will do the refactoring in a follow on PR because creating the module that will be needed has some long tendrils.

PR #1177, opened 2019-03-27
Updated permissioning CLI options and permissioned network tutorial 

MadelineMurray:
Updated permissioning CLI options and permissioned network tutorial


docs/Getting-Started/Run-Docker-Image.md
```diff
@@ -144,7 +145,11 @@ Where `mygenesis.json` is your custom configuration file and `path` is the absol
 
 ## Permissions Configuration File 
 
-Specify a permissions configuration file. This is equivalent to specifying the `--permissions-config-file` option.
+Specify a permissions configuration file. This is equivalent to specifying the `--permissions-accounts-config-file`
```
NicolasMassart: you write about two options but to me they are the same `--permissions-accounts-config-file` and `--permissions-accounts-config-file`

docs/Reference/Pantheon-CLI-Syntax.md
```diff
@@ -550,60 +550,123 @@ The default is 30303. Ports must be [exposed appropriately](../Configuring-Panth
...
+    This option is not used when running Pantheon from the [Docker image](../Getting-Started/Run-Docker-Image.md#permissions-configuration-file).
+
+### permissions-nodes-contract-address
+
+```bash tab="Syntax"
+--permissions-nodes-contract-address=<ContractAddress>
+```
+
+```bash tab="Example Command Line"
+--permissions-nodes-contract-address=xyz
```
NicolasMassart: shouldn't we have a valid address that would also show if we have to use the 0x prefix ?

docs/Getting-Started/Run-Docker-Image.md
MadelineMurray: Good point - fixed the second node to be nodes.

docs/Reference/Pantheon-CLI-Syntax.md
MadelineMurray: Yes - I'll add this when I'm doc'ing the functionality (not available yet).  Adding a note to the JI so I don't forget.

PR #1187, opened 2019-03-29
Added droppedPendingTransactions subscription

MadelineMurray:
Added droppedPendingTransactions subscription


docs/Using-Pantheon/RPC-PubSub.md
```diff
@@ -234,6 +235,41 @@ This means the subscription can publish notifications for the same pending trans
...
+    
+### Dropped Transactions
+
+Use the `droppedPendingTransactions` parameter with `eth_subscribe` to be notified of transactions 
+dropped from the transaction pool for the node. 
+
+The dropped transactions subscription returns the transaction hashes of the dropped transactions. 
+
+If a chain reorganization occurs, transactions are resubmitted to be included in the new canonical chain. 
+This means the subscription can publish notifications for the same dropped transaction more than once.
```
ajsutton: This gives me the impression that I'd get a transaction dropped notification if the transaction is added to a block, but we specifically don't send a notification when the transaction is included in a block on the canonical chain.

There are some slightly contrived examples where reorgs may cause a dropped transaction to be re-added to the transaction pool, but the most common case is just that we dropped the transaction then received it from another peer again.  We should probably just say something like:
Transactions may be re-added to the transaction pool from a variety of sources after being dropped. As a result it's possible to receive multiple transaction dropped notifications for the same transaction.

MadelineMurray: Fixed

docs/Using-Pantheon/RPC-PubSub.md
```diff
@@ -243,8 +243,9 @@ dropped from the transaction pool for the node.
 
 The dropped transactions subscription returns the transaction hashes of the dropped transactions. 
 
-If a chain reorganization occurs, transactions are resubmitted to be included in the new canonical chain. 
-This means the subscription can publish notifications for the same dropped transaction more than once.
+Transactions can be re-added to the transaction pool from a variety of sources after being dropped. For example, 
+re-receiving the dropped transaction from a peer. As a result it's possible to receive multiple dropped transaction
```
ajsutton: I'm not convinced re-receiving is really a word (even if it was me that originally used it).  "For example, receiving a previously dropped transaction from a peer." ?

MadelineMurray: Your suggestion is much better than re-receiving. I'll update it

PR #1192, opened 2019-03-29
[PRIV] - WIP Privacy documentation

iikirilov:
## PR description

Improvements to privacy documentation.

### DIFF
--- a/docs/Privacy/Start-Pantheon-Privacy.md
+++ b/docs/Privacy/Start-Pantheon-Privacy.md
@@ -1,111 +1,95 @@
+## Start Pantheon with Privacy Enabled
 
-## Start Pantheon with Privacy
-
-The EEA methods are not enabled by default, follow the steps above to
-use the command line options. Pantheon/Enclave(Orion) needs to be started 
-when using privacy.
-
-### Pantheon
-
-#### rpc-http-api
-
-```bash tab="Example Command Line"
---rpc-http-api=EEA
-```
-
-```bash tab="Example Configuration File"
-rpc-http-api=["EEA"]
-```
-
-Comma-separated APIs to enable on the HTTP JSON-RPC channel.
-When you use this option, the `--rpc-http-enabled` option must also be specified.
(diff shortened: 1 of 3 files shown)


docs/Privacy/Start-Pantheon-Privacy.md
```diff
@@ -1,111 +1,95 @@
...
-### Set-up Privacy
-
-## Prerequisites 
-
-[Pantheon](../Installation/Install-Binaries.md) 
-
-[Curl (or similar web service client)](https://curl.haxx.se/download.html) 
-
-## Steps
+## Private Transaction walkthrough
```
iikirilov: I have reservations to write a walkthrough in the main Pantheon docs. I prefer to create a separate project (pantheon-privacy-examples or something) that will have the basic structure set-up. Then we can just link that project here and put docs about it.

In addition - I would like to question the relevance of having this walkthrough in the first-place. Quorum does not have such a walkthrough and I do not think it is popular enough to run multiple nodes on one local machine. I would like to point out here that we are also working on a dockerized environment for showcasing this. I feel I should put my attention on documentation there.

Does anyone strongly feel that the information below this heading is useful?

MadelineMurray: The information below the heading is more about describing the configuration (command line options, Orion config file etc) needed to set up multiple nodes rather than running multiple nodes on one machine (the nodes could be on different machines and the configuration steps are still required).  I'm going to do some updates to it to focus it on the configuration that's specific to setting up privacy which should cut down the overall size. I'm happy to own updating/rework etc on the privacy documentation merged so far. And agree it makes sense for you to focus on the documentation around the dockerized environment.

rojotek: So to restate what Madeline just said, having seperate documentation that helps people get their own environment setup is useful, and the point of the documentation.  I think that this type of documentation makes sense, and that Madeline would do a great job of writing it.

iikirilov: Perfect. So I have already outlined the basic command line arguments you need to set-up a Pantheon and Orion node on a local machine. Good luck with the rest! :)

MadelineMurray:
Superseeded by updated privacy documentation

PR #1194, opened 2019-03-31
[PAN-2427] Prep chain downloader for branch by abstraction

ajsutton:
## PR description
Refactor to make `ChainDownloader` an interface so that we can implement alternate chain download implementations. The primary benefit is allow us to use branch-by-abstraction to develop a pipeline based chain download in smaller steps rather than having to wholesale replace the chain download mechanism.

Also makes the abstractions more consistent between fast and full sync by introducing `FullSynchronizer` which bundles together the full sync chain download, block propagation manager and trailing peers config so `DefaultSynchronizer` is more clearly just switching between fast and full sync. It also let the full sync chain downloader adhere to the `ChainDownload` api and not need to expose the trailing peer config as well.


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/fullsync/FullSynchronizer.java
```diff
@@ -0,0 +1,73 @@
...
+      final ProtocolSchedule<C> protocolSchedule,
+      final ProtocolContext<C> protocolContext,
+      final EthContext ethContext,
+      final SyncState syncState,
+      final MetricsSystem metricsSystem) {
+    this.syncConfig = syncConfig;
+    this.protocolContext = protocolContext;
+    this.syncState = syncState;
+
+    this.blockPropagationManager =
```
mbaxter: I think the terminology here is getting a bit fuzzy.  I think the "FullSynchronizer" is really a "FullDownloaderFactory" (similarly "FastSynchronizer" -> "FastDownloaderFactory").  And I think that the `blockPropagationManager` should belong to the `Synchronizer`.  In my mind, the synchronizer encapsulates all things related to staying up-to-date with the network:  tracking peer stats, managing propagated blocks, downloading from the network.  And full / fast modes are only concerned with how we download data from the network.

ajsutton: Yep, I can see that working.  Done.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/fullsync/FullDownloaderFactory.java
```diff
@@ -0,0 +1,58 @@
...
+    this.protocolContext = protocolContext;
+    this.syncState = syncState;
+
+    this.chainDownloader =
+        FullSyncChainDownloader.create(
+            syncConfig, protocolSchedule, protocolContext, ethContext, syncState, metricsSystem);
+  }
+
+  public void start() {
+    chainDownloader.start();
```
mbaxter: To follow-through on the factory change, should this class now `create()` a `Downloader` that has a `start()` method?  Just seems weird to call `Factory.start()`.

ajsutton: We can't really because there are additional methods the "factories" need to expose - trailing peers for full sync and deleting the state for fast.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/DefaultSynchronizer.java
```diff
@@ -71,19 +79,11 @@ public DefaultSynchronizer(
...
-
-    this.fullSyncDownloader =
-        new FullSyncDownloader<>(
+    this.fullDownloaderFactory =
+        new FullDownloaderFactory<>(
             syncConfig, protocolSchedule, protocolContext, ethContext, syncState, metricsSystem);
-
-    fastSynchronizer =
-        FastSynchronizer.create(
+    this.fastSynchronizer =
```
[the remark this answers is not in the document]
ajsutton: Done

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/fullsync/FullDownloaderFactory.java
ajsutton: Maybe these are `FastDownloadManager` and `FullDownloadManager`?  Manager is a bit of a weasel word but I'm not seeing a better layout that would give clearer responsibilities at the moment.

mbaxter: I think you could get away with a `Downloader` that just exposes `start` and `calculateTrailingPeerRequirements`, then does any cleanup before the `start` future completes.  But feel free to just rename if you think that makes more sense.

ajsutton: I'm not sure we've got this structure quite right yet but I think I've found a reasonable landing point.  Fast sync has a `FastSyncDownloadFactory` which returns a `FastSyncDownload` - full sync is much simpler so doesn't need the factory and just creates a `FullSyncDownload` directly.

End result is that `DefaultSynchronizer` winds up with a `Optional<FastSyncDownload>` and a `FullSyncDownload` both of which have `start` and `calculateTrailingPeerRequirements`.

There's more tidy-up that could be done in this area but I'm hesitant to get too distracted.

PR #1195, opened 2019-04-01
Added content on using Truffle with Pantheon 

MadelineMurray:
Added content on using Truffle with Pantheon


docs/Using-Pantheon/Truffle.md
```diff
@@ -0,0 +1,59 @@
+description: Using Pantheon with Truffle
+<!--- END of page meta data -->
+
+# Using Pantheon with Truffle 
+
+Developing for Pantheon using Truffle is the same as using Truffle to develop for the public Ethereum networks. 
+Truffle supports Pantheon with the only difference being Pantheon does not implement private key management. 
+To use Pantheon with Truffle, you must configure the Truffle wallet.
```
lucassaldanha: I'm not sure if we should call it "THE Truffle wallet" or "A Truffle wallet". I believe there are other wallet implementations for truffle out there.

docs/Using-Pantheon/Truffle.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Install Truffle Wallet 
+
+To install the [Truffle wallet](https://github.com/trufflesuite/truffle-hdwallet-provider):
```
lucassaldanha: This seems to be deprecated now. We might need to update it soon. I believe the instructions will be the same. I'm not sure on what we need to update though.

docs/Using-Pantheon/Truffle.md
lucassaldanha: But this seems to be the "official" implementation so it should be fine.

macfarla: I would go with "a Truffle wallet" - there are a few

docs/Using-Pantheon/Truffle.md
macfarla: yeah instructions are the same, maybe just remove the github link

MadelineMurray: Done

docs/Using-Pantheon/Truffle.md
MadelineMurray: Done

docs/Using-Pantheon/Truffle.md
```diff
@@ -0,0 +1,59 @@
...
+
+```bash
+npm install --save truffle-hdwallet-provider@web3-one
+```
+
+!!!note
+    With Truffle 5, you must use a Web3 1.0 enabled wallet or the Truffle tasks hang.
+
+    `npm` requires `sudo` on Linux.
```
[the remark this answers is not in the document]
MadelineMurray: Removed

PR #1196, opened 2019-04-01
Pipeline chain download - target selection

ajsutton:
## PR description
Begins implementation of a pipeline based chain downloader.

Builds on #1194 so the new chain download process isn't currently in use (but unfortunately that makes the diff quite large).  `PipelineChainDownloader`, `DownloadPipelineFactory` and `PipelineChainDownloaderTest` are the new changes in this PR.

So far only the process of selecting a sync target, handling failures and cancellation are implemented.  The actual pipeline to download and import blocks once a sync target is selected is behind an interface which doesn't yet have an available implementation.


ajsutton:
Superseded by #1203

PR #1201, opened 2019-04-02
Privacy overview and transaction processing 

MadelineMurray:
Added Privacy overview and transaction processing


docs/Privacy/Privacy.md
```diff
@@ -1,92 +1,93 @@
...
+* How to configure Pantheon and the Private Transaction Manager (for example, Orion) to send private transactions
+* Creating and sending private transactions 
+
+## Private Transactions
+
+Private transactions are defined as *restricted* or *unrestricted*:  
+
+* In restricted private transactions the payload of the private transaction is stored only in the nodes
+participating in the transaction. 
+
```
antonydenyer: @rojotek can you confirm that with Orion only the participating nodes receive the payload. Might be worth calling out? 

 `In restricted private transactions, when using Orion, the payload of the private transaction is stored and only available to the participating nodes`

MadelineMurray: Updated to reflect that private transactions are only received by participating nodes

PR #1203, opened 2019-04-02
Pipeline chain download - Checkpoints

ajsutton:
## PR description
Implement an `Iterator` that retrieves checkpoint headers and creates ranges based on those to act as the start of the chain download pipeline.

Also adds the plumbing required to hook the pipeline download into Pantheon, but currently toggled off (it only gets as far as creating checkpoint ranges currently).

Builds on #1196.


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/CheckpointHeaderFetcher.java
```diff
@@ -0,0 +1,75 @@
...
+  private final ProtocolSchedule<?> protocolSchedule;
+  private final EthContext ethContext;
+  private final UnaryOperator<List<BlockHeader>> checkpointFilter;
+  private final MetricsSystem metricsSystem;
+
+  public CheckpointHeaderFetcher(
+      final SynchronizerConfiguration syncConfig,
+      final ProtocolSchedule<?> protocolSchedule,
+      final EthContext ethContext,
+      final UnaryOperator<List<BlockHeader>> checkpointFilter,
```
mbaxter: (optional) You could set this as a `Supplier<OptionalLong>` that provides the max block number to request, then we can avoid going over the limit for the last fast sync request.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/PipelineChainDownloader.java
```diff
@@ -0,0 +1,101 @@
...
+      final DownloadPipelineFactory downloadPipelineFactory,
+      final EthScheduler scheduler) {
+    this.syncTargetManager = syncTargetManager;
+    this.downloadPipelineFactory = downloadPipelineFactory;
+    this.scheduler = scheduler;
+  }
+
+  @Override
+  public CompletableFuture<Void> start() {
+    return performDownload();
```
mbaxter: You may want to use an `AtomicBoolean` to make sure start is only called once.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/PipelineChainDownloader.java
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  @Override
+  public synchronized void cancel() {
```
mbaxter: Why do you need these `synchronized` keywords?  If you only allow cancellation once, after starting (`if(started.get() && cancelled.compareAndSet(false, true)){...}`), do you need to synchronize?

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/CheckpointRangeSource.java
```diff
@@ -0,0 +1,116 @@
...
+  private int requestFailureCount = 0;
+
+  public CheckpointRangeSource(
+      final CheckpointHeaderFetcher checkpointFetcher,
+      final SyncTargetChecker syncTargetChecker,
+      final EthPeer peer,
+      final BlockHeader commonAncestor,
+      final int checkpointTimeoutsPermitted,
+      final int newHeaderWaitDuration,
+      final TimeUnit newHeaderWaitDurationUnit) {
```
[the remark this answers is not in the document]
ajsutton: True, we can't pass it on as a duration to the `get` call that needs it but no need to share that pain.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/PipelineChainDownloader.java
ajsutton: We need the synchronised keyword to protect access to the `currentDownloadPipeline` field and to ensure that when we're cancelling, there isn't another thread starting a different pipeline.  We need the creation, starting and assignment to `currentDownloadPipeline` to be a single atomic operation so we know `cancel` cancels the pipeline that's actually running.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/CheckpointHeaderFetcher.java
ajsutton: I'm going to leave this for now but have a note to revisit.  We need to ensure the last checkpoint header for fast sync is the pivot block and it sometimes needs to be explicitly inserted.  We could use an `Optional<BlockHeader> finalCheckpoint` and include all the logic inside `CheckpointHeaderFetcher` but then it winds up with a bunch of fast sync specific logic mixed in.  Need to play around with it but want to get something actually working first.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/PipelineChainDownloader.java
ajsutton: We do tend to follow that pattern.  Done.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/CheckpointRangeSource.java
```diff
@@ -0,0 +1,116 @@
...
+
+  @Override
+  public CheckpointRange next() {
+    if (!retrievedRanges.isEmpty()) {
+      return retrievedRanges.poll();
+    }
+    if (pendingCheckpointsRequest.isPresent()) {
+      return getCheckpointRangeFromPendingRequest();
+    }
+    pendingCheckpointsRequest =
```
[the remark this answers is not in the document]
ajsutton: Probably a good thing to do.

PR #1213, opened 2019-04-03
Add Unstable Options to the CLI

shemnon:
## PR description

Create a standard way for various modules to expose unstable options such as performance related options to the CLI but not have it exposed in the help menu.


ajsutton:
I generally like this approach, particularly the ability to provide options we don't guarantee will be around forever.  It would be nice if there was a way to get help for all of these performance related options (like `java -X` gives).

shemnon:
Here's what the `--X` help output looks like.

```
Performance options for RocksDBPerformanceOptions
      --Xrocksdb-max-open-files=<INTEGER>
         Max number of files RocksDB will open (default: 1024)

Performance options for SynchronizerPerformanceOptions
      --Xsynchronizer-computationParallelism=<INTEGER>
         Performance config for computationParallelism (default: # of processors)
      --Xsynchronizer-downloaderChainSegmentSize=<INTEGER>
         Performance config for downloaderChainSegmentSize (default: 200)
      --Xsynchronizer-downloaderChainSegmentTimeoutsPermitted=<INTEGER>
         Performance config for downloaderChainSegmentTimeoutsPermitted (default: 5)
      --Xsynchronizer-downloaderChangeTargetThresholdByHeight=<LONG>
         Performance config for downloaderChangeTargetThresholdByHeight (default: 20)
      --Xsynchronizer-downloaderCheckpointTimeoutsPermitted=<INTEGER>
         Performance config for downloaderCheckpointTimeoutsPermitted (default: 5)
      --Xsynchronizer-downloaderHeaderRequestSize=<INTEGER>
         Performance config for downloaderHeaderRequestSize (default: 200)
      --Xsynchronizer-downloaderParallelism=<INTEGER>
         Performance config for downloaderParallelism (default: 4)
      --Xsynchronizer-fastSyncFullValidationRate=<FLOAT>
         Performance config for fastSyncFullValidationRate (default: 0.1)
      --Xsynchronizer-fastSyncPivotDistance=<INTEGER>
         Performance config for fastSyncPivotDistance (default: 50)
      --Xsynchronizer-transactionsParallelism=<INTEGER>
         Performance config for transactionsParallelism (default: 2)
      --Xsynchronizer-worldStateHashCountPerRequest=<INTEGER>
         Performance config for worldStateHashCountPerRequest (default: 348)
      --Xsynchronizer-worldStateMaxRequestsWithoutProgress=<INTEGER>
         Performance config for worldStateMaxRequestsWithoutProgress (default: 1000)
      --Xsynchronizer-worldStateMinMillisBeforeStalling=<LONG>
         Performance config for worldStateMinMillisBeforeStalling (default: 5)
      --Xsynchronizer-worldStateRequestParallelism=<INTEGER>
         Performance config for worldStateRequestParallelism (default: 10)

```

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/SynchronizerPerformanceOptions.java
```diff
@@ -0,0 +1,174 @@
...
+ *
+ * <p>All of these options are also hidden so that they will not show up in the CLI help screen.
+ */
+public class SynchronizerPerformanceOptions {
+
+  private static final Supplier<SynchronizerPerformanceOptions> cachedInstance =
+      Suppliers.memoize(SynchronizerPerformanceOptions::new);
+
+  public static SynchronizerPerformanceOptions get() {
+    return cachedInstance.get();
```
ajsutton: Is there any way we can avoid using a singleton here? It's probably ok in the command line stuff since it's one-pantheon-per-vm but it's so much simpler to just have a blanket avoid singletons rule.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/SynchronizerConfiguration.java
```diff
@@ -216,24 +204,36 @@ public int getMaxTrailingPeers() {
...
-    private int downloaderParallelism = 4;
-    private int transactionsParallelism = 2;
-    private int computationParallelism = Runtime.getRuntime().availableProcessors();
-    private int fastSyncPivotDistance = DEFAULT_PIVOT_DISTANCE_FROM_HEAD;
-    private float fastSyncFullValidationRate = DEFAULT_FULL_VALIDATION_RATE;
-    private int fastSyncMinimumPeerCount = DEFAULT_FAST_SYNC_MINIMUM_PEERS;
-    private int worldStateHashCountPerRequest = DEFAULT_WORLD_STATE_HASH_COUNT_PER_REQUEST;
-    private int worldStateRequestParallelism = DEFAULT_WORLD_STATE_REQUEST_PARALLELISM;
+    private long downloaderChangeTargetThresholdByHeight =
+        SynchronizerPerformanceOptions.get().downloaderChangeTargetThresholdByHeight;
```
ajsutton: Using the singleton here probably is bad since `SynchronizerConfig.Builder` is used when creating multiple instances in the same VM.  We probably want it to be completely isolated from the command line parsing.  Probably going to create a bunch more boiler plate unfortunately. I think having a clear separation between command line and runtime code is probably worth it.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/SynchronizerPerformanceOptions.java
shemnon: Not easily in the general case.  For the synchronizer I can add this to the SynchronizerConfiguration.  For RocksDB... there is no good configuration to hang it off.  And then how much about the details of how pantheon works should the command shell know?

The best answer I see you may not like:  DI such as Guice or Dagger.  An external module could tell the DI module that it provides a set of configuration mixins.  These are singleton within the scope of the injector, so the RocksDB module can provide it and use it and the command shell just passes the values.  But even without Guide and Dagger we could do this in a generic fashion but we quickly produces the services that Guice and Dagger are already providing.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/SynchronizerConfiguration.java
shemnon: This may be the correct time to look at DI again.

ajsutton: Maybe, but if you can do it with DI you can do it without - it's just a question of how many places you have to pass the value through.  Showing the two different approaches in this particular situation is probably a good way to evaluate DI...

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/SynchronizerPerformanceOptions.java
ajsutton: In terms of how much the command shell should know about how pantheon works, I think its one job is to setup the configuration for Pantheon.  So anything we expose on the command line should be also available as a config option when you use "embedded" Pantheon via `RunnerBuilder`.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/SynchronizerConfiguration.java
```diff
@@ -216,23 +217,143 @@ public int getMaxTrailingPeers() {
   public static class Builder {
     private SyncMode syncMode = SyncMode.FULL;
     private Range<Long> blockPropagationRange = Range.closed(-10L, 30L);
+
+    @CommandLine.Option(
+        names = "--Xsynchronizer-downloader-change-target-threshold-by-height",
+        hidden = true,
+        defaultValue = "20",
```
ajsutton: I take it we wound up having to have the defaults duplicated then?  Should we raise an issue for PicoCLI so hopefully we can have them magically picked up at some point in the future?

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/SynchronizerConfiguration.java
```diff
@@ -216,23 +217,143 @@ public int getMaxTrailingPeers() {
...
     private long downloaderChangeTargetThresholdByHeight = 20L;
+
+    //  @CommandLine.Option(
+    //      names = "--Xsynchronizer-downloader-change-target-threshold-by-td",
+    //      hidden = true,
+    //      defaultValue = "1000000000",
+    //      paramLabel = "<UINT256>",
+    //      description =
+    //          "Minimum total difficulty difference before switching fast sync download peers
+    // (default: ${DEFAULT-VALUE})")
```
ajsutton: nit: Probably should just remove the annotation instead of commenting it out.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/SynchronizerConfiguration.java
```diff
@@ -216,23 +217,143 @@ public int getMaxTrailingPeers() {
...
+            "Fast sync world state hashes queried per request (default: ${DEFAULT-VALUE})")
     private int worldStateHashCountPerRequest = DEFAULT_WORLD_STATE_HASH_COUNT_PER_REQUEST;
+
+    @CommandLine.Option(
+        names = "--Xsynchronizer-world-state-request-parallelism",
+        hidden = true,
+        defaultValue = "10",
+        paramLabel = "<INTEGER>",
+        description =
+            "Number of threads committed to fast sync world state downloading (default: ${DEFAULT-VALUE})")
```
ajsutton:
```suggestion
            "Number of concurrent requests to use when downloading fast sync world state (default: ${DEFAULT-VALUE})")
```

There's only ever one thread used, just multiple concurrent requests.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/SynchronizerConfiguration.java
```diff
@@ -216,23 +217,143 @@ public int getMaxTrailingPeers() {
...
     private int worldStateRequestParallelism = DEFAULT_WORLD_STATE_REQUEST_PARALLELISM;
+
+    @CommandLine.Option(
+        names = "--Xsynchronizer-world-state-max-requests-without-progress",
+        hidden = true,
+        defaultValue = "1000",
+        paramLabel = "<INTEGER>",
+        description =
+            "Number of world state requests accepted without progress before stopping (default: ${DEFAULT-VALUE})")
```
ajsutton:
```suggestion
            "Number of world state requests accepted without progress before considering the download stalled (default: ${DEFAULT-VALUE})")
```

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/SynchronizerConfiguration.java
```diff
@@ -216,23 +217,143 @@ public int getMaxTrailingPeers() {
...
     private int worldStateMaxRequestsWithoutProgress =
         DEFAULT_WORLD_STATE_MAX_REQUESTS_WITHOUT_PROGRESS;
+
+    @CommandLine.Option(
+        names = "--Xsynchronizer-world-state-min-millis-before-stalling",
+        hidden = true,
+        defaultValue = "5",
+        paramLabel = "<LONG>",
+        description =
+            "Minimum time in ms to permit world state queries to stall (default: ${DEFAULT-VALUE})")
```
ajsutton:
```suggestion
            "Minimum time in ms without progress before considering a world state download as stalled (default: ${DEFAULT-VALUE})")
```

pantheon/src/main/java/tech/pegasys/pantheon/cli/UnstableOptionsSubCommand.java
```diff
@@ -0,0 +1,92 @@
...
+    if ("mixinStandardHelpOptions".equals(mixinName)) return;
+
+    // Recreate the options but flip hidden to false.
+    final CommandLine.Model.CommandSpec cs = CommandLine.Model.CommandSpec.create();
+    commandSpec.options().stream()
+        .filter(option -> option.hidden() && option.names()[0].startsWith("--X"))
+        .forEach(option -> cs.addOption(option.toBuilder().hidden(false).build()));
+
+    // Print out the help text.
+    out.printf("Unstable options for %s%n", mixinName);
```
ajsutton: The help output includes `Unstable options for standaloneCommands` at the top so we probably want to exclude that module as well (or probably even better, exclude all modules that have no unstable options).

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/SynchronizerConfiguration.java
abdelhamidbakhta: +1 for the DI evaluation

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/SynchronizerConfiguration.java
shemnon: I didn't look too deeply.  There is support for a defaults provider but getting it linked to the fields that were annotated is a good bit of work.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/SynchronizerConfiguration.java
shemnon: I added support for UInt256 and restored this annotation.

pantheon/src/main/java/tech/pegasys/pantheon/cli/UnstableOptionsSubCommand.java
shemnon: Fixed this with a change of approach: don't list mixins with zero relevant options.  That way I don't have to enumerate help or standalone commands.

PR #1215, opened 2019-04-04
Support async processing while maintaining output order

ajsutton:
## PR description
Adds a `thenProcessAsyncOrdered` function to `PipelineBuilder` so that operations can be processed asynchronously, with multiple operations happening concurrently, but ensure that outputs are in the same order as inputs.

Also fixes a potential performance issue caused by using `CompletableFuture.anyOf` too much. Each call adds completion actions to the futures involved which are never removed, so when they eventually complete they have a huge stack of actions to notify which can take significant time.  Caching the result keeps the number of actions to a reasonable level.

Also fixes an issue where the pipeline was not aborted if an asynchronous operation completed exceptionally.

## Fixed Issue(s)
https://pegasys1.atlassian.net/browse/PAN-2527


services/pipeline/src/main/java/tech/pegasys/pantheon/services/pipeline/AsyncOperationProcessor.java
```diff
@@ -69,26 +77,48 @@ private void outputNextCompletedTask(final WritePipe<O> outputPipe) {
...
+        break;
       }
     }
+    if (inProgressChanged) {
+      updateNextOutputAvailableFuture();
+    }
+  }
+
+  /**
+   * CompletableFuture.anyOf adds a completion handler to every future its passed so if we call it
```
mbaxter: Does it help to cancel the future returned from anyOf when you overwrite it??

ajsutton: Sadly no, it looks like it only cleans the completion stack when the underlying futures complete.

PR #1218, opened 2019-04-04
[PAN-2265] Expose fast-sync options on command line

abdelhamidbakhta:
## PR description
- add `--fast-sync-min-peers` hidden cli option

- add `--fast-sync-max-wait-time` hidden cli option
- update `FastSyncActions` to handle no timeout behaviour (when timeout set to 0 or no timeout set then we just trigger asynchrounously the waitPeersTask)
- update unit tests mock related to sync config builder
- add fast sync options in `everything_config.toml`
- re enable `PantheonCommandTest.syncModeOptionMustBeUsed`

## Fixed Issue(s)

- PAN-2265


pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -221,13 +223,28 @@ void setBootnodes(final List<String> values) {
...
   @Option(
-      hidden = true,
       names = {"--sync-mode"},
       paramLabel = MANDATORY_MODE_FORMAT_HELP,
       description =
           "Synchronization mode, possible values are ${COMPLETION-CANDIDATES} (default: ${DEFAULT-VALUE})")
   private final SyncMode syncMode = DEFAULT_SYNC_MODE;
 
+  @Option(
+      hidden = true,
```
NicolasMassart: Tip: try displaying the help with hidden=false first, then make it true.

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -221,13 +223,28 @@ void setBootnodes(final List<String> values) {
...
+      names = {"--fast-sync-min-peers"},
+      paramLabel = MANDATORY_INTEGER_FORMAT_HELP,
+      description =
+          "Minimum number of peers before starting fast sync, possible values are ${COMPLETION-CANDIDATES} (default: ${DEFAULT-VALUE})")
```
NicolasMassart: Completion candidates placeholder is not relevant when the value is integer. You can't list all possible integers and then the placeholder will be replaced by empty string in the final description (visible when option hidden=false for testing). If needed, a range would be better.

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -221,13 +223,28 @@ void setBootnodes(final List<String> values) {
...
+  private final Integer fastSyncMinPeerCount = DEFAULT_FAST_SYNC_MIN_PEER_COUNT;
+
+  @Option(
+      hidden = true,
+      names = {"--fast-sync-max-wait-time"},
+      paramLabel = MANDATORY_INTEGER_FORMAT_HELP,
+      description =
+          "Maximum time to wait the required number of peers before starting fast sync, expressed in seconds, 0 means no timeout, possible values are ${COMPLETION-CANDIDATES} (default: ${DEFAULT-VALUE})")
```
NicolasMassart: Same as previous option for the completion candidates.

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -590,25 +607,28 @@ public void run() {
     }
 
     // Check that P2P options are able to work or send an error
-    CommandLineUtils.checkOptionDependencies(
+    checkOptionDependencies(
```
NicolasMassart: 👍 for the static imports

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -658,6 +678,19 @@ public void run() {
     }
   }
 
+  private void checkFastSyncOptions() throws ParameterException {
+    if (fastSyncMaxWaitTime < 0) {
+      throw new ParameterException(
+          commandLine, "--fast-sync-max-wait-time must be a positive or zero integer");
```
NicolasMassart: this range should be explained in the option description too. Otherwise we are going to have issues documenting CLI if we have to read each error case.

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -658,6 +678,19 @@ public void run() {
...
+    }
+    checkOptionDependencies(
+        logger,
+        commandLine,
+        "--sync-mode",
+        syncMode.equals(SyncMode.FAST),
```
NicolasMassart: I personally prefer the reverse pattern to prevent NPEs, even if here I don't see a way for it to be null, but imagine PicoCLI has something strange and injects it null in some cases. And I think this is a good habit to do it all the time in case of.
```suggestion
        SyncMode.FAST.equals(syncMode),
```

pantheon/src/test/java/tech/pegasys/pantheon/cli/PantheonCommandTest.java
```diff
@@ -1035,7 +1034,6 @@ public void maxpeersOptionMustBeUsed() {
     assertThat(commandErrorOutput.toString()).isEmpty();
   }
 
-  @Ignore("Ignored as we only have one mode available for now. See NC-1057/NC-1681")
   @Test
   public void syncModeOptionMustBeUsed() {
```
NicolasMassart: Also 2 new tests similar to `syncModeOptionMustBeUsed` to check the two new options values are well passed from the CLI to the SynchronizerConfiguration would help make sure we don't have an issue with a valid integer passed but not matching the entered value on the command line.

abdelhamidbakhta: tests added

pantheon/src/test/java/tech/pegasys/pantheon/cli/PantheonCommandTest.java
```diff
@@ -1047,6 +1048,27 @@ public void syncModeOptionMustBeUsed() {
     assertThat(commandErrorOutput.toString()).isEmpty();
   }
 
+
+  @Test
+  public void fastSyncTimeoutOptionMustBeUsed() {
+
+    parseCommand("--sync-mode", "FAST", "--fast-sync-max-wait-time", "30");
```
NicolasMassart: You may try with values that are not the default values to be sure that they are taken in account and that it's not just the default.

pantheon/src/test/java/tech/pegasys/pantheon/cli/PantheonCommandTest.java
```diff
@@ -1047,6 +1048,27 @@ public void syncModeOptionMustBeUsed() {
...
+    verify(mockSyncConfBuilder).syncMode(ArgumentMatchers.eq(SyncMode.FAST));
+    verify(mockSyncConfBuilder).fastSyncMaximumPeerWaitTime(ArgumentMatchers.eq(Duration.ofSeconds(30)));
+    assertThat(commandOutput.toString()).isEmpty();
+    assertThat(commandErrorOutput.toString()).isEmpty();
+  }
+
+  @Test
+  public void fastSyncMinPeersOptionMustBeUsed() {
+
+    parseCommand("--sync-mode", "FAST", "--fast-sync-min-peers", "5");
```
NicolasMassart: same as for the other option

pantheon/src/test/java/tech/pegasys/pantheon/cli/PantheonCommandTest.java
abdelhamidbakhta: done

pantheon/src/test/java/tech/pegasys/pantheon/cli/PantheonCommandTest.java
abdelhamidbakhta: done

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -14,6 +14,9 @@
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.time.Duration.ofSeconds;
```
shemnon: We've been inconsistent with this, but our preference has been to not import every static method.  In fact there are some ErrorProne inspections that will report as errors if we do . (*.Builder and *.of in particular).  So I think we should shy away from most method imports.

These particular imports causes a lot of changes in lines that are note relevant to the PR at hand.

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -658,6 +678,19 @@ public void run() {
     }
   }
 
+  private void checkFastSyncOptions() throws ParameterException {
+    if (fastSyncMaxWaitTime < 0) {
```
shemnon: Our parameter validation has been more just in time and directly in the configuration options rather than directly between the CLI and the config.  It keeps the validations closer to where the parameters are actually stored.  Validating in the CLI also provides a barrier to splitting the code apart.

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -658,6 +678,19 @@ public void run() {
...
+      throw new ParameterException(
+          commandLine, "--fast-sync-max-wait-time must be a positive or zero integer");
+    }
+    checkOptionDependencies(
```
shemnon: This check option dependencies should be moved to be with the others rather than in its own method.

pantheon/src/test/java/tech/pegasys/pantheon/cli/PantheonCommandTest.java
```diff
@@ -1049,6 +1048,27 @@ public void syncModeOptionMustBeUsed() {
     assertThat(commandErrorOutput.toString()).isEmpty();
   }
 
+  @Test
```
shemnon: :+1:  More tests :smile:

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -221,13 +223,28 @@ void setBootnodes(final List<String> values) {
   private final Collection<String> bannedNodeIds = new ArrayList<>();
 
   @Option(
-      hidden = true,
```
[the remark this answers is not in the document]
abdelhamidbakhta: Done.

pantheon/src/test/java/tech/pegasys/pantheon/cli/PantheonCommandTest.java
```diff
@@ -1035,7 +1035,6 @@ public void maxpeersOptionMustBeUsed() {
     assertThat(commandErrorOutput.toString()).isEmpty();
   }
 
-  @Ignore("Ignored as we only have one mode available for now. See NC-1057/NC-1681")
```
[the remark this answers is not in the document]
abdelhamidbakhta: Ok. Let's keep this for the followup ticket

pantheon/src/test/java/tech/pegasys/pantheon/cli/PantheonCommandTest.java
```diff
@@ -1049,6 +1048,27 @@ public void syncModeOptionMustBeUsed() {
     assertThat(commandErrorOutput.toString()).isEmpty();
   }
 
+
+  @Test
+  public void fastSyncTimeoutOptionMustBeUsed() {
```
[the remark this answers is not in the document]
abdelhamidbakhta: Done.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/fastsync/FastSyncActions.java
```diff
@@ -71,8 +72,24 @@ public FastSyncActions(
             ethContext, syncConfig.getFastSyncMinimumPeerCount(), metricsSystem);
 
     final EthScheduler scheduler = ethContext.getScheduler();
+    final CompletableFuture<Void> fastSyncTask;
+    if (!syncConfig.getFastSyncMaximumPeerWaitTime().isZero()) {
+      LOG.info(
```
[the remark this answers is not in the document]
abdelhamidbakhta: Done. Switch to Debug

pantheon/src/main/java/tech/pegasys/pantheon/cli/DefaultCommandValues.java
```diff
@@ -55,6 +55,8 @@
   // Default should be FAST for the next release
   // but we use FULL for the moment as Fast is still in progress
   SyncMode DEFAULT_SYNC_MODE = SyncMode.FULL;
+  int FAST_SYNC_MAX_WAIT_TIME = 0;
+  int FAST_SYNC_MIN_PEER_COUNT = 5;
```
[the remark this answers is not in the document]
abdelhamidbakhta: Done.

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
abdelhamidbakhta: Ok. I was just thinking that the other method will become bigger and bigger.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/fastsync/FastSyncActions.java
```diff
@@ -71,8 +72,24 @@ public FastSyncActions(
...
+          format(
+              "Waiting for at least %d peers, maximum wait time set to %s.",
+              syncConfig.getFastSyncMinimumPeerCount(),
+              syncConfig.getFastSyncMaximumPeerWaitTime().toString()));
+      fastSyncTask =
+          scheduler.timeout(waitForPeersTask, syncConfig.getFastSyncMaximumPeerWaitTime());
+    } else {
+      LOG.info(
+          format(
```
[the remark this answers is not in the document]
abdelhamidbakhta: ok

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/fastsync/FastSyncActions.java
```diff
@@ -71,8 +72,24 @@ public FastSyncActions(
             ethContext, syncConfig.getFastSyncMinimumPeerCount(), metricsSystem);
 
     final EthScheduler scheduler = ethContext.getScheduler();
+    final CompletableFuture<Void> fastSyncTask;
+    if (!syncConfig.getFastSyncMaximumPeerWaitTime().isZero()) {
+      LOG.info(
+          format(
```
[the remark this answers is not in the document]
abdelhamidbakhta: ok

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
abdelhamidbakhta: ok

pantheon/src/test/java/tech/pegasys/pantheon/cli/PantheonCommandTest.java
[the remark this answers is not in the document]
mbaxter: I think you still need to update those tests for these new flags though

pantheon/src/test/java/tech/pegasys/pantheon/cli/PantheonCommandTest.java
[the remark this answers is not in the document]
mbaxter: Still need to add tests for the failure cases:

> We should also add tests with invalid options and check that parsing fails.

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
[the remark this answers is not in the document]
mbaxter: Looks like you flipped the [wrong flag](https://github.com/PegaSysEng/pantheon/pull/1218/files#diff-3ea9c74bd15a6eed2f3b581e6b609544R233) :)

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
shemnon: It will.  When we do something different we should be consistent and do it differently for all, not just new arriving code. But fixing the size problem is out of scope for this PR.

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
NicolasMassart: A custom converter to restrict the integer to a positive non null one ?

abdelhamidbakhta: None of other integers have restriction. For example metrics interval in seconds are not checked too. So what is the best practice to perform those checks ?

pantheon/src/test/java/tech/pegasys/pantheon/cli/PantheonCommandTest.java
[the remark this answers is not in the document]
abdelhamidbakhta: Failure cases added.

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
abdelhamidbakhta: I suggest we keep like this for this PR. We need to define a way to do this kind of checks and implement it for all options in a separate ticket imo.

pantheon/src/test/java/tech/pegasys/pantheon/cli/PantheonCommandTest.java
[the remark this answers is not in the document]
abdelhamidbakhta: I have updated those tests for these new flags

PR #1224, opened 2019-04-05
[PIE-1224] Different request limits for different request types

abdelhamidbakhta:
## PR description
- add fields in `SynchronizerConfiguration` with different request limits according to geth (https://github.com/ethereum/go-ethereum/blob/master/eth/downloader/downloader.go)
- update `EthServer` to use those limits instead of the default one
- remove requestLimit from `EthServer` and `EthProtocolManager` constructors
## Fixed Issue(s)
fixes PIE-1224


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/SynchronizerConfiguration.java
```diff
@@ -36,6 +36,10 @@
   private static final int DEFAULT_WORLD_STATE_MAX_REQUESTS_WITHOUT_PROGRESS = 1000;
   private static final long DEFAULT_WORLD_STATE_MIN_MILLIS_BEFORE_STALLING =
       TimeUnit.MINUTES.toMillis(5);
+  private static final int DEFAULT_MAX_GET_BLOCK_HEADERS = 192;
+  private static final int DEFAULT_MAX_GET_BLOCK_BODIES = 128;
+  private static final int DEFAULT_MAX_GET_RECEIPTS = 256;
+  private static final int DEFAULT_MAX_GET_NODE_DATA = 384;
```
mbaxter: These constants aren't really related to the synchronizer, I'd move them to `EthServer` or `EthProtocolManager`

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/EthServer.java
```diff
@@ -47,20 +47,30 @@
...
   private final Blockchain blockchain;
   private final WorldStateArchive worldStateArchive;
   private final EthMessages ethMessages;
-  private final int requestLimit;
+  private final int maxGetBlockHeaders;
+  private final int maxGetBlockBodies;
+  private final int maxGetReceipts;
+  private final int maxGetNodeData;
 
   EthServer(
```
mbaxter: How about adding another constructor that uses the default request size values, so we don't need to add all of these new constructor parameters at higher levels?

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/manager/EthServerTest.java
```diff
@@ -43,13 +42,16 @@
...
   private final EthMessages ethMessages = new EthMessages();
 
-  @Before
-  public void setUp() {
-    new EthServer(blockchain, worldStateArchive, ethMessages, 2);
-  }
-
   @Test
   public void shouldRespondToNodeDataRequests() throws Exception {
+    new EthServer(
```
mbaxter: Why not just set up a single instance in the `setUp`?

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/manager/EthServerTest.java
```diff
@@ -68,6 +78,14 @@ public void shouldHandleDataBeingUnavailableWhenRespondingToNodeDataRequests() t
...
   @Test
   public void shouldLimitNumberOfResponsesToNodeDataRequests() throws Exception {
+    new EthServer(
+        blockchain,
+        worldStateArchive,
+        ethMessages,
+        EthServer.DEFAULT_REQUEST_LIMIT,
+        EthServer.DEFAULT_REQUEST_LIMIT,
+        EthServer.DEFAULT_REQUEST_LIMIT,
+        EthServer.DEFAULT_REQUEST_LIMIT);
```
mbaxter: Not sure how this is testing is passing with the default limit at 200.  The intention of this test is to check that the server caps the amount of data returned.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/manager/EthServerTest.java
abdelhamidbakhta: Can't test differents limits with single instance. Some tests would fail

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/SynchronizerConfiguration.java
abdelhamidbakhta: fixed

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/EthServer.java
abdelhamidbakhta: fixed

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/EthServer.java
```diff
@@ -47,23 +47,53 @@
 class EthServer {
   private static final Logger LOG = LogManager.getLogger();
 
+  static final int DEFAULT_REQUEST_LIMIT = 200;
```
mbaxter: I'd cut this since its only getting used in one test util - just set the values manually there.

abdelhamidbakhta: fixed

abdelhamidbakhta:
- enable differents request limits on Ethereum Wire Protocol. Reference values took from Geth
- update tests accordingly

PR #1241, opened 2019-04-09
Added fast sync options

MadelineMurray:
Added fast sync options (under note about them being experimental)


docs/Reference/Pantheon-CLI-Syntax.md
```diff
@@ -1059,6 +1059,60 @@ Default is `INFO`.
...
+--sync-mode=FAST
+```
+
+```bash tab="Example Configuration File"
+sync-mode="FAST"
+```
+
+Specifies the synchronization mode. Default is `FULL`.
+
+### fast-sync-max-wait-time
```
mbaxter: As I understand it, this flag is for debugging - do we need to document debugging flags?

PR #1243, opened 2019-04-09
Added note about using node permissioning only

MadelineMurray:
Added note about using node permissioning only


docs/Permissions/Permissioning-Overview.md
```diff
@@ -7,6 +7,11 @@ A permissioned network is a network where only specified nodes and accounts (par
 Nodes and accounts outside those specified are prevented from participating. Permissioned networks can have node permissioning enabled, 
 account permissioning enabled, or both. 
 
+!!! note
+    Implementing node-level permissions only is not a trustless solution. Each node
+    has no control or view of who other nodes in the network are communicating with. With either off-chain or 
+    on-chain permissions, a bad actor could violate the network governance agreed on, and act as a proxy to other nodes.
+
```
arash009: How about:
"In a peer-to-peer network, Node-level permissions can be used to enforce rules on nodes that you control. It is still possible that a bad actor could violate the network governance, and act as a proxy to other nodes.

MadelineMurray: Updated.

PR #1251, opened 2019-04-10
[MINOR] migrate TransactionPool (& affiliated test) from 'core' to 'eth'

s-matthew-english:
## PR description

Necessary pre-condition of using `SyncState` in `TransactionPool` as a filtering mechanism. 

The gotcha was adding these lines: 

```
testImplementation project(path: ':config', configuration: 'testSupportArtifacts')
integrationTestImplementation project(path: ':config', configuration: 'testSupportArtifacts')
```

to `ethereum/eth/build.gradle`, since `TransactionPoolTest` has `ExecutionContextTestFixture` which requires `GenesisConfigFile`, or anyway one of those files in `tech.pegasys.pantheon.config`.


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/PendingTransactions.java
```diff
@@ -96,7 +96,7 @@ public boolean addRemoteTransaction(final Transaction transaction) {
     return addTransaction;
   }
 
-  boolean addLocalTransaction(final Transaction transaction) {
+  public boolean addLocalTransaction(final Transaction transaction) {
```
mbaxter: You should move all of the `TransactionPool` related classes to `eth`.  I believe that includes: `TransactionPool`, `PendingTransactions`, `PendingTransactionListener`, `PendingTransactionDroppedListener`.

ethereum/eth/build.gradle
```diff
@@ -55,4 +55,7 @@ dependencies {
   testImplementation 'org.mockito:mockito-core'
 
   jmhImplementation project(path: ':ethereum:core', configuration: 'testSupportArtifacts')
+
+  testImplementation project(path: ':config', configuration: 'testSupportArtifacts')
```
shemnon: This like should be alphabetized with the rest of the testImplementation dependenceies.  Between line 45 and 46

s-matthew-english: Nice catch -- thank you.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/PendingTransactions.java
s-matthew-english: Yeah -- I was initially wondering about whether or not to do that -- I've updated the branch accordingly

PR #1246, opened 2019-04-09
Feature/pie 1502 ewp options cli

abdelhamidbakhta:
## PR description
Expose Ethereum Wire Protocol request limits as hidden CLI options.

**Acceptance Criteria**

The following options are exposed as hidden CLI options
- --ewp-max-get-headers
- --ewp-max-get-bodies
- --ewp-max-get-receipts
- --ewp-max-get-node-data
## Fixed Issue(s)

fixes PIE-1502

PR #1255, opened 2019-04-10
Feature/pie 1502 ewp options cli

abdelhamidbakhta:
## PR description
Follow UP PR 
https://github.com/PegaSysEng/pantheon/pull/1246
Fix discussion https://github.com/PegaSysEng/pantheon/pull/1246#discussion_r273931560


ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/manager/EthServerTest.java
```diff
@@ -45,7 +46,11 @@
 
   @Before
   public void setUp() {
-    new EthServer(blockchain, worldStateArchive, ethMessages, 2, 2, 2, 2);
+    new EthServer(
+        blockchain,
+        worldStateArchive,
+        ethMessages,
+        new EthereumWireProtocolConfiguration(2, 2, 2, 2));
```
mbaxter: Can we make this constructor private so that modifications must go through the builder?  If we stick with the builder pattern, it will save us pain in the future.

PR #1257, opened 2019-04-10
[PAN-2433] Fix netty pipeline

mbaxter:
## PR description
This PR fixes a few issues where the `DeFramer` channel handler would forward messages down the netty pipeline before downstream channel handlers were set up.   

Previously, the `DeFramer` would only handle the initial `Hello` message, forwarding everything else down the pipeline.  It now checks for `Disconnect` messages and other invalid messages received before downstream handlers are set up.  

Additionally, we no longer return early from this handler in the case where a disconnect is sent.  There's some buffer time between sending a disconnect message and actually terminating a connection.  In that window of time, we could receive other messages from the peer which need to be handled.


shemnon:
+1, tests look good.

PR #1259, opened 2019-04-11
[MINOR / PAN-1339 (A)] Fetch local transactions in isolation

s-matthew-english:
## PR description

In an effort to make the change set inevitably necessitated by the acceptance criteria of `PAN-1339` less onerous, and consequently the review more palatable, I've decided to serialize it into constituent elements -- each of which constituting a relatively discrete semantic chunk.  

## Fixed Issue(s)

In order to disseminate local transactions to new peers we'll require a mechanism to fetch local transactions in isolation -- this is that mechanism (_along with associated test infrastructure_).

### DIFF
--- a/ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/transactions/PendingTransactions.java
+++ b/ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/transactions/PendingTransactions.java
@@ -92,6 +92,18 @@ public class PendingTransactions {
             "operation");
   }
 
+  List<Transaction> getLocalTransactions() {
+    synchronized (pendingTransactions) {
+      List<Transaction> localTransactions = new ArrayList<>();
+      for (Map.Entry<Hash, TransactionInfo> transaction : pendingTransactions.entrySet()) {
+        if (transaction.getValue().isReceivedFromLocalSource()) {
+          localTransactions.add(transaction.getValue().getTransaction());
+        }
+      }
+      return localTransactions;
+    }
+  }
+
   public boolean addRemoteTransaction(final Transaction transaction) {
     final TransactionInfo transactionInfo =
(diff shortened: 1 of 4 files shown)


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/PendingTransactions.java
```diff
@@ -88,6 +88,18 @@ public PendingTransactions(
             "operation");
   }
 
+  List<Transaction> getLocalTransactions() {
```
abdelhamidbakhta: Maybe we could do it using parallel stream to optimize.
```suggestion
  List<Transaction> getLocalTransactions() {
    synchronized (pendingTransactions) {
      List<Transaction> localTransactions = new ArrayList<>();
      pendingTransactions
          .entrySet()
          .parallelStream()
          .filter(transaction -> transaction.getValue().isReceivedFromLocalSource())
          .forEach(transaction -> transaction.getValue().getTransaction());
      return localTransactions;
    }
  }
```

abdelhamidbakhta: Suggestion tip : cache the filtered list and store a boolean indicating whether the list is still consistent or not. Set it to false when a new transaction is received. We could have a method `invalidateCachedLocalTransactions()` to set this boolean to false when appropriate.

abdelhamidbakhta: Class level field to avoid instantiating `localTransactions` each time `localTransactions` is called ?

ajsutton: I'd avoid any such caching until we know we have a performance issue.  We're currently having issues with memory usage so don't want to waste it on extra caches if they aren't necessary.

PR #1260, opened 2019-04-11
Explicitly stop synchronizer

abdelhamidbakhta:
## PR description
When Pantheon shuts down, it stops the network, including the EthScheduler, and then closes the `PantheonController`. The `Synchronizer` never gets an explicit message to stop.
- add `stop` method in `Synchronizer` interface
- implement `stop` method in `DefaultSynchronizer`
- implement `cancel` method in `FastSyncDownloader`
- implement `cancel` method in `FullSyncDownloader`
## Fixed Issue(s)

fixes #PAN-2535


pantheon/src/main/java/tech/pegasys/pantheon/controller/IbftLegacyPantheonController.java
```diff
@@ -176,6 +176,11 @@ private IbftLegacyPantheonController(
           } catch (final IOException e) {
             LOG.error("Failed to close storage provider", e);
           }
+          try {
```
shemnon: We should stop the synchronizer prior to the storage.  The shutdown process could try to write more data.

pantheon/src/main/java/tech/pegasys/pantheon/controller/IbftPantheonController.java
```diff
@@ -318,6 +318,11 @@ private IbftPantheonController(
           } catch (final IOException e) {
             LOG.error("Failed to close storage provider", e);
           }
+          try {
```
shemnon: We should stop the synchronizer prior to the storage.  The shutdown process could try to write more data.

pantheon/src/main/java/tech/pegasys/pantheon/controller/MainnetPantheonController.java
```diff
@@ -206,6 +206,12 @@ private MainnetPantheonController(
           } catch (final IOException e) {
             LOG.error("Failed to close storage provider", e);
           }
+
```
shemnon: We should stop the synchronizer prior to the storage.  The shutdown process could try to write more data.

pantheon/src/main/java/tech/pegasys/pantheon/controller/CliquePantheonController.java
```diff
@@ -235,6 +235,11 @@ private CliquePantheonController(
           } catch (final IOException e) {
             LOG.error("Failed to close storage provider", e);
           }
+          try {
```
shemnon: We should stop the synchronizer prior to the storage.  The shutdown process could try to write more data.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/PipelineChainDownloader.java
```diff
@@ -113,12 +113,6 @@ public synchronized void cancel() {
   private CompletionStage<Void> handleFailedDownload(final Throwable error) {
     pipelineErrorCounter.inc();
     final Throwable rootCause = ExceptionUtils.rootCause(error);
-    if (!cancelled.get() && rootCause instanceof CancellationException) {
```
shemnon: Is the synchronizer shutdown what is causing this failure?  If so is there a reliable test case that can induce this failure?

shemnon: Also, it looks like there are tests that depend on this being present: https://jenkins.pegasys.tech/job/Pantheon/job/PR-1260/2/testReport/junit/tech.pegasys.pantheon.ethereum.eth.sync/PipelineChainDownloaderTest/UnitTests___shouldStopWithoutCompletingFutureWhenTargetSelectionUnexpectedCancelled/

abdelhamidbakhta: Yes. What should we do then ?

shemnon: after reading the PR it looks like we should delete the tests.  However given that this is to ensure proper resume I think we need to spin up an AWS instance and run a fast sync, stop it a few hours in, and then see if it resumes and finishes.  We could do this against ropsten but I would prefer to see a mainnset success before we commit.

abdelhamidbakhta: Ok. let's do this

ajsutton: I only added the test in question around the same time I added this ticket.  We should check that we can start a fast sync - wait for it to begin downloading the chain, kill pantheon and check that when it restarts the fast sync resumes.  If that works it's all good.

Testing locally, that doesn't appear to be working.  When restarting we get:
```
2019-04-15 10:51:02.965+10:00 | main | INFO  | FastDownloaderFactory | Fast sync was requested, but cannot be enabled because the local blockchain is not empty.
```

To test I ran pantheon with `./build/install/pantheon/bin/pantheon --network goerli --fast-sync-min-peers=1 --sync-mode=FAST` (after a `./gradlew installDist`)

abdelhamidbakhta:
replaced by https://github.com/PegaSysEng/pantheon/pull/1302

PR #1262, opened 2019-04-11
PAN-2404 - Make PeerDiscovertAgentTest less flakey.

shemnon:
## PR description

Because of deep issues sometimes we get duplicate Peers that differ
in small ways.  Tweak the test to verify we get all of the peers
we are expecting and to be tolerant of quasi-duplicate peers.

## Fixed Issue(s)

PAN-2404


ethereum/p2p/src/test/java/tech/pegasys/pantheon/ethereum/p2p/discovery/PeerDiscoveryAgentTest.java
```diff
@@ -73,7 +73,7 @@ public void neighborsPacketLimited() {
 
     // Start another peer pointing to those 20 agents.
     final MockPeerDiscoveryAgent agent = helper.startDiscoveryAgent(otherPeers);
-    assertThat(agent.getPeers()).hasSize(20);
+    assertThat(agent.getPeers()).contains(otherPeers.toArray(new DiscoveryPeer[20]));
```
mbaxter: Might be good to add a comment here related to this change.

shemnon: Done

PR #1264, opened 2019-04-11
[PAN-2560] Normalize EnodeURLs

mbaxter:
## PR description
Normalize `EnodeURL`'s so that values representing the same enode will have equivalent `EnodeURL` representations.  The [spec](https://github.com/ethereum/wiki/wiki/enode-url-format) specifies that the `discport` query param should only be present when the discovery port differs from the listening port.  This PR centralizes `EnodeURL` construction so that this condition can be enforced.


util/src/main/java/tech/pegasys/pantheon/util/enode/EnodeURL.java
```diff
@@ -71,10 +73,15 @@ public EnodeURL(
     }
     this.ip = address;
     this.listeningPort = listeningPort;
-    this.discoveryPort = discoveryPort;
+    // Only explicitly define a discovery port if it differs from the listening port
+    if (discoveryPort.isPresent() && discoveryPort.getAsInt() != listeningPort) {
```
mbaxter: This is where normalization happens.

PR #1271, opened 2019-04-12
Implement max message size rather than limiting with fixed number of transactions

abdelhamidbakhta:
## PR description

Adding transactions to the RLP until the message size exceeds the limit and then send that.

## Fixed Issue(s)

fixes #PAN-1005


ethereum/rlp/src/main/java/tech/pegasys/pantheon/ethereum/rlp/AbstractRLPOutput.java
```diff
@@ -141,6 +141,10 @@ public int encodedSize() {
     return payloadSizes[0];
   }
 
+  public int uncheckedEncodedSize() {
```
mbaxter: Would be helpful to add some javadoc here.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/messages/TransactionsMessage.java
```diff
@@ -47,7 +49,24 @@ public static TransactionsMessage create(final Iterable<Transaction> transaction
...
-  private TransactionsMessage(final BytesValue data) {
+  public static LimitedTransactionsMessages createLimited(
+      final Iterable<Transaction> transactions, final TransactionsMessageLimiter limiter) {
+    final Set<Transaction> includedTransactions = new HashSet<>();
+    final BytesValueRLPOutput tmp = new BytesValueRLPOutput();
+    tmp.startList();
+    for (final Transaction transaction : transactions) {
+      transaction.writeTo(tmp);
+      includedTransactions.add(transaction);
+      if (limiter.isLastTransactionToIncludeWithoutExceedingLimit(tmp)) {
```
mbaxter: Do we really need the extra complexity of this `Limiter` class when it just runs a >= check?  I think it would be simpler to just pass the size constraints to this method.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/messages/TransactionsMessage.java
```diff
@@ -47,7 +49,24 @@ public static TransactionsMessage create(final Iterable<Transaction> transaction
...
+        break;
```
mbaxter: It looks like the logic here is packing up transactions until we exceed the 1MB limit.  There's no check that the last transaction isn't huge such that the whole message could exceed the hard 16MB limit.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/messages/TransactionsMessage.java
```diff
@@ -47,7 +49,24 @@ public static TransactionsMessage create(final Iterable<Transaction> transaction
     return new TransactionsMessage(tmp.encoded());
   }
 
-  private TransactionsMessage(final BytesValue data) {
+  public static LimitedTransactionsMessages createLimited(
```
mbaxter: I would put this factory method on `LimitedTransactionsMessages` rather than `TransactionsMessage` since it returns a `LimitedTransactionsMessages`.

ethereum/rlp/src/main/java/tech/pegasys/pantheon/ethereum/rlp/AbstractRLPOutput.java
abdelhamidbakhta: Done.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/messages/TransactionsMessage.java
abdelhamidbakhta: Done.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/messages/TransactionsMessage.java
abdelhamidbakhta: I think yes. I mean in terms of being future proof if we want to have multiple policies, it might then be configurable through multiple instances of the interface. And for testing purpose it is easier to mock the Limiter interface.

mbaxter: I don't see how this makes the code easier to test.  In terms of future-proofing, one of our design principles is YAGNI (You Ain't Gonna Need It) :).  I think it's better to start simple and make the code more complex as needed.

mbaxter: Not a blocker though^ . I'll leave it up to you.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/messages/TransactionsMessageLimiter.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package tech.pegasys.pantheon.ethereum.eth.messages;
+
+import tech.pegasys.pantheon.ethereum.rlp.BytesValueRLPOutput;
+
+public interface TransactionsMessageLimiter {
```
mbaxter: (optional) One other idea to simplify the class structure would be to make this a static internal class of `LimitedTransactionsMessages` marked as a `@FunctionalInterface`, and keep the default limiter (`DefaultTransactionMessageLimiter`) as a static field on `LimitedTransactionsMessages`.  You could then have a method `createLimited(transactions)` that uses the default limiter.  This means fewer files and fewer dependencies that have to be passed around.

abdelhamidbakhta: I have opted for this idea. Indeed it simplifies the design.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/messages/TransactionsMessage.java
abdelhamidbakhta: I have simplified according to your idea of using a FunctionalInterface

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/messages/TransactionsMessage.java
abdelhamidbakhta: We create a follow up ticket to implement this check at lower level (devp2p layer)

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/messages/LimitedTransactionsMessages.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+package tech.pegasys.pantheon.ethereum.eth.messages;
+
+import tech.pegasys.pantheon.ethereum.core.Transaction;
+import tech.pegasys.pantheon.ethereum.rlp.BytesValueRLPOutput;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public final class LimitedTransactionsMessages {
```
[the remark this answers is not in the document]
abdelhamidbakhta: Done.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/messages/LimitedTransactionsMessages.java
```diff
@@ -0,0 +1,64 @@
...
+
+  public final TransactionsMessage getTransactionsMessage() {
+    return transactionsMessage;
+  }
+
+  public final Set<Transaction> getIncludedTransactions() {
+    return includedTransactions;
+  }
+
+  public static void setLimiter(final TransactionsMessageLimiter limiter) {
```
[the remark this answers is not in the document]
abdelhamidbakhta: Aborted.

ethereum/rlp/src/main/java/tech/pegasys/pantheon/ethereum/rlp/AbstractRLPOutput.java
```diff
@@ -141,6 +141,15 @@ public int encodedSize() {
     return payloadSizes[0];
   }
 
+  /**
+   * Computes a temporary encoded data size
+   *
+   * @return The size of the RLP-encoded data written to this output.
+   */
+  public int uncheckedEncodedSize() {
```
mbaxter: Can you cut this method since we don't need it and it's a bit confusing in its behavior?

abdelhamidbakhta: Ok.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/Transaction.java
```diff
@@ -494,7 +494,7 @@ public Transaction signAndBuild(final SECP256K1.KeyPair keys) {
       return build();
     }
 
-    protected SECP256K1.Signature computeSignature(final SECP256K1.KeyPair keys) {
+    SECP256K1.Signature computeSignature(final SECP256K1.KeyPair keys) {
```
mbaxter: Doesn't look like you need to expose this.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/messages/LimitedTransactionsMessages.java
```diff
@@ -0,0 +1,69 @@
...
+      transaction.writeTo(encodedTransaction);
+      BytesValue encodedBytes = encodedTransaction.encoded();
+      // Break if individual transaction size exceeds limit
+      if (encodedBytes.size() > LIMIT && (messageSize != 0)) {
+        break;
+      }
+      message.writeRLPUnsafe(encodedBytes);
+      includedTransactions.add(transaction);
+      // Check if last transaction to add to the message
+      if (messageSize + encodedBytes.size() > LIMIT) {
```
mbaxter: (optional) You could update messageSize and then check the limit.  That way we're not performing that addition twice. 
```suggestion
      messageSize += encodedBytes.size();
      if (messageSize > LIMIT) {
```

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/transactions/TransactionsMessageSender.java
```diff
@@ -12,36 +12,36 @@
...
 
   private void sendTransactionsToPeer(final EthPeer peer) {
     final Set<Transaction> allTxToSend = transactionTracker.claimTransactionsToSendToPeer(peer);
     while (!allTxToSend.isEmpty()) {
-      final Set<Transaction> subsetToSend =
-          allTxToSend.stream().limit(MAX_BATCH_SIZE).collect(toSet());
-      allTxToSend.removeAll(subsetToSend);
+      final LimitedTransactionsMessages limitedTransactionsMessages =
+          LimitedTransactionsMessages.createLimited(allTxToSend);
+      allTxToSend.removeAll(limitedTransactionsMessages.getIncludedTransactions());
```
mbaxter: (optional) You could also have LimitedTransactionsMessages.createLimited return a stream of `TransactionMessages`, then you don't need the extra `removeAll` bookkeeping and the extra memory for the included tx's list.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/messages/LimitedTransactionsMessagesTest.java
```diff
@@ -0,0 +1,57 @@
...
+  private final TransactionsMessage sampleTransactionMessages =
+      TransactionsMessage.create(sampleTxs);
+  private final LimitedTransactionsMessages sampleLimitedTransactionsMessages =
+      new LimitedTransactionsMessages(sampleTransactionMessages, sampleTxs);
+
+  @Test
+  public void createLimited() {
+    final Set<Transaction> txs = generator.transactions(6000);
+    final LimitedTransactionsMessages firstMessage = LimitedTransactionsMessages.createLimited(txs);
+    assertEquals(5219, firstMessage.getIncludedTransactions().size());
```
mbaxter: I think its worth summing up the final size (in bytes) of all these messages and asserting the total size is < 2 x limit

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/messages/LimitedTransactionsMessagesTest.java
```diff
@@ -0,0 +1,57 @@
...
+  private final BlockDataGenerator generator = new BlockDataGenerator();
+  private final Set<Transaction> sampleTxs = generator.transactions(1);
+  private final TransactionsMessage sampleTransactionMessages =
+      TransactionsMessage.create(sampleTxs);
+  private final LimitedTransactionsMessages sampleLimitedTransactionsMessages =
+      new LimitedTransactionsMessages(sampleTransactionMessages, sampleTxs);
+
+  @Test
+  public void createLimited() {
+    final Set<Transaction> txs = generator.transactions(6000);
```
mbaxter: (optional) If you add a `createLimited` method that accepts a configurable limit, you won't need to generate so many txs.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/messages/LimitedTransactionsMessagesTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+  private final BlockDataGenerator generator = new BlockDataGenerator();
+  private final Set<Transaction> sampleTxs = generator.transactions(1);
+  private final TransactionsMessage sampleTransactionMessages =
+      TransactionsMessage.create(sampleTxs);
+  private final LimitedTransactionsMessages sampleLimitedTransactionsMessages =
+      new LimitedTransactionsMessages(sampleTransactionMessages, sampleTxs);
+
+  @Test
+  public void createLimited() {
```
mbaxter: I'd also add a test to see how the limiter behaves when the first tx in a list exceeds its size limit.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/Transaction.java
abdelhamidbakhta: Ok

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/messages/LimitedTransactionsMessagesTest.java
abdelhamidbakhta: I think it is valuable to test with the real limit

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/messages/LimitedTransactionsMessagesTest.java
abdelhamidbakhta: Done.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/messages/LimitedTransactionsMessagesTest.java
abdelhamidbakhta: Done.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/messages/LimitedTransactionsMessagesTest.java
```diff
@@ -0,0 +1,82 @@
...
+    final LimitedTransactionsMessages firstMessage = LimitedTransactionsMessages.createLimited(txs);
+    assertEquals(5219, firstMessage.getIncludedTransactions().size());
+    txs.removeAll(firstMessage.getIncludedTransactions());
+    assertEquals(781, txs.size());
+    final LimitedTransactionsMessages secondMessage =
+        LimitedTransactionsMessages.createLimited(txs);
+    assertEquals(781, secondMessage.getIncludedTransactions().size());
+    txs.removeAll(secondMessage.getIncludedTransactions());
+    assertEquals(0, txs.size());
+    assertTrue(
```
mbaxter: This test isn't quite right: if you have different inputs (say you generate 2 x 5219 tx's), this check won't necessarily hold.  

The intention here is to check that no single message will ever exceed the max expected size.  I think that would be easier to check with a separate test that explicitly tries to hit the max message size.  You can do that by processing 3 transactions with a size of `LIMIT` and checking that the first 2 are packed up together.  Then process another 3 tx's with sizes {`LIMIT`,` LIMIT` + 1, `LIMIT`} so that each transaction is packaged individually.

So, I would cut this assertion and write a separate test that checks the max message size.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/messages/LimitedTransactionsMessagesTest.java
mbaxter: Using the real limit doesn't help us a whole lot here.  We just end up sprinkling magic numbers (5219, for example) throughout the tests.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/messages/LimitedTransactionsMessagesTest.java
```diff
@@ -0,0 +1,106 @@
...
+
+import java.util.HashSet;
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+import org.junit.Test;
+
+public class LimitedTransactionsMessagesTest {
+
+  private static final int LIMIT = 1048576;
```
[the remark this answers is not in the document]
abdelhamidbakhta: Done

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/messages/LimitedTransactionsMessagesTest.java
```diff
@@ -0,0 +1,106 @@
...
+    assertTrue(
+        (firstMessage.getTransactionsMessage().getSize()
+                + secondMessage.getTransactionsMessage().getSize())
+            < 2 * LIMIT);
+  }
+
+  @Test
+  public void createLimitedWithFirstTransactionExceedingLimit() {
+    final Set<Transaction> txs = new HashSet<>();
+    txs.add(generator.transaction(BytesValue.wrap(new byte[LIMIT - 180])));
```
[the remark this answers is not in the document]
abdelhamidbakhta: i have created a descriptive variable

PR #1273, opened 2019-04-12
Use NATS broker to wrap JSON RPC calls asynchronously

abdelhamidbakhta:
## PR description
Use NATS message broker system to wrap JSON RPC calls using publish / subscribe mechanism.
- one topic used to write JSON RPC request
- process RPC call
- one topic used to write JSON RPC response


ajsutton:
Closing as discussed offline.  We still have the code in the branch but it won't be landing anytime soon.

PR #1282, opened 2019-04-15
[PAN-2560] Cleanup PeerConnection interface

mbaxter:
## PR description
Small cleanup to the PeerConnection interface:
- Keep local / remote addresses as `InetSocketAddress` values
- Add `getEnode` method, remove `isRemoteEnode` since this can be checked directly now
- Rename `getPeer` to `getPeerInfo`

### DIFF
--- a/ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/netty/NettyPeerConnection.java
+++ b/ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/netty/NettyPeerConnection.java
@@ -26,7 +26,7 @@ import tech.pegasys.pantheon.ethereum.p2p.wire.messages.WireMessageCodes;
 import tech.pegasys.pantheon.metrics.Counter;
 import tech.pegasys.pantheon.metrics.LabelledMetric;
 
-import java.net.SocketAddress;
+import java.net.InetSocketAddress;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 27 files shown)


ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/netty/NettyPeerConnection.java
```diff
@@ -156,13 +156,13 @@ public boolean isDisconnected() {
   }
 
   @Override
-  public SocketAddress getLocalAddress() {
-    return ctx.channel().localAddress();
+  public InetSocketAddress getLocalAddress() {
+    return (InetSocketAddress) ctx.channel().localAddress();
```
shemnon: Is this always an `InetStocketAddress` in the manner we use netty?  Are there other things it could be, that we might actually use?

mbaxter: According to the docs on `SocketAddress`:
> This class represents a Socket Address with no protocol attachment. As an abstract class, it is meant to be subclassed with a specific, protocol dependent, implementation.

And `InetSocketAddress`:
> This class implements an IP Socket Address (IP address + port number)

I think the change makes sense.  

We were casting it all over the place assuming it was an `InetSocketAddress` anyway and I think its better to just be explicit.

mbaxter: Also, `InetSocketAddress` is actually the only subclass of `SocketAddress`

mbaxter: I actually think we could probably just get rid of these methods.  I'll try to do that.

shemnon: It would localize where the cast is in the unlikely case it's ever a problem.  But if we can get rid of the methods entirely that is also less code, which is generally better.

PR #1284, opened 2019-04-15
Fix RunnerBuilder method names

mbaxter:
## PR description
I noticed the method name `RunnerBuilder.discoveryPort` is inaccurate because it actually defines the listening port for both wire-level tcp connections and discovery-level udp messages. Updated this method name to be more generic.  And also ported `discoveryHost` for symmetry.


pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -976,8 +976,8 @@ private void synchronize(
             .p2pEnabled(p2pEnabled)
             .discovery(peerDiscoveryEnabled)
             .ethNetworkConfig(ethNetworkConfig)
-            .discoveryHost(discoveryHost)
-            .discoveryPort(discoveryPort)
+            .advertisedHost(discoveryHost)
+            .listenPort(discoveryPort)
```
ajsutton: Is it weird to have the names mismatching here?

pantheon/src/main/java/tech/pegasys/pantheon/RunnerBuilder.java
```diff
@@ -137,12 +137,12 @@ public RunnerBuilder ethNetworkConfig(final EthNetworkConfig ethNetworkConfig) {
...
 
-  public RunnerBuilder discoveryHost(final String discoveryHost) {
-    this.discoveryHost = discoveryHost;
+  public RunnerBuilder advertisedHost(final String advertisedHost) {
+    this.advertisedHost = advertisedHost;
     return this;
   }
 
-  public RunnerBuilder discoveryPort(final int listenPort) {
+  public RunnerBuilder listenPort(final int listenPort) {
```
ajsutton: `listenPort` feels too generic here.  I wonder if these should be p2pAdvertisedHost and p2pListenPort?  A bigger refactoring would be to make an actual `P2pConfiguration` object to match `JsonRpcConfiguration` and friends but that's probably scope creep on what you were working on. :)

mbaxter: Updated the names.  We do have a `NetworkingConfiguration` object, although it's probably due for some cleanup.

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
mbaxter: Yep, thats weird :) . Updated the variable names.  Also updated the description of the corresponding options.

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -239,14 +239,14 @@ void setBootnodes(final List<String> values) {
   @Option(
       names = {"--p2p-host"},
       paramLabel = MANDATORY_HOST_FORMAT_HELP,
-      description = "Host for P2P peer discovery to listen on (default: ${DEFAULT-VALUE})",
+      description = "Ip address this node advertises to its peers (default: ${DEFAULT-VALUE})",
```
mbaxter: fyi @PegaSysEng/pliny - I'm updating the CLI descriptions here and below.  Let me know if you see any issues.

PR #1286, opened 2019-04-16
Switch to new sync target if it exceeds the td threshold

ajsutton:
## PR description
When evaluating if the sync target should be changed, we would only switch if the new best peer's height was better by some amount.  We should switch if the difficulty or height exceeds the "better" threshold (as long as the difficulty is at least equal).

Also splits this calculation into a separate class so it can be reused in the pipeline chain download and adds unit tests for it.


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/fullsync/BetterSyncTargetEvaluator.java
```diff
@@ -0,0 +1,63 @@
...
+  private final SynchronizerConfiguration config;
+  private final EthPeers ethPeers;
+
+  public BetterSyncTargetEvaluator(
+      final SynchronizerConfiguration config, final EthPeers ethPeers) {
+    this.config = config;
+    this.ethPeers = ethPeers;
+  }
+
+  public boolean shouldSwitchSyncTarget(final EthPeer currentPeer) {
```
mbaxter:
```suggestion
  public boolean shouldSwitchSyncTarget(final EthPeer currentSycTarget) {
```

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/fullsync/BetterSyncTargetEvaluator.java
```diff
@@ -0,0 +1,63 @@
...
+              }
+              // Require some threshold to be exceeded before switching targets to keep some
+              // stability when multiple peers are in range of each other
+              final ChainState bestPeerChainState = bestPeer.chainState();
+              final UInt256 tdDifference =
+                  bestPeerChainState
+                      .getBestBlock()
+                      .getTotalDifficulty()
+                      .minus(currentPeerChainState.getBestBlock().getTotalDifficulty());
+              if (tdDifference.compareTo(config.downloaderChangeTargetThresholdByTd()) > 0) {
```
mbaxter: The defaults we have for these values don't seem to make much sense.  Looks like mainnet is changing by ~10^15 per block but the default threshold is 10^9.  And td is tricky because the rate of change should depend on the particular chain.  In any case, we should probably try to pick better defaults as part of this PR while we're here.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/sync/fullsync/BetterSyncTargetEvaluatorTest.java
```diff
@@ -0,0 +1,157 @@
...
+import tech.pegasys.pantheon.ethereum.eth.manager.EthPeer;
+import tech.pegasys.pantheon.ethereum.eth.manager.EthPeers;
+import tech.pegasys.pantheon.ethereum.eth.sync.SynchronizerConfiguration;
+import tech.pegasys.pantheon.util.uint.UInt256;
+
+import java.util.Optional;
+
+import org.junit.Test;
+
+public class BetterSyncTargetEvaluatorTest {
```
mbaxter: nice test coverage :D

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/fullsync/BetterSyncTargetEvaluator.java
ajsutton: Done.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/fullsync/BetterSyncTargetEvaluator.java
ajsutton: That's a particularly astute yet inconvenient observation... 

The situation we want to avoid is where our first peer has a very long but low difficulty chain, trying to draw us into a tar pit and spending a long time sync'ing their chain. Currently if their chain height was higher we'd continue syncing to them until we reached their head and only then switch to the actual best peer and begin syncing the right chain.

That said, it takes quite a lot of effort to generate a really long PoW chain because the difficulty will always keep blocks around 15s apart. 

I'm inclined to set it to a high value like 10^18 which means the default is almost always switch on chain height before the TD threshold was reached. It would still differentiate between ETH and ETC though we have an explicit Dao block check for that as well now. Having the setting around is potentially quite useful for cases like when the Ropsten fork went wrong - you'd want to set it to something much smaller to ensure you switched to the best chain faster.

mbaxter: Sounds reasonable.  The height threshold is pretty low as well - should we bump that up?

PR #1288, opened 2019-04-16
[PAN-2560] Consolidate p2p node info methods

mbaxter:
## PR description
We currently have 3 different methods on `P2PNetwork` for retrieving local node information: `getSelfEnodeUrl`, `getAdvertisedPeer`, and `getLocalPeerInfo`.  This PR consolidates the three methods down to 1: `getLocalEnode()`.  And adds a more explicit method for checking on the status of the discovery service: `isDiscoveryEnabled()`.  

Other incidental changes include:
- Adding `EnodeURL.getEffectiveDiscoveryPort()` returning an `int` indicating the port where discovery should be listening if it is enabled
- A bit of cleanup for some JSON-RPC methods that were catching and logging errors: updated server logic so that unexpected errors are always logged.


ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/JsonRpcHttpService.java
```diff
@@ -455,7 +455,7 @@ private JsonRpcResponse process(final JsonObject requestJson, final Optional<Use
         LOG.debug(e);
         return errorResponse(id, JsonRpcError.INVALID_PARAMS);
       } catch (final RuntimeException e) {
-        LOG.debug(e);
+        LOG.error(e);
```
ajsutton: This will lose the stack trace because it resolves to LOG.error(Object) which converts the param to a String with toString().  
```suggestion
        LOG.error("Error processing JSON-RPC request", e);
```

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/AdminNodeInfo.java
```diff
@@ -62,53 +58,46 @@ public String getName() {
...
-                    advertisedPeer.getEndpoint().getHost() + ":" + peerInfo.getPort());
-              });
-      response.put("id", nodeId.toString().substring(2));
-      response.put("name", clientVersion);
+    final BytesValue nodeId = enode.getNodeId();
+    response.put("enode", enode.toString());
+    ports.put("discovery", enode.getEffectiveDiscoveryPort());
+    response.put("ip", enode.getIp());
+    response.put("listenAddr", enode.getIp() + ":" + enode.getListeningPort());
+    response.put("id", nodeId.toString().substring(2));
```
ajsutton:
```suggestion
    response.put("id", nodeId.toUnprefixedString());
```

pantheon/src/main/java/tech/pegasys/pantheon/Runner.java
```diff
@@ -195,7 +196,8 @@ private void writePantheonPortsToFile() {
     }
   }
 
-  public Optional<? extends Peer> getAdvertisedPeer() {
-    return networkRunner.getNetwork().getAdvertisedPeer();
+  @VisibleForTesting
+  P2PNetwork getP2PNetwork() {
```
ajsutton: It's quite unfortunate to have to expose the whole `P2PNetwork`.  Is it possible for this to only return the local enode (via delegation)?

PR #1289, opened 2019-04-16
Allow use of large chain ids

jframe:
## PR description
This updates chainId to be a BigInteger so that large values for a chainId can be used. Decided on using BigInteger as Geth currently using BigInt.


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/Transaction.java
```diff
@@ -87,14 +89,14 @@ public static Transaction readFrom(final RLPInput input) throws RLPException {
...
-      recId = (byte) (v - REPLAY_UNPROTECTED_V_BASE);
-    } else if (v > REPLAY_PROTECTED_V_MIN) {
-      chainId = (v - REPLAY_PROTECTED_V_BASE) / 2;
-      recId = (byte) (v - (2 * chainId + REPLAY_PROTECTED_V_BASE));
+    BigInteger chainId = BigInteger.valueOf(-1);
+    if (v.equals(REPLAY_UNPROTECTED_V_BASE) || v.equals(REPLAY_UNPROTECTED_V_BASE_PLUS_1)) {
+      recId = v.subtract(REPLAY_UNPROTECTED_V_BASE).byteValue();
+    } else if (v.compareTo(REPLAY_PROTECTED_V_MIN) > 0) {
+      chainId = v.subtract(REPLAY_PROTECTED_V_BASE).divide(TWO);
+      recId = v.subtract(TWO.multiply(chainId).add(REPLAY_PROTECTED_V_BASE)).byteValue();
```
ajsutton: Probably should use `byteValueExact` instead of `byteValue` here and line 96.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/Transaction.java
```diff
@@ -87,14 +89,14 @@ public static Transaction readFrom(final RLPInput input) throws RLPException {
...
-    final int v = input.readIntScalar();
+    final BigInteger v = input.readBigIntegerScalar();
     final byte recId;
-    int chainId = -1;
-    if (v == REPLAY_UNPROTECTED_V_BASE || v == REPLAY_UNPROTECTED_V_BASE + 1) {
-      recId = (byte) (v - REPLAY_UNPROTECTED_V_BASE);
-    } else if (v > REPLAY_PROTECTED_V_MIN) {
-      chainId = (v - REPLAY_PROTECTED_V_BASE) / 2;
-      recId = (byte) (v - (2 * chainId + REPLAY_PROTECTED_V_BASE));
+    BigInteger chainId = BigInteger.valueOf(-1);
```
ajsutton: This probably shouldn't be using `-1` as a sentinel value now that we're not using a primitive.  Probably should be `Optional<BigInteger> chainId = Optional.empty();` and that can flow through to `Builder.chainId`.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/Transaction.java
```diff
@@ -426,9 +429,9 @@ public String toString() {
 
     protected Address sender;
 
-    protected int chainId = -1;
+    protected BigInteger chainId = BigInteger.valueOf(-1);
```
ajsutton:
```suggestion
    protected Optional<BigInteger> chainId = Optional.empty();
```

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/Transaction.java
```diff
@@ -426,9 +429,9 @@ public String toString() {
...
 
-    public Builder chainId(final int chainId) {
+    public Builder chainId(final BigInteger chainId) {
       this.chainId = chainId;
```
ajsutton: Two possible approaches here, either `public Builder chainId(final Optional<BigInteger> chainId)` which is just directly assigned or `public Builder chainId(final BigInteger chainId)` which is wrapped with `Optional.of` - if you don't have a chain ID don't call `chainId` on the builder.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/Transaction.java
```diff
@@ -495,8 +498,8 @@ public Transaction signAndBuild(final SECP256K1.KeyPair keys) {
     }
 
     protected SECP256K1.Signature computeSignature(final SECP256K1.KeyPair keys) {
-      final OptionalInt optionalChainId =
-          chainId > 0 ? OptionalInt.of(chainId) : OptionalInt.empty();
+      final Optional<BigInteger> optionalChainId =
+          chainId.signum() == 1 ? Optional.of(chainId) : Optional.empty();
```
ajsutton: With the changes above you then don't need this and can just use `chainId` directly since it's already an optional.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/MainnetTransactionValidator.java
```diff
@@ -37,7 +38,7 @@
  */
 public class MainnetTransactionValidator implements TransactionValidator {
 
-  public static final int NO_CHAIN_ID = -1;
+  public static final BigInteger NO_CHAIN_ID = BigInteger.valueOf(-1);
```
ajsutton: Shouldn't be using a sentinel value here - I suspect this constant can just go away and `Optional.empty()` used in its place.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/MainnetTransactionValidator.java
```diff
@@ -57,10 +58,10 @@ public MainnetTransactionValidator(
   public MainnetTransactionValidator(
       final GasCalculator gasCalculator,
       final boolean checkSignatureMalleability,
-      final int chainId) {
+      final BigInteger chainId) {
     this.gasCalculator = gasCalculator;
     this.disallowSignatureMalleability = checkSignatureMalleability;
-    this.chainId = chainId > 0 ? OptionalInt.of(chainId) : OptionalInt.empty();
+    this.chainId = chainId.signum() == 1 ? Optional.of(chainId) : Optional.empty();
```
ajsutton: Just pass in `Optional<BigInteger>` and avoid the sentinel check here too.

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -1067,10 +1067,16 @@ private EthNetworkConfig updateNetworkConfig(final NetworkName network) {
...
+                  .map(BigInteger::intValueExact)
                   .orElse(EthNetworkConfig.getNetworkConfig(MAINNET).getNetworkId()));
         } catch (final DecodeException e) {
           throw new ParameterException(
               this.commandLine, String.format("Unable to parse genesis file %s.", genesisFile), e);
+        } catch (final ArithmeticException e) {
+          throw new ParameterException(
+              this.commandLine,
+              "No networkId specified and chainId in "
+                  + "genesis file is too large to be used as a networkId");
```
ajsutton: ugh, this is quite unfortunate but probably better to bail out with an error than to default to something else silently.

config/src/main/java/tech/pegasys/pantheon/config/ConfigUtil.java
```diff
@@ -22,4 +23,15 @@ public static OptionalLong getOptionalLong(final JsonObject jsonObject, final St
         ? OptionalLong.of(jsonObject.getLong(key))
         : OptionalLong.empty();
   }
+
+  public static BigInteger getBigInteger(final JsonObject jsonObject, final String key) {
```
CjHare: `Optional<BigInteger>` for consistency with the other method in this class

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/Transaction.java
```diff
@@ -426,13 +429,18 @@ public String toString() {
...
-    protected int chainId = -1;
+    protected Optional<BigInteger> chainId = Optional.empty();
 
-    public Builder chainId(final int chainId) {
+    public Builder chainId(final Optional<BigInteger> chainId) {
       this.chainId = chainId;
       return this;
     }
 
+    public Builder chainId(final BigInteger chainId) {
```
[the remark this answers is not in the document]
jframe: just removed actually. don't think it's necessary afterall

config/src/main/java/tech/pegasys/pantheon/config/ConfigUtil.java
jframe: done

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/Transaction.java
jframe: done

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/Transaction.java
jframe: done

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/Transaction.java
jframe: done

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/Transaction.java
jframe: done. used approach of defaulting to Optional.empty only calling chainId on builder if you have a chainId

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/Transaction.java
jframe: done

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/MainnetTransactionValidator.java
jframe: done

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/MainnetTransactionValidator.java
jframe: done

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
jframe: yeah it's not nice, but I think it's the best thing to do

jframe:
Updated so make chainId optional in ProtocolSchedule as well. This allows having no chainId when using FixedDifficultyProtocolSchedule without the need to special -1 chainId value being used.

PR #1291, opened 2019-04-16
Pipeline based full sync

ajsutton:
## PR description
Add a full sync implementation that uses the pipeline based chain downloader.  It's currently toggled off.

 * Adjusts `CheckpointHeaderFetcher` so that the remote peer's reported head is requested as a check point if the next checkpoint header would be beyond the peer's head. This ensures we actually reach the peer's reported head rather than just the checkpoint prior to it
 * Handle checkpoint ranges where the two headers are adjacent (thus we don't need to download any headers in between)
 * There are a few tests in FullSyncChainDownloaderTest which depend on internals of the eth task based downloader so are skipped when the pipeline downloader is in use.  #1286 adds unit tests to cover those cases without needing to depend on the internals of the downloader (and those changes are also in the PR).


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/CheckpointHeaderFetcher.java
```diff
@@ -66,6 +67,12 @@ public CheckpointHeaderFetcher(
       if (additionalHeaderCount == 0) {
         return completedFuture(singletonList(targetHeader));
       }
+    } else if (lastHeader.getNumber() + skip >= peer.chainState().getEstimatedHeight()) {
```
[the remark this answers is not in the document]
mbaxter: To mimic our current approach.  What about making `CheckpointRange.end` an `Optional`?  In the case where `end` is empty, we can just download the next contiguous set of blocks from the previous header.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/DownloadHeadersStep.java
```diff
@@ -58,13 +64,20 @@ public DownloadHeadersStep(
 
   private CompletableFuture<List<BlockHeader>> downloadHeaders(
       final CheckpointRange checkpointRange) {
+    LOG.debug(
+        "Downloading headers for range {} to {}",
+        checkpointRange.getStart().getNumber(),
+        checkpointRange.getEnd().getNumber());
+    if (checkpointRange.getSegmentLengthExclusive() == 0) {
```
mbaxter: (optional) We could also update `DownloadHeaderSequenceTask` to just return the reference header if the segment length is 0.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/DownloadHeadersStep.java
```diff
@@ -58,13 +64,20 @@ public DownloadHeadersStep(
...
+      return completedFuture(emptyList());
+    }
     return DownloadHeaderSequenceTask.endingAtHeader(
             protocolSchedule,
             protocolContext,
             ethContext,
             checkpointRange.getEnd(),
-            // -1 because we don't want to request the range starting header
-            checkpointRange.getSegmentLength() - 1,
+            checkpointRange.getSegmentLengthExclusive(),
```
mbaxter: nice :D

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/fullsync/FullSyncDownloadPipelineFactory.java
```diff
@@ -0,0 +1,116 @@
...
+    final CheckpointRangeSource checkpointRangeSource =
+        new CheckpointRangeSource(
+            new CheckpointHeaderFetcher(
+                syncConfig, protocolSchedule, ethContext, Optional.empty(), metricsSystem),
+            this::shouldContinueDownloadingFromPeer,
+            ethContext.getScheduler(),
+            target.peer(),
+            target.commonAncestor(),
+            syncConfig.downloaderCheckpointTimeoutsPermitted(),
+            Duration.ofSeconds(5));
```
mbaxter: Should this duration be stored as a constant somewhere?

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/sync/fullsync/FullSyncChainDownloaderTest.java
```diff
@@ -92,15 +95,12 @@ public void setupTest() {
...
 
-  // A few tests break encapsulation and access the "current task", hence casting to a concrete type
-  @SuppressWarnings("unchecked")
-  private EthTaskChainDownloader<Void> downloader(final SynchronizerConfiguration syncConfig) {
-    return (EthTaskChainDownloader<Void>)
-        FullSyncChainDownloader.create(
-            syncConfig, protocolSchedule, protocolContext, ethContext, syncState, metricsSystem);
+  private ChainDownloader downloader(final SynchronizerConfiguration syncConfig) {
+    return FullSyncChainDownloader.create(
+        syncConfig, protocolSchedule, protocolContext, ethContext, syncState, metricsSystem);
```
mbaxter: Should you set up these tests to run on both versions of the downloader?

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/fullsync/FullSyncDownloadPipelineFactory.java
ajsutton: Done.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/DownloadHeadersStep.java
ajsutton: We could but it would wind up being a bit late in the process - after the peer has been selected etc.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/CheckpointHeaderFetcher.java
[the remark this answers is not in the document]
ajsutton: Making end optional is going to take a fair bit of refactoring so I've started with just shifting to adjusting the skip value.  If the peer has re-orged so our previous checkpoint header is no longer on the canonical chain we won't be able to get any extra checkpoints from it or any extra headers if end was optional anyway.  Ultimately we'll wind up switching sync targets because we don't get new headers after a number of retries and things will continue ok.

Will take a look at making the end optional though and see what it winds up looking like.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/CheckpointHeaderFetcher.java
```diff
@@ -66,6 +67,12 @@ public CheckpointHeaderFetcher(
       if (additionalHeaderCount == 0) {
```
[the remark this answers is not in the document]
ajsutton: Done.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/CheckpointHeaderFetcher.java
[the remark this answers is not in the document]
ajsutton: Making the end optional came out pretty well - it's not as isolated, spreading across 3 classes but I think that's reasonable and it's probably good to take the opportunity to potentially get a few more blocks past the sync target's chain head.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/sync/fullsync/FullSyncChainDownloaderTest.java
ajsutton: Oh forgot this one sorry.  I typically don't bother with the aim of just switching fairly quickly and deleting the old implementation but since I went and made it so easily configurable I've parameterised the main unit test so it covers both.  Things like ATs will still only use the default but we have at least some coverage.

PR #1295, opened 2019-04-17
PAN-2485#Add net_services json rpc endpoint

usmananwar:
## PR description
Implemented a JSON RPC method "net_services" that takes no input
parameter and returns an object
containing a list of enabled services (jsonrpc, p2p, ws, metrics) with
"host" and "port", as show below;

{
  "id":1,
  "jsonrpc": "2.0",
  "result": {
    "jsonrpc": {"host": "127.0.0.1", "port":3000},
    "p2p":{"host": "127.0.0.1", "port":3000},
    "ws":{"host": "127.0.0.1", "port":3000},
    "metrics":{"host": "127.0.0.1", "port":3000}
  }
}
In case a service is not enabled, that will not be included in the list;

{
  "id":1,
  "jsonrpc": "2.0",
  "result": {}
}

**This PR includes all the changes requested by reviewer in the PR# 1269 (closed uncommitted)

### DIFF
--- a/ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/NetServices.java
+++ b/ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/NetServices.java
@@ -21,7 +21,6 @@ import tech.pegasys.pantheon.ethereum.p2p.api.P2PNetwork;
 import tech.pegasys.pantheon.metrics.prometheus.MetricsConfiguration;
 
 import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.ImmutableMap.Builder;
 
 public class NetServices implements JsonRpcMethod {
 
@@ -33,7 +32,8 @@ public class NetServices implements JsonRpcMethod {
       final P2PNetwork p2pNetwork,
       final MetricsConfiguration metricsConfiguration) {
 
-    Builder<String, ImmutableMap<String, String>> servicesMapBuilder = ImmutableMap.builder();
+    ImmutableMap.Builder<String, ImmutableMap<String, String>> servicesMapBuilder =
+        ImmutableMap.builder();
 
     if (jsonRpcConfiguration.isEnabled()) {
       servicesMapBuilder.put(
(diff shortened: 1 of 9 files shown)


ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/NetServices.java
```diff
@@ -0,0 +1,86 @@
...
+                });
+      }
+    }
+    if (metricsConfiguration.isEnabled()) {
+      activeServices.put(
+          "metrics",
+          createServiceDetailsMap(metricsConfiguration.getHost(), metricsConfiguration.getPort()));
+    }
+
+    services =
```
shemnon: We should use an ImmutalbeMap.Builder instead of filling a mutable map and copying it into an ImmutableMap

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/NetServices.java
```diff
@@ -20,10 +20,8 @@
 import tech.pegasys.pantheon.ethereum.p2p.api.P2PNetwork;
 import tech.pegasys.pantheon.metrics.prometheus.MetricsConfiguration;
 
-import java.util.LinkedHashMap;
-import java.util.Map;
-
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableMap.Builder;
```
shemnon: This causes ErrorProne to throw an error.  Please use ImmutableMap.Builder instead of importing the inner class.

https://jenkins.pegasys.tech/job/Pantheon/job/PR-1295/2/execution/node/59/log/
https://errorprone.info/bugpattern/BadImport

usmananwar: Thanks. I have changed it and performed these gradle task aswell locally; (spotlessApply clean check test). I hope there are no more errors :)

PR #1301, opened 2019-04-18
Added sending private transactions initial content 

MadelineMurray:
Added sending private transactions initial content


docs/Privacy/Creating-Sending-Private-Transactions.md
```diff
@@ -0,0 +1,11 @@
+description: Creating and sending private transactions
+<!--- END of page meta data -->
+
+# Creating and Sending Private Transactions 
+
+The [EEA JavaScript library](https://github.com/PegaSysEng/eeajs) is provided to create and send signed 
+RLP-encoded private transactions. 
+
+!!! note
+    Private transactions either deploy contracts or call contract functions.
```
NicolasMassart: I guess we would need more than a note about what we can do with private tx then.

MadelineMurray: I'll be added more content in a follow up PR now the EEA JS library is working.

PR #1305, opened 2019-04-18
Remove duplicate init code from PantheonController instances

ajsutton:
## PR description
Introduce a `ConsensusAwarePantheonControllerBuilder` with subclasses for each of our consensus algorithms which is responsible for creating the `PantheonController`.  This allows the common init code to be shared with consensus specific subclasses only overriding parts they actually need to change.


pantheon/src/main/java/tech/pegasys/pantheon/controller/PantheonControllerBuilder.java
```diff
@@ -0,0 +1,335 @@
...
+import java.time.Clock;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.OptionalLong;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public abstract class PantheonControllerBuilder<C> {
```
shemnon: The generic `C` is an impediment to understanding what it really is.  The Google Style guide [section 5.2.8](https://google.github.io/styleguide/javaguide.html#s5.2.8-type-variable-names) allows for longer type names as long as it ends with a `T`.  I think we should adopt that pattern and replace `C` with `ConsensusContextT`, which should improve the readability.

pantheon/src/main/java/tech/pegasys/pantheon/controller/PantheonControllerBuilder.java
```diff
@@ -0,0 +1,335 @@
...
+  protected PrivacyParameters privacyParameters;
+  protected Path dataDirectory;
+  protected Clock clock;
+  protected Integer maxPendingTransactions;
+  protected KeyPair nodeKeys;
+  private StorageProvider storageProvider;
+  private final List<Runnable> shutdownActions = new ArrayList<>();
+  private RocksDbConfiguration rocksdDbConfiguration;
+
+  public static PantheonControllerBuilder<?> fromGenesisConfig(
```
shemnon: A builder with constructor parameters feels a bit off.  Perhaps these two `fromGenesisConfig` methods could be moved to a PantheonControllerBuilderFactory and this class could be AbstractPantheonControllerBuilder. The subclasses also don't add any builder functionality but are actually more of a strategy object for a generic factory.

Could this builder be a very simple builder that just holds the fields, and then in the build method delegates to the strategy/factory objects specific to the consensus types?  As late as possible?  Or perhaps keep the old class names and add a `.fromBuilder` method that will return an appropriate controller or throw an exception?

pantheon/src/main/java/tech/pegasys/pantheon/controller/PantheonControllerBuilder.java
ajsutton: We're pretty consistent in using `C` to indicate the consensus context across the code base.  Changing it to a longer name would add quite a lot of noise. It's also an invasive enough change that we should do it in a separate PR if we're going to.

pantheon/src/main/java/tech/pegasys/pantheon/controller/PantheonControllerBuilder.java
ajsutton: Well it's not actually a constructor param - it's a static factory method to create the right type of builder based on the config.  I've moved it to a separate `PantheonControllerBuilderFactory` class since that's essentially what `PantheonCommand` wants. `PantheonCommandTest` wants to inject a mock `PantheonControllerBuilder` and was using a `Function<EthNetworkConfig, PantheonControllerBuilder>` but can now use a mock `PantheonControllerBuilderFactory`.

Whether it's a static method or a separate class the idea of having a builder factory displeases me greatly but I can't see any real way around it.

The key challenge in all of this is getting Java's generics to play nice. We need the genesis config before we can decide what generic type is required but if the builder isn't generic then it when it creates the `ProtocolSchedule` it has to declare it as `ProtocolSchedule<?>` but then it doesn't have the matching consensus context type the subclasses need.

pantheon/src/main/java/tech/pegasys/pantheon/controller/PantheonControllerBuilder.java
shemnon: I agree it should not be done in this PR, it is quite invasive.  

But unless you already know how consensus state objects are used and passed around `C` may as well be `I`, `X`, or `T`.  It is present in broad swaths of the code but what it actually means is limited to very few locations that are hard to navigate to unless you already know exactly what it does.

pantheon/src/main/java/tech/pegasys/pantheon/controller/PantheonControllerBuilder.java
```diff
@@ -0,0 +1,308 @@
...
+    this.clock = clock;
+    return this;
+  }
+
+  public PantheonControllerBuilder<C> maxPendingTransactions(final int maxPendingTransactions) {
+    this.maxPendingTransactions = maxPendingTransactions;
+    return this;
+  }
+
+  public PantheonController<C> build() throws IOException {
```
shemnon: What do you think about moving this class inside `PantheonController` and calling it `PantheonController.Builder`?  It would be more consistent with our other builders.  What complicates it is the fact that the Builder class is subclassed by the other controllers. But to typical users that would be transparent.

Either way, your call.

pantheon/src/main/java/tech/pegasys/pantheon/controller/PantheonControllerBuilderFactory.java
```diff
@@ -0,0 +1,44 @@
...
+      final EthNetworkConfig ethNetworkConfig) {
+    return fromGenesisConfig(GenesisConfigFile.fromConfig(ethNetworkConfig.getGenesisConfig()))
+        .networkId(ethNetworkConfig.getNetworkId());
+  }
+
+  public PantheonControllerBuilder<?> fromGenesisConfig(final GenesisConfigFile genesisConfig) {
+    final GenesisConfigOptions configOptions = genesisConfig.getConfigOptions();
+    final PantheonControllerBuilder<?> builder;
+
+    if (configOptions.isEthHash()) {
```
shemnon: Not for this PR, but we should think of a way to make this pluggable without prior knowledge of all the controller builders.

ajsutton: Absolutely agreed. I can see a few places that pluggability is starting to emerge and would make this setup code cleaner (creating the storage provider being a prime example).

pantheon/src/main/java/tech/pegasys/pantheon/controller/PantheonControllerBuilder.java
ajsutton: Done.

PR #1313, opened 2019-04-23
Use header validation mode for ommers

ajsutton:
## PR description
During a fast sync, we should only fully validate a random sampling of ommer headers and otherwise skip the proof of work validity and ancestry checks which are expensive.

Profiling shows the PoW checks in particular were causing the majority of CPU usage during a fast sync for an ethash chain.


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/MainnetBlockBodyValidator.java
```diff
@@ -224,6 +224,17 @@ private boolean isOmmerValid(
       return false;
     }
 
+    if (ommerValidationMode != HeaderValidationMode.LIGHT
+        && ommerValidationMode != HeaderValidationMode.LIGHT_SKIP_DETACHED
+        && ommerValidationMode != HeaderValidationMode.LIGHT_DETACHED_ONLY) {
```
mbaxter: Could this be clarified with something like:  `!HeaderValidationMode.isLight(ommerValidationMode)` ?

PR #1316, opened 2019-04-23
Don't consider no available peers as a failure, just keep waiting

ajsutton:
## PR description
Change `AbstractRetryingPeerTask` to not consider no available peers a failed attempt.  This prevents the download pipeline doesn't fail if there are few peers and a lot of requests being made. When we do fail the pipeline, we discard any content we've already downloaded but not yet imported which then has to be re-requested so we don't make progress.


ajsutton:
Closing this in favour of a more comprehensive approach: https://github.com/PegaSysEng/pantheon/pull/1331

PR #1322, opened 2019-04-23
WIP: [Ethereal Hackathon] GraphQL EIP-1767 Implementation For Pantheon

fredlacs:
## Initial WIP PR
Implemented initial functionality to query for blocks with an integration test for `ethgetBlockByHash`.


iikirilov:
We believe the graphql methods should be separated from the json-rpc methods because it is possible in future extension work to greatly improve the performance by performing look ahead querying. This means that for the query `{ block { parent { number } } }` we can look ahead when fetching the block to realize that we need to fetch the parent. This would mean we do not need to actually fetch the block! Instead, we can directly fetch the parent. We can look ahead further to realise that we don't even need the full parent, instead we only need the number. This can greatly improve execution of nested queries such as `{ block { parent { parent { parent { parent { ... } } } } } }`

iikirilov:
However, on reflection, it may be possible to share a significant number of components including the vertx engine, the authentication procedure, and the blockchain querying component amongst other.

iikirilov:
We also appreciate that the dates in the copyright is wrong in many of the files.

shemnon:
Thank your for participating in the Ethereal 2019 hackathon.

PR #1323, opened 2019-04-23
[PAN-2465] Priority based on gas price when adding local transaction.

abdelhamidbakhta:
## PR description
- accept local transaction if the pool is full and gas price higher than the lowest price
- reject if transaction underpriced
## Fixed Issue(s)

fixes #PAN-2465


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/transactions/PendingTransactions.java
```diff
@@ -52,8 +52,9 @@
   private final Map<Hash, TransactionInfo> pendingTransactions = new HashMap<>();
   private final SortedSet<TransactionInfo> prioritizedTransactions =
       new TreeSet<>(
-          comparing(TransactionInfo::isReceivedFromLocalSource)
+          comparing(PendingTransactions.TransactionInfo::isReceivedFromLocalSource)
               .thenComparing(TransactionInfo::getSequence)
+              .thenComparing(txInfo -> txInfo.getTransaction().getGasPrice())
```
shemnon: I think this will still remove the tx with the highest nonce, then use gas as a tie-breaker.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/transactions/PendingTransactionsTest.java
```diff
@@ -132,15 +120,29 @@ public void shouldPrioritizeLocalTransaction() {
...
+    assertThat(transactions.size()).isEqualTo(MAX_TRANSACTIONS);
+    assertTrue(
+        transactions.addLocalTransaction(
+            createTransaction(MAX_TRANSACTIONS + 1, Wei.of(startGasPrice + 1))));
+  }
 
-    for (int i = 1; i <= MAX_TRANSACTIONS; i++) {
-      transactions.addLocalTransaction(createTransaction(i));
+  @Test
+  public void shouldNotAcceptLocalTransactionIfFullAndUnderpriced() {
```
shemnon: I think we will need more test cases, so a parameterized unit test may be in order.

All of these assume full local pools.


* For a queue with a single account and nonces 1,2,3 when 2 is the lowest price, and a new TX at 4 comes in higher than 2, do we drop 2 or reject?
* For a queue with a single account and nonces 1,2,3 when 2 is the lowest price, and a new TX on a different account higher than 2 comes in, do we  drop 2 or reject?  if we drop 2 do we drop 3?
* account A with nonce 1, gas 1, account B with nonce 3, gas 3.  new TX with nonce 2, gas 2, we should drop TX A1.  If I read it correctly we will reject.
* account A with nonce 1, gas 3, account B with nonce 3, gas 3.  new TX with nonce 2, gas 2, we should reject.

abdelhamidbakhta:
Won't do.

PR #1327, opened 2019-04-24
[PAN-2576] Allow whitespace in path

abdelhamidbakhta:
## PR description
- handle whitespace in path
- handle parentheses in path
## Fixed Issue(s)

fixes #PAN-2576


abdelhamidbakhta:
Follow up ticket has been created to generalize the approach to other cases where file paths are loaded the same way.
#PIE-1531

PR #1331, opened 2019-04-25
Queue pending requests when all peers are busy

ajsutton:
## PR description
Modify how idle peers are found so that instead of picking one and then separately sending the request, that's done in a single atomic operation.  This ensures that two threads don't select the same peer and both send requests to it, potentially breaching the outstanding request limit.

Additionally, if all suitable peers are busy queue the request and perform it when a peer is available instead of throwing NoAvailablePeersException. This avoids aborting pipelines when there is a lot of contention for peers and provides a fairer sharing of peers between the various threads that need to request data.


ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/manager/EthPeersTest.java
```diff
@@ -64,4 +86,187 @@ public void comparesPeersWithTdAndNoHeight() {
...
+
+  private void useAllAvailableCapacity(final EthPeer peer) throws PeerNotConnected {
+    while (peer.hasAvailableRequestCapacity()) {
+      useRequestSlot(peer);
+    }
+    assertThat(peer.hasAvailableRequestCapacity()).isFalse();
+  }
+
+  private void useRequestSlot(final EthPeer peer) throws PeerNotConnected {
+    peer.getNodeData(emptyList());
```
mbaxter: Seems like `getNodeData()` should throw an exception here.  For now, it's probably worth at least sending a non-empty list to this method in order to future-proof these tests.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/PendingPeerRequest.java
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  /**
+   * Register callbacks for when the request is made or
+   *
+   * @param onSuccess handler for when a peer becomes available and the request is sent
+   * @param onError handler for when there is no peer with sufficient height or the request fails to
+   *     send
+   */
+  public void thenEither(
```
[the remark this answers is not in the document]
ajsutton: Works for me.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/manager/EthPeersTest.java
ajsutton: Done.

PR #1336, opened 2019-04-26
JsonRpcError decoding to include message

CjHare:
## PR description
With JsonRpcErrors now having multiple errors with the same code, the message need to be included when decoding the JSON to correctly map the error. (bug identified in EthSigner, fix ported)


CjHare:
@ajsutton an interesting point, I'm not entirely sure (or can't recall) the use cases where the JSON errors are needed decoding by Pantheon ...but as the code was already there, I really just keenon keeping things logical correct and consistent across the PegaSys products 🙂

ajsutton:
Maybe we should try deleting it and see if any tests fail? :)

CjHare:
👍  I'll experiment with deleting it (different PR maybe 🤔)

PR #1338, opened 2019-04-26
PAN-2580 - missing p2p info when queried live 

shemnon:
## PR description

It turns out we don't have the p2p host and port at RPC initialization.
To fix this change to query hosts and ports at method invocaiton time
instead of whent he constructor is called.

(cherry picked from commit 105dc5b51e50c8dff7f9af9bcb7c29312c5a6dbd, #1310)

## Fixed Issue(s)

PAN-2580


shemnon:
It has unit tests, but it's the integration that broke.  Should I add new tests that validate this first?

macfarla:
> It has unit tests, but it's the integration that broke. Should I add new tests that validate this first?
If you're confident it works, I'd be happy with a JI to add them later

shemnon:
PAN-2589

PR #1342, opened 2019-04-26
[PAN-2585] Prefer EnodeURL over Endpoint

mbaxter:
## PR description
`EnodeURL` is a more complete representation of a node's address than `Endpoint` and many API's are being built around `EnodeURL`.  To simplify the code base, this PR tries to remove and contain the use of `Endpoint` in favor of `EnodeURL` by:
- Moving `Endpoint` into the discovery package, as it is really an object that is specific to discovery message parsing.
- Move discovery-specific serialization logic from various `Peer` classes to `DiscoveryPeer`
- Removing `getEndpoint()` from `Peer` interface.
- Try to build peers from `EnodeURL` objects so we can leverage the `EnodeURL.Builder` rather than duplicating logic across many constructors (removed a lot of `DefaultPeer` constructors).

A few tangential changes are included as well:
- URI parsing was moved from `DefaultPeer` to `EnodeURL` and replaces the existing String-based URI parsing code.
- `EnodeURL.getIp()` returns an `InetAddress` and `getIpAsString()` was added to support the cases where a string is required.
- `EnodeURL.getListeningPort()` returns a primitive `int` rather than an `Integer`.
- The default P2P_PORT constant was moved to `DefaultCommandValues` as there's really no case where the p2p logic should assume a default port.
- Update `PeerDiscoveryController` to extract and deduplicate peers from neighbors packets before passing those peers on to `RecursivePeerRefreshState`.


mbaxter:
Closing while I debug failing acceptance tests

PR #1343, opened 2019-04-26
[PAN-2588] Create P2PNetwork Builder

mbaxter:
## PR description
Create a `P2PNetwork` builder.  Using this builder pattern simplifies tests and should make it simpler to update the `P2PNetwork` in the future.

This PR also:
- creates a `network` package
- moves the existing `netty` package to `network.netty`
- splits up the existing `NettyP2PNetworkTest` into two tests: one that depends on the concrete implementation details of the network and one that doesn't
- rename `NettyP2PNetwork` to `DefaultP2PNework`
- configure `PeerRequirement`'s on the `PeerDiscoveryAgent` instance rather than passing the requirements to the constructor - this allows us to construct the discovery agent outside of the `P2PNetwork` and pass it in as a dependency

### Reviewer's Notes
To review, it may be easier to look through individual commits in order to view changes without the noise of the file reorganization.


ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/api/P2PNetwork.java
```diff
@@ -34,6 +36,14 @@
    */
   Collection<PeerConnection> getPeers();
 
+  /**
+   * Returns a stream of peers that have been discovered on the network. These peers are not
+   * necessarily connected.
+   *
+   * @return A stream of discovered peers on the network.
+   */
+  Stream<DiscoveryPeer> getDiscoveredPeers();
```
shemnon: I'm not sure a `Stream` should be returned as a property.  Streams are single use and have transient state, and mutations are not propagated back into the returning object. Could this be renamed `streamDiscoveredPeers`?  This may need another PR to push this `getFoo` -> `streamFoo` done properly, but the name would then reflect reality.  

Another option is to return the backing collection, but that looks to be an even deeper refactor to do properly.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/PeerDiscoveryAgent.java
```diff
@@ -70,7 +71,7 @@
   private static final long PEER_REFRESH_INTERVAL_MS = MILLISECONDS.convert(30, TimeUnit.MINUTES);
 
   protected final List<DiscoveryPeer> bootstrapPeers;
-  private final PeerRequirement peerRequirement;
+  private final List<PeerRequirement> peerRequirements = new CopyOnWriteArrayList<>();
```
shemnon: :smile: `CopyOnWriteArray` is very concurrent friendly.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/VertxPeerDiscoveryAgent.java
```diff
@@ -58,21 +61,33 @@ public VertxPeerDiscoveryAgent(
...
         config,
-        peerRequirement,
         peerBlacklist,
         nodeWhitelistController,
         nodePermissioningController,
         metricsSystem);
     checkArgument(vertx != null, "vertx instance cannot be null");
     this.vertx = vertx;
+
+    metricsSystem.createIntegerGauge(
```
shemnon: Just to clarify, is the VertxPeerDiscoveryAgent created once and only once per server?  If not the `createIntegerGauge` call will result in errors on the second and subsequent calls.  Not sure there is any way to hedge this.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/network/DefaultP2PNetwork.java
```diff
@@ -741,4 +716,140 @@ private void onConnectionEstablished(final PeerConnection connection) {
     connections.registerConnection(connection);
     connectCallbacks.forEach(callback -> callback.accept(connection));
   }
+
+  public static class Builder {
```
shemnon: Not for this PR, but we should consider using a Builder library for these in the future to cut down on busy-work code.  Lombok and Guava autoValues both have support for the builder pattern like we use it.  Guava has the advantage we use part of it in our code already.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/api/P2PNetwork.java
mbaxter: Up to now, we haven't used any special naming for streams.  But we can discuss offline and maybe create a follow-up PR.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/VertxPeerDiscoveryAgent.java
mbaxter: Yes, currently only created once.

PR #1344, opened 2019-04-26
Enable the pipeline chain downloader by default

ajsutton:
## PR description
The pipeline based downloader is proving to be more stable in low memory conditions and slightly faster than the EthTask based one so switch to the pipeline chain downloader by default for full sync.

Leaving the EthTask based downloader as an option for now but will follow up with a future PR to remove it entirely.


acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/ProcessPantheonNodeRunner.java
```diff
@@ -257,7 +258,7 @@ public boolean isActive(final String nodeName) {
   private void killPantheonProcess(final String name, final Process process) {
     LOG.info("Killing " + name + " process");
 
-    Awaitility.waitAtMost(30, TimeUnit.SECONDS)
+    Awaitility.waitAtMost(60, TimeUnit.SECONDS)
```
mbaxter: Did you mean to keep this duration change?

ajsutton: Yeah it seems to be slower to shutdown in CI than previously.  The request open ended headers fix and this increase were needed to get tests passing reliably but I wasn't keen on merging with that many changes after the original approval.

ajsutton: For the record, shutdown delay was found and fixed in another PR so this duration change wasn't needed.

PR #1349, opened 2019-04-28
Update Bootnodes.md

jacobhackman:
Touched up language slightly.


docs/Configuring-Pantheon/Networking/Bootnodes.md
```diff
@@ -9,13 +9,13 @@ Bootnodes are used to initially discover peers.
 
 ## Mainnet and Public Testnets
 
-For mainnet, Rinkeby, Ropsten, and Görli, Pantheon predefines a list of enode URLs.  
+For mainnet, Rinkeby, Ropsten, and Görli, Pantheon predefine a list of enode URLs.
```
MadelineMurray: I think this should stay plural because Pantheon is singular.  Compare it with: 
Bob and Mary predefine a list. 
Bob predefines a list.

jacobhackman: You're totally right. I see it now.

PR #1372, opened 2019-04-30
Handle the pipeline being aborted while finalizing an async operation

ajsutton:
## PR description
When an async processing stage of a pipeline has started the async process for all inputs it enters finalisation mode where it waits for the outstanding tasks to complete.  If Pantheon is shutdown while this finalisation is in progress it may delay shutdown because it doesn't check for the abort signal.

This changes the finalisation process so that when the thread is interrupted (or a task completes) it checks if the input pipe has been aborted and aborts finalisation to ensure we get a clean shutdown.


services/pipeline/src/test/java/tech/pegasys/pantheon/services/pipeline/ProcessingStageTest.java
```diff
@@ -68,22 +70,42 @@ public void shouldCallSingleStepStageForEachInput() {
...
+
+  @Test
+  public void shouldAbortIfPipeIsCancelledWhileAttemptingToFinalise() {
+    when(singleStep.attemptFinalization(outputPipe))
+        .thenAnswer(
+            invocation -> {
+              inputPipe.abort();
+              return false;
+            });
+    inputPipe.close();
```
mbaxter: Why do you need to close the inputPipe here?

ajsutton: Because otherwise the processor won't enter the finalisation stage, it will wait forever for the next input.

PR #1373, opened 2019-04-30
Remove EthTaskChainDownloader and supporting code

ajsutton:
## PR description
Remove the old EthTaskChainDownloader class and code that is only used by it (plus a few other bits of unused code detected by the static analysis).

We should merge #1372 and #1344 first.


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/EthereumWireProtocolConfiguration.java
```diff
@@ -68,15 +70,23 @@ public int getMaxGetNodeData() {
...
-        && maxGetNodeData == other.maxGetNodeData;
+    final EthereumWireProtocolConfiguration that = (EthereumWireProtocolConfiguration) o;
+    return maxGetBlockHeaders == that.maxGetBlockHeaders &&
+        maxGetBlockBodies == that.maxGetBlockBodies &&
+        maxGetReceipts == that.maxGetReceipts &&
+        maxGetNodeData == that.maxGetNodeData;
+  }
+
+  @Override
+  public int hashCode() {
```
abdelhamidbakhta: Optional. Maybe we could test equals and hashCode. Although it is non blocking.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/EthereumWireProtocolConfiguration.java
```diff
@@ -91,9 +92,12 @@ public int hashCode() {
 
   @Override
   public String toString() {
-    return String.format(
-        "maxGetBlockHeaders=%s\tmaxGetBlockBodies=%s\tmaxGetReceipts=%s\tmaxGetReceipts=%s",
-        maxGetBlockHeaders, maxGetBlockBodies, maxGetReceipts, maxGetNodeData);
+    return MoreObjects.toStringHelper(this)
```
abdelhamidbakhta: Optional. Unit test.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/EthereumWireProtocolConfiguration.java
ajsutton: Equals is only used by tests and hash code isn't used (but should exist to match equals just for good measure).  I really only updated them to match our code style which is the generated code.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/EthereumWireProtocolConfiguration.java
ajsutton: toString is always just for debugging purposes so the only time we should test it is when there's sensitive info that definitely shouldn't appear in a log (ie anything containing a password).

PR #1376, opened 2019-04-30
[PAN-2595] Consolidate local enode representation

mbaxter:
## PR description
This PR consolidates local enode representations around a single canonical representation.  Previously, there were a few places where we would construct `EnodeURL` instances to represent the local node.  However, this representation is only truly accurate once the network starts up as ports can be chosen on startup.


ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/transactions/TestNodeList.java
```diff
@@ -59,12 +59,6 @@ public TestNode create(
     return node;
   }
 
-  public void startNetworks() {
```
mbaxter: Each individual network is already started [here](https://github.com/PegaSysEng/pantheon/blob/e95d0681450b12df36102d16c1b99c02ea2aca2f/ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/transactions/TestNode.java#L158)

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/InsufficientPeersPermissioningProvider.java
```diff
@@ -21,13 +21,14 @@
...
 import java.util.Optional;
+import java.util.function.Supplier;
 
 /**
  * A permissioning provider that only provides an answer when we have no peers outside of our
  * bootnodes
  */
 public class InsufficientPeersPermissioningProvider implements ContextualNodePermissioningProvider {
-  private final EnodeURL selfEnode;
+  private final Supplier<Optional<EnodeURL>> selfEnode;
```
mbaxter: Previously, we passed in a static `EnodeURL` constructed in `RunnerBuilder`.  This enode wasn't 100% accurate as the network can be configured to choose a port on startup, and the representation constructed in the builder would not accurately reflect the real ports being used.  We're now passing a supplier that returns the canonical local enode representation when available.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/InsufficientPeersPermissioningProvider.java
```diff
@@ -64,17 +65,23 @@ private long countP2PNetworkNonBootnodeConnections() {
   @Override
   public Optional<Boolean> isPermitted(
       final EnodeURL sourceEnode, final EnodeURL destinationEnode) {
+    Optional<EnodeURL> maybeSelfEnode = selfEnode.get();
     if (nonBootnodePeerConnections > 0) {
       return Optional.empty();
-    } else if (checkEnode(sourceEnode) && checkEnode(destinationEnode)) {
+    } else if (!maybeSelfEnode.isPresent()) {
+      // The local node is not yet ready, so we can't validate enodes yet
+      return Optional.empty();
```
mbaxter: We'll now only return an answer when the network is ready.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/api/P2PNetwork.java
```diff
@@ -77,21 +77,23 @@
...
+   * regardless of maxPeer limits. Non-permitted peers may be added to this list, but will not
+   * actually be connected to as long as they are prohibited.
    *
    * @param peer The peer that should be connected to
-   * @return boolean representing whether or not the peer has been added to the list or was already
-   *     on it
+   * @return boolean representing whether or not the peer has been added to the list, false is
+   *     returned if the peer was already on the list
    */
   boolean addMaintainConnectionPeer(final Peer peer);
```
mbaxter: Previously, peer permission would be checked as peers were added to this list.  The changes in this PR delay the point at which we're able to check permission to the point where our local node is up and running.  As a result, if we were to gate entrance into the list based on permissions, there would be a period where no nodes could be added to this list while the node is starting up.  So, this API has been updated such that a peer can be added to the "maintain connection" list whether or not it is permitted on the network.  However, no peers will actually be connected to without first checking permissions.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/network/DefaultP2PNetwork.java
```diff
@@ -346,7 +347,7 @@ protected void initChannel(final SocketChannel ch) {
                 return;
               }
 
-              if (!isPeerConnectionAllowed(connection)) {
+              if (!isPeerAllowed(connection)) {
```
mbaxter: Permission checks have been centralized to one method.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/network/DefaultP2PNetwork.java
```diff
@@ -394,12 +387,11 @@ public boolean removeMaintainedConnectionPeer(final Peer peer) {
...
-  public void checkMaintainedConnectionPeers() {
-    for (final Peer peer : peerMaintainConnectionList) {
-      if (!(isConnecting(peer) || isConnected(peer))) {
-        connect(peer);
-      }
-    }
+  void checkMaintainedConnectionPeers() {
+    peerMaintainConnectionList.stream()
+        .filter(p -> !isConnectingOrConnected(p))
+        .filter(this::isPeerAllowed)
```
mbaxter: Check permissions before connecting :\

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/network/DefaultP2PNetwork.java
```diff
@@ -529,33 +521,36 @@ public void subscribeDisconnect(final DisconnectCallback callback) {
...
-    peerDroppedObserverId =
-        OptionalLong.of(peerDiscoveryAgent.observePeerDroppedEvents(handlePeerDroppedEvents()));
-
-    if (nodePermissioningController.isPresent()) {
-      if (blockchain.isPresent()) {
-        synchronized (this) {
-          if (!blockAddedObserverId.isPresent()) {
-            blockAddedObserverId =
-                OptionalLong.of(blockchain.get().observeBlockAdded(this::handleBlockAddedEvent));
+    if (started.compareAndSet(false, true)) {
```
mbaxter: Start can be called at most once

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/network/DefaultP2PNetwork.java
```diff
@@ -595,38 +590,36 @@ private synchronized void checkCurrentConnections() {
...
+    if (peerBlacklist.contains(enode.getNodeId())) {
+      return false;
+    }
+    if (enode.getNodeId().equals(ourPeerInfo.getNodeId())) {
+      // Peer matches our node id
       return false;
     }
 
+    Optional<EnodeURL> maybeEnode = getLocalEnode();
+    if (!maybeEnode.isPresent()) {
```
mbaxter: Be restrictive around permissions until we have a valid representation of our local enode.

ethereum/p2p/src/test/java/tech/pegasys/pantheon/ethereum/p2p/network/DefaultP2PNetworkTest.java
```diff
@@ -103,17 +104,19 @@ public void closeVertx() {
...
   public void addingMaintainedNetworkPeerStartsConnection() {
     final DefaultP2PNetwork network = mockNetwork();
+    network.start();
     final Peer peer = mockPeer();
 
     assertThat(network.addMaintainConnectionPeer(peer)).isTrue();
 
     assertThat(network.peerMaintainConnectionList).contains(peer);
-    verify(network, times(1)).connect(peer);
+    verify(network, atLeast(1)).connect(peer);
```
mbaxter: The function `peerMaintainConnectionList` is now called shortly after startup so that that we'll initiate connections to peers that were added before startup (we were previously initiating connections as peers were added in the [`RunnerBuilder`](https://github.com/PegaSysEng/pantheon/blob/e95d0681450b12df36102d16c1b99c02ea2aca2f/pantheon/src/main/java/tech/pegasys/pantheon/RunnerBuilder.java#L317)).  Update these checks to be looser in case the startup check runs in addition to the manual invocation.

ethereum/p2p/src/test/java/tech/pegasys/pantheon/ethereum/p2p/network/DefaultP2PNetworkTest.java
```diff
@@ -122,13 +124,29 @@ public void addingRepeatMaintainedPeersReturnsFalse() {
...
+
     final Peer peer = mockPeer();
     network.peerMaintainConnectionList.add(peer);
 
     network.checkMaintainedConnectionPeers();
     verify(network, times(1)).connect(peer);
   }
 
+  @Test
+  public void checkMaintainedConnectionPeersDoesNotConnectToDisallowedPeer() {
```
ajsutton: @lucassaldanha can you confirm that the permissioning controls should prevent peers added via staticnodes.json or admin_addPeer from connecting?  It makes sense to  me but I seem to recall some discussions and details...

lucassaldanha: For the staticnodes.json list we apply the same rules as for bootnodes: they must be permitted.

For admin_addPeer is the same logic as connecting to another peer. If the peer is not allowed it won't connect to the peer (and I believe the addPeer method will give you a nice error response saying why).

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/network/DefaultP2PNetwork.java
lucassaldanha: Shouldn't we drop them from the maintained list if they aren't permitted?

ajsutton:
I'd be inclined to keep forbidden peers in the maintained list because they may become permitted later - especially with on-chain permissioning.  You may have setup a new node, added it as a peer for your existing nodes and are now getting approvals from the consortium to allow it to actually connect.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/network/DefaultP2PNetwork.java
Errorific: I don't think so. The peer might later be permitted.

ethereum/p2p/src/test/java/tech/pegasys/pantheon/ethereum/p2p/network/DefaultP2PNetworkTest.java
mbaxter: The "admin_addPeer" method will no longer return error messages if a non-permitted peer is added.  Discussed this with @lucassaldanha offline and we determined that there is no need to keep these error messages.

PR #1377, opened 2019-05-01
Add flat file based task collection

ajsutton:
## PR description
Add an alternative implementation of `TaskCollection` which uses a simple file as an alternative to the RocksDB based one.


ajsutton:
Worth noting a major concern at this point is how big the file will get. I don't have data on that but suspect it should be changed to move to a new file when the size reaches a certain point.  Once the reader moves past the end of the file it can then be deleted which limits the amount of space used.

ajsutton:
Files are now rolled at a certain size and deleted once all tasks are read from them which keeps the disk space under control.

ajsutton:
I really like the idea of annotations to make it clear how things are synchronized but *only* if they are actually enforced as part of our CI process.  Otherwise they'll wind up lying which is worse than not having them.  I've added a ticket to investigate further.

PR #1378, opened 2019-05-01
[PAN-2585] Prefer EnodeURL over Endpoint

mbaxter:
## PR description
`EnodeURL` is a more complete representation of a node's address than `Endpoint` and many API's are being built around `EnodeURL`.  To simplify the code base, this PR tries to remove and contain the use of `Endpoint` in favor of `EnodeURL` by:
- Moving `Endpoint` into the discovery package, as it is really an object that is specific to discovery message parsing.
- Move discovery-specific serialization logic from various `Peer` classes to `DiscoveryPeer`
- Removing `getEndpoint()` from `Peer` interface.
- Try to build peers from `EnodeURL` objects so we can leverage the `EnodeURL.Builder` rather than duplicating logic across many constructors (removed a lot of `DefaultPeer` constructors).

A few tangential changes are included as well:
- URI parsing was moved from `DefaultPeer` to `EnodeURL` and replaces the existing String-based URI parsing code.
- `EnodeURL.getIp()` returns an `InetAddress` and `getIpAsString()` was added to support the cases where a string is required.
- `EnodeURL.getListeningPort()` returns a primitive `int` rather than an `Integer`.
- The default P2P_PORT constant was moved to `DefaultCommandValues` as there's really no case where the p2p logic should assume a default port.
- Update `PeerDiscoveryController` to extract and deduplicate peers from neighbors packets before passing those peers on to `RecursivePeerRefreshState`.


ethereum/p2p/src/test/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/PeerDiscoveryControllerTest.java
```diff
@@ -718,8 +716,8 @@ public void shouldNotBondWithBlacklistedPeer() {
         MockPacketDataFactory.mockNeighborsPacket(discoPeer, otherPeer, otherPeer2);
     controller.onMessage(neighborsPacket, discoPeer);
 
-    verify(controller, times(0)).bond(otherPeer);
-    verify(controller, times(1)).bond(otherPeer2);
+    verify(controller, times(0)).bond(eq(otherPeer));
+    verify(controller, times(1)).bond(eq(otherPeer2));
```
ajsutton: nit: I don't think we need the `eq` here.  The default is equals comparison so you typically only use `eq` if you need to use other matchers for some params (because all params must use matchers if any do).

util/src/main/java/tech/pegasys/pantheon/util/enode/EnodeURL.java
```diff
@@ -25,19 +25,14 @@
...
-      "^enode://"
-          + "(?<nodeId>\\w+)"
-          + "@"
-          + "(?<ip>.*)"
-          + ":"
-          + "(?<listening>\\d+)"
-          + "(\\?discport=(?<discovery>\\d+))?$";
+  private static final int NODE_ID_SIZE = 64;
+  private static final Pattern DISCPORT_QUERY_STRING_REGEX =
+      Pattern.compile("discport=([0-9]{1,5})");
```
ajsutton: If we're being really strict this should probably be `"(^|&)discport=([0-9]{1,4})"` I think. So discport should either be the first arg or after a & to prevent matching `enode://...@...:30303?notthediscport=1234`

util/src/main/java/tech/pegasys/pantheon/util/enode/EnodeURL.java
```diff
@@ -103,60 +138,23 @@ public static URI asURI(final String url) {
...
     return ip.getHostAddress();
   }
 
+  public InetAddress getIp() {
+    return ip;
+  }
+
   public InetAddress getInetAddress() {
     return ip;
   }
```
ajsutton: We probably shouldn't have both `getIp` and `getInetAddress` both returning the same thing.

ethereum/p2p/src/test/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/PeerDiscoveryControllerTest.java
mbaxter: Removed

util/src/main/java/tech/pegasys/pantheon/util/enode/EnodeURL.java
mbaxter: Made this check stricter and added more tests.  For now, I have the parsing fail if there are any additional params outside of a singular discport param.  It's simpler to implement this way and if we need additional params for some reason we can add that later.

util/src/main/java/tech/pegasys/pantheon/util/enode/EnodeURL.java
mbaxter: Good call - not sure how I missed this since the method is right below the one I added 😂

PR #1391, opened 2019-05-02
Handle case where peers advertise a listening port of 0

mbaxter:
## PR description
It is common for peers to advertise a listening port of 0.  Handle this case by using a default listening port of 30303.

### DIFF
--- a/ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/network/netty/DeFramer.java
+++ b/ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/network/netty/DeFramer.java
@@ -168,12 +168,17 @@ final class DeFramer extends ByteToMessageDecoder {
   }
 
   private Peer createPeer(final PeerInfo peerInfo, final ChannelHandlerContext ctx) {
-    InetSocketAddress remoteAddress = ((InetSocketAddress) ctx.channel().remoteAddress());
+    final InetSocketAddress remoteAddress = ((InetSocketAddress) ctx.channel().remoteAddress());
+    int port = peerInfo.getPort();
+    if (port == 0) {
+      // Most peers advertise a port of "0", just set a default best guess in this case
+      port = EnodeURL.DEFAULT_LISTENING_PORT;
+    }
     return DefaultPeer.fromEnodeURL(
         EnodeURL.builder()
             .nodeId(peerInfo.getNodeId())
             .ipAddress(remoteAddress.getAddress())
-            .listeningPort(peerInfo.getPort())
+            .listeningPort(port)
             .build());
   }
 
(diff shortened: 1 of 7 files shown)


ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/network/netty/DeFramer.java
```diff
@@ -168,12 +168,17 @@ protected void decode(final ChannelHandlerContext ctx, final ByteBuf in, final L
   }
 
   private Peer createPeer(final PeerInfo peerInfo, final ChannelHandlerContext ctx) {
-    InetSocketAddress remoteAddress = ((InetSocketAddress) ctx.channel().remoteAddress());
+    final InetSocketAddress remoteAddress = ((InetSocketAddress) ctx.channel().remoteAddress());
+    int port = peerInfo.getPort();
+    if (port == 0) {
+      // Most peers advertise a port of "0", just set a default best guess in this case
+      port = EnodeURL.DEFAULT_LISTENING_PORT;
```
shemnon: Instead of the default port perhaps we could use the remote port of the current communication?  We are already doing this with the remoteAddress.

Also, can these two values ever differ?  Such as if we have an eth loadbalancer or proxy in front of us (in case those ever exist for devp2p).

mbaxter: I experimented with trying to use the remote port, but it doesn't match the listening port.  When the other nodes are initiating outbound connections, they're free to use whatever port they want for their side of the connection.

mbaxter: Un-resolving this comment because I think its a good question and will be helpful to have it visible here as a kind of documentation.  Actually, I think I might add a comment in the codebase as well :D

xavierlepretre:
As a side note, my Geth reports its own `admin.nodeInfo.enode` as `"enode://3dffa8485a02fcb969e8afa47ba7134c785c3b04be00086907b3766df7280dfe1b9d38813be7f7f90043215c9fffca5f7778e6a1c403fb4cac087c13b908a9b5@127.0.0.1:30303?discport=0"`. If I don't truncate the `?discport=0` before I add the string to Pantheon's `bootnodes`, then Pantheon complains that:

```
picocli.CommandLine$ParameterException: Invalid enode URL syntax. Enode URL should have the following format 'enode://<node_id>@<ip>:<listening_port>[?discport=<discovery_port>]'. Invalid discovery port.  Port should be between 1 - 65535.
	at tech.pegasys.pantheon.cli.PantheonCommand.setBootnodes(PantheonCommand.java:187)
```

mbaxter:
Thanks for reporting @xavierlepretre !  I've added a ticket and we will address this soon: https://pegasys1.atlassian.net/browse/PAN-2618

PR #1399, opened 2019-05-03
Fix non-deterministic test caused by variable size of generated transactions

abdelhamidbakhta:
## PR description
Fix non-deterministic test caused by variable size of generated transactions.  Validate transactions against a size range rather than a fixed size
## Fixed Issue(s)

fixes #PAN-2617


ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/messages/LimitedTransactionsMessagesTest.java
```diff
@@ -40,12 +40,12 @@
   public void createLimited() {
     final Set<Transaction> txs = generator.transactions(6000);
     final LimitedTransactionsMessages firstMessage = LimitedTransactionsMessages.createLimited(txs);
-    assertEquals(5219, firstMessage.getIncludedTransactions().size());
+    assertBetween(firstMessage.getIncludedTransactions().size(), 4990, 5250);
```
shemnon: Use libraries instead of rolling our own.  Something like 
```
org.assertj.core.api.Assertions.assertThat(firstMessage.getIncludedTransactions().size()).isBetween(4990, 5250);
```
I've also found that the asserj libraries, because of their fluent API, can do more interesting checks.

abdelhamidbakhta: Ok. Fixed

PR #1401, opened 2019-05-04
[PAN-2586] Reference tests fail on Windows

smatthewenglish:
## PR description

Resolves path issues with running on Windows. 

At the moment it seems we aren't evaluating `VMTests`, `TrieTests`, or `TransitionTests` -- why is that?

## Fixed Issue(s)

[![enter image description here][1]][1]

  [1]: https://i.stack.imgur.com/CVmCy.png

### DIFF
--- a/ethereum/core/build.gradle
+++ b/ethereum/core/build.gradle
@@ -91,26 +91,32 @@ test {
   exclude 'tech/pegasys/pantheon/ethereum/core/TransactionTest.class'
 }
 
-def generateTestFiles(FileTree jsonPath, File resourcesPath, File templateFile, String pathstrip, String destination, String namePrefix) {
+def generateTestFiles(FileTree jsonPath, File templateFile, String pathstrip, String destination, String namePrefix) {
   def referenceTestTemplate = templateFile.text
 
   // This is how many json files to include in each test file
   def fileSets = jsonPath.getFiles().collate(5)
 
   fileSets.eachWithIndex { fileSet, idx ->
-    def resPath = resourcesPath.getPath().replaceAll("\\\\", "/")
-
     def paths = []
     fileSet.each { testJsonFile ->
       if (!testJsonFile.getName().toString().startsWith(".")) {
-        paths << testJsonFile.getPath().toString()
-          .replaceAll(resPath + "/", "")
+        String parentDirectory = testJsonFile.getParentFile().getName() + "/"
+        String testFile = testJsonFile.getName()
+        if(pathstrip == "BlockchainTests/"){
+          String generalStateTests = "GeneralStateTests/"
+          if(testJsonFile.getParentFile().getParentFile().getName().toString() == generalStateTests){
+            paths << pathstrip + generalStateTests + parentDirectory + testFile
+          }
+        } else {
+          paths << pathstrip + parentDirectory + testFile
+        }
       }
     }
 
     def testFile = file(destination + "/" + namePrefix + "_" + idx + ".java")
 
-    def allPaths = '"' + paths.join('", "') + '"';
+    def allPaths = '"' + paths.join('", "') + '"'
 
     def testFileContents = referenceTestTemplate
       .replaceAll("%%TESTS_FILE%%", allPaths)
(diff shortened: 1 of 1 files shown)


ethereum/core/build.gradle
```diff
@@ -91,26 +91,32 @@ test {
...
     def paths = []
     fileSet.each { testJsonFile ->
       if (!testJsonFile.getName().toString().startsWith(".")) {
-        paths << testJsonFile.getPath().toString()
-          .replaceAll(resPath + "/", "")
+        String parentDirectory = testJsonFile.getParentFile().getName() + "/"
+        String testFile = testJsonFile.getName()
+        if(pathstrip == "BlockchainTests/"){
+          String generalStateTests = "GeneralStateTests/"
+          if(testJsonFile.getParentFile().getParentFile().getName().toString() == generalStateTests){
```
shemnon: This condition will never file.  The name of the parent file will never end in a path separator ("/").  You need to test for "GeneralStateTests" and then concat "GeneralStateTests/"

smatthewenglish: nice catch. thanks. this was a result of my overzealous beautification, I've reimplemented your proposed changes and run it again on windows to ensure that it's got the right test coverage

ethereum/core/build.gradle
```diff
@@ -91,26 +91,36 @@ test {
...
-          .replaceAll(resPath + "/", "")
+        String parentDirectory = testJsonFile.getParentFile().getName() + "/"
+        String testFile = testJsonFile.getName()
+        if(pathstrip == "BlockchainTests/"){
+          String generalStateTests = "GeneralStateTests"
+          if(testJsonFile.getParentFile().getParentFile().getName().toString() == generalStateTests){
+            paths << pathstrip + generalStateTests + "/" + parentDirectory + testFile
+          }
+          String transitionTests = "TransitionTests"
+          if(testJsonFile.getParentFile().getParentFile().getName().toString() == transitionTests){
```
shemnon: It looks like we are losing all tests that are not `TransitionTests` or `GeneralStateTests`, such as `bcInvalidHeaderTest`.  Instead of doing a one-off for each test directory can these two if statements be makde to work for any parent name?  maybe - 

```
paths << parentstrip + testJsonFile?.parentFile?.parentFile?.name + "/" + parentDirectory + testFile
```

smatthewenglish: I can see why you would say that, but that's not how it works -- Transition and General State are themselves subdirectories of Blockchain Test - both of which contain still other directories, the only ones like that, so if the test came from one of those two folders we need to add that to the path. You'll see what I mean if you check out the project structure there

shemnon: I have checked out the project structure, which is why I know this PR results in us not running all the tests.

Check out `tech.pegasys.pantheon.ethereum.vm.blockchain.BlockchainReferenceTest_0` line 21 used to read

```
  private static final String[] TEST_CONFIG_FILE_DIR_PATH = new String[] {"BlockchainTests/bcForgedTest/bcForkBlockTest.json", "BlockchainTests/bcForgedTest/bcForkUncle.json", "BlockchainTests/bcForgedTest/bcInvalidRLPTest.json", "BlockchainTests/bcMultiChainTest/CallContractFromNotBestBlock.json", "BlockchainTests/bcMultiChainTest/UncleFromSideChain.json"};
```

It now reads

```
  private static final String[] TEST_CONFIG_FILE_DIR_PATH = new String[] {""};
```

We are missing many subdirectories: `bcBlockGasLimitTest`, `bcExploitTest`, and 12 others. i.e. too many to reasonable enumerate with if blocks like this. Plus, if a new directories were to be added we would miss them if we only take enumerated directories.  We should _continue_ to take all the subdirectories of `BlockchainTest` as discovered and not as enumerated.

smatthewenglish: If you look in the directory `BlockchainTests/`, there are only *two* subdirectories that contain directories themselves as opposed to -- just a bunch of tests. 

Those directories are `pantheon/ethereum/referencetests/src/test/resources/BlockchainTests/GeneralStateTests` and `pantheon/ethereum/referencetests/src/test/resources/BlockchainTests/TransitionTests`.


You mentioned `bcBlockGasLimitTest `, see here: 

[![enter image description here][2]][2]

it follows the pattern `paths << pathstrip + "/" + parentDirectory + "/" + testFile`

However, if you look here:

[![enter image description here][1]][1]

we see the pattern `paths << pathstrip + "/" + testJsonFile.getParentFile().getParentFile().getName() + "/" + parentDirectory + "/" + testFile`, for tests inside of subdirectories of `pantheon/ethereum/referencetests/src/test/resources/BlockchainTests/GeneralStateTests`.

---

The point about not hardcoding those two directories is valid, I've updated the pr to be more dynamic. 



  [1]: https://i.stack.imgur.com/YpNtA.png
  [2]: https://i.stack.imgur.com/anAyq.png

smatthewenglish:
[![enter image description here][1]][1]


  [1]: https://i.stack.imgur.com/f3Dyn.png

confirmed that latest update passes on windows

PR #1404, opened 2019-05-07
Add constantinople fix block for Rinkeby

ajsutton:
## PR description
The ConstantinopleFix hardfork has been scheduled for Rinkeby (and has actually already happened).  Update our Rinkeby config.


ajsutton:
Announced in the geth release notes, in the geth discord, in Péter Szilágyi's twitter and reddit.  I've suggested on twitter that things like this would be worth mentioning on all core devs.  Theoretically Rinkeby is a geth testnet but there's a few clients that can sync to it now.

We should pay more attention to geth release notes in the mean time.

PR #1418, opened 2019-05-08
[PAN-2624] Start listening for p2p connections after start() is invoked

mbaxter:
## PR description

Start listening for incoming p2p connections only after start() is invoked.  Additionally, since we're no longer setting up listening in the constructor, we can now gate the stop() method so that it only runs after the network has been started.

### DIFF
--- a/ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/network/DefaultP2PNetwork.java
+++ b/ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/network/DefaultP2PNetwork.java
@@ -187,6 +187,7 @@ public class DefaultP2PNetwork implements P2PNetwork {
   private OptionalLong blockAddedObserverId = OptionalLong.empty();
 
   private final AtomicBoolean started = new AtomicBoolean(false);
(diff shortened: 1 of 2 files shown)


ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/network/DefaultP2PNetwork.java
```diff
@@ -242,14 +257,22 @@
...
+  private Supplier<Integer> pendingTaskCounter(final EventLoopGroup eventLoopGroup) {
+    return () ->
+        StreamSupport.stream(eventLoopGroup.spliterator(), false)
+            .filter(eventExecutor -> eventExecutor instanceof SingleThreadEventExecutor)
+            .mapToInt(eventExecutor -> ((SingleThreadEventExecutor) eventExecutor).pendingTasks())
+            .sum();
+  }
 
+  /** Start listening for incoming connections */
+  private void startListening() {
```
mbaxter: This diff has ended up looking pretty confusing.  I'm just moving this listening logic out of the constructor and into a separate method.  I also did some reordering of the constructor logic so that it reads better.

PR #1423, opened 2019-05-09
Unwrap DelegatingBytes32 and prevent Hash from wrapping other Hash instances

ajsutton:
## PR description
When blocks are generated without any change to the world state (only possible with empty blocks in Clique or IBFT where there's no block reward), the world state root hash was being repeatedly wrapped in a new `Hash` instance until eventually it triggered a stack overflow.  The same `Hash` instance survives because the actual `BlockHeader` instance is cached by the blockchain and the state root hash round trips through the block creator, world state Merkle trie entirely unchanged except for being wrapped an additional time.

Fix is to apply the same `unwrap` process to `DelegatingBytes32` as already existed in `DelegatingBytesValue`.  To save creating a new object at all a check has also been added to `Hash.wrap` so that it just casts the existing value if it's already a `Hash`.


util/src/main/java/tech/pegasys/pantheon/util/bytes/DelegatingBytes32.java
```diff
@@ -14,7 +14,17 @@
...
-    super(wrapped);
+    super(unwrap(wrapped));
+  }
+
+  // Make sure we don't end-up with giant chains of delegating through wrapping.
+  private static Bytes32 unwrap(final Bytes32 v) {
+    Bytes32 wrapped = v;
+
+    if (wrapped instanceof DelegatingBytes32) {
+      wrapped = ((DelegatingBytes32) wrapped).wrapped;
```
rain-on: just return the thing - then at the bottom return v.

PR #1426, opened 2019-05-09
Repair static-nodes post P2P upgrades

rain-on:
With the upgrades done to P2P networking, the "ourpeer" information is not available until after DefaultP2PNetwork has started.

Currently loading static-nodes requires this information to prevent attempting to create static-connections to ourselves - thus static-node loading needs to be delayed until after network start.

The solution in this PR  may not be the optimal way to make this work... open to suggestions.


mbaxter:
My bad for missing this.  Opened a PR with an alternative approach here: https://github.com/PegaSysEng/pantheon/pull/1428

PR #1428, opened 2019-05-09
Fix p2p PeerInfo handling

mbaxter:
## PR description
This PR reworks handling of potentially null / empty values in `DefaultP2PNetwork`.  `PeerInfo` is wrapped in an `Optional` to make it more obvious that this value may be empty.  Some additional checks have been added to shortcut operations that should only execute when the network is fully up and running: 
- Peers are not permitted to connect until the network is fully up
- `connect()` will return an exceptional result if invoked before the network is up


mbaxter:
This is not the BEST rationale, but I set the initial timeout to 2 rather 0 to make tests a little simpler.  There are some tests where an immediately scheduled check interferes because the scheduled check may or may not run during the test.

PR #1429, opened 2019-05-09
rocksDB Options

shemnon:
## PR description

Expose more RocksDB options for db configuration.

Update default options as well.

### DIFF
--- a/services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/RocksDbKeyValueStorage.java
+++ b/services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/RocksDbKeyValueStorage.java
@@ -12,6 +12,9 @@
  */
 package tech.pegasys.pantheon.services.kvstore;
 
+import static org.rocksdb.util.SizeUnit.KB;
+import static org.rocksdb.util.SizeUnit.MB;
+
 import tech.pegasys.pantheon.metrics.Counter;
 import tech.pegasys.pantheon.metrics.MetricCategory;
 import tech.pegasys.pantheon.metrics.MetricsSystem;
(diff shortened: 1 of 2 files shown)


services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/RocksDbKeyValueStorage.java
```diff
@@ -61,12 +64,26 @@ private RocksDbKeyValueStorage(
...
-              .setTableFormatConfig(rocksDbConfiguration.getBlockBasedTableConfig())
+              .setTableFormatConfig(new BlockBasedTableConfig()
+                  .setBlockSize(16 * 1024)
+                  .setPinL0FilterAndIndexBlocksInCache(true)
+                  .setCacheIndexAndFilterBlocks(true) // make this an --X option
+                  .setFilter(new BloomFilter(10, true))
+                  .setFormatVersion(2)
+                  .setBlockCache(new LRUCache(rocksDbConfiguration.getCacheCapacity()))
+              )
+              .optimizeForPointLookup(rocksDbConfiguration.getCacheCapacity() >> 20) // API wants MBs
```
ajsutton: FYI ` * org.rocksdb.util.SizeUnit.MB` would be more readable than `>> 20` here.

shemnon: done

shemnon:
These tuning defaults made the memory situation worse, rospten on 8GB crashed several times by the linux system out of memory checker.  I still want to come back and refactor the generation of the table defaults.  But it would not be a full set of flags like this.

PR #1434, opened 2019-05-10
Equals cleanup

shemnon:
## PR description

* Don't copy collections if we don't need to.  Change types higher up if needed.
* Don't use Guava's Object.equal, use Java's Objects.equals.


consensus/ibft/src/main/java/tech/pegasys/pantheon/consensus/ibft/payload/PreparedCertificate.java
```diff
@@ -68,7 +68,7 @@ public boolean equals(final Object o) {
     }
     final PreparedCertificate that = (PreparedCertificate) o;
     return Objects.equals(proposalPayload, that.proposalPayload)
-        && Objects.equals(new ArrayList<>(preparePayloads), new ArrayList<>(that.preparePayloads));
```
mbaxter: nice cleanup 👍

PR #1437, opened 2019-05-12
Capture all logs and errors in the Pantheon log output

ajsutton:
## PR description
Pantheon uses some libraries that log to SLF4j but we didn't include a binding from SLF4j to Log4J at runtime so those logs were silently dropped.  Include the binding so the libraries log correctly and we don't get an annoying warning about missing the binding at startup.

Additionally, set a default exception handler at the start of our main method so any uncaught exceptions are output to our logging rather than printed to system error which may get missed if a simple redirect is used.


ajsutton:
There was a pass done to use `LogManager.getLogger()` consistently a while back. I think its pretty consistent but there are probably some exceptions.

PR #1446, opened 2019-05-14
[PAN-2614] Add simple PeerPermissions interface

mbaxter:
## PR description
This PR adds a simple `PeerPermissions` interface as a first step in trying to define a common interface for permission-related logic within the p2p package.  

Summary of changes:
* Create a basic `PeerPermissions` interface
* Break up `PeerBlacklist` into several smaller classes and put a subset of these classes behind the new interface
  * Create a `PeerPermissionsBlacklist` class that just holds a set of peers that are not allowed.  The list can be optionally capped with a max number of tracked peers.
  * Create a `LimitedSet` class to handle creating a capped set
  * Move logic for blacklisting peers based on disconnect into a `PeerReputationManager` class
  * Separate out lists of nodes that are banned because of bad behavior vs banned via the CLI
  * Delete a bunch of duplicate tests in the p2p test suite that were validating logic that now belongs in `PeerReputationManager`
* Validate that arguments to `--banned-node-ids` are valid node ids
* Don't indiscriminately drop peers from our peer table for any disconnect
  * Instead, drop peers when they are not allowed
  * Since we're no longer dropping peers based on any disconnect, when pulling discovery peers to connect to, sort them so that we start with the least recently contacted peers first
* Delete dead code related to tracking dropped peer events
* Add a `PeerDiscoveryController` Builder to make updates to the long list of constructor dependencies easier
* Validate peers before initiating outbound connections


acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/ThreadPantheonNodeRunner.java
```diff
@@ -108,7 +107,6 @@ public void startNode(final PantheonNode node) {
             .jsonRpcConfiguration(node.jsonRpcConfiguration())
             .webSocketConfiguration(node.webSocketConfiguration())
             .dataDir(node.homeDirectory())
-            .bannedNodeIds(Collections.emptySet())
```
mbaxter: We no longer need to configure an empty set of banned nodes

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/PeerDiscoveryAgent.java
```diff
@@ -272,29 +263,6 @@ public boolean removePeerBondedObserver(final long observerId) {
...
-   * Adds an observer that will get called when a peer is dropped from the peer table.
-   *
-   * <p><i>No guarantees are made about the order in which observers are invoked.</i>
-   *
-   * @param observer The observer to call.
-   * @return A unique ID identifying this observer, to that it can be removed later.
-   */
-  public long observePeerDroppedEvents(final Consumer<PeerDroppedEvent> observer) {
-    checkNotNull(observer);
-    return peerDroppedObservers.subscribe(observer);
```
mbaxter: This was dead code - listeners were never actually notified of dropped peers

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/PeerDiscoveryAgent.java
```diff
@@ -80,7 +72,6 @@
   /* The keypair used to sign messages. */
   protected final SECP256K1.KeyPair keyPair;
   private final BytesValue id;
-  private final PeerTable peerTable;
```
mbaxter: `PeerTable` isn't used in `PeerDiscoveryAgent` - moved it into the `PeerDiscoveryController` where it is used.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
```diff
@@ -367,27 +382,11 @@ private boolean addToPeerTable(final DiscoveryPeer peer) {
     return true;
   }
 
-  @VisibleForTesting
-  boolean dropFromPeerTable(final DiscoveryPeer peer) {
```
mbaxter: Dead code

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/network/DefaultP2PNetwork.java
```diff
@@ -348,6 +349,7 @@ protected void initChannel(final SocketChannel ch) {
 
               if (!isPeerAllowed(connection)) {
                 connection.disconnect(DisconnectReason.UNKNOWN);
+                peerDiscoveryAgent.dropPeer(connection.getPeer());
```
mbaxter: Previously we were dropping all peers who disconnected.  Now, we're explicitly dropping a subset of disconnected peers.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/network/DefaultP2PNetwork.java
```diff
@@ -407,8 +411,9 @@ void attemptPeerConnections() {
         streamDiscoveredPeers()
             .filter(peer -> peer.getStatus() == PeerDiscoveryStatus.BONDED)
             .filter(peer -> !isConnected(peer) && !isConnecting(peer))
+            .sorted(Comparator.comparing(DiscoveryPeer::getLastAttemptedConnection))
```
mbaxter: Try connecting to discovery peers who haven't recently been contacted.

ethereum/p2p/src/test/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/PeerDiscoveryControllerTest.java
```diff
@@ -1065,27 +1052,6 @@ public void shouldNotRespondToPingFromNonWhitelistedDiscoveryPeer() {
...
-    final List<DiscoveryPeer> peers = createPeersInLastBucket(localPeer, 1);
-    final DiscoveryPeer peer = peers.get(0);
-    final PeerTable peerTableSpy = spy(peerTable);
-    final Consumer<PeerDroppedEvent> peerDroppedEventConsumer = mock(Consumer.class);
-    final Subscribers<Consumer<PeerDroppedEvent>> peerDroppedSubscribers = new Subscribers();
-    peerDroppedSubscribers.subscribe(peerDroppedEventConsumer);
-
-    doReturn(EvictResult.absent()).when(peerTableSpy).tryEvict(any());
-
-    controller = getControllerBuilder().peerDroppedObservers(peerDroppedSubscribers).build();
```
mbaxter: This test hit the dead code that was removed

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/PeerDiscoveryAgent.java
```diff
@@ -61,17 +54,16 @@
...
  * via UDP.
  */
-public abstract class PeerDiscoveryAgent implements DisconnectCallback {
+public abstract class PeerDiscoveryAgent {
   protected static final Logger LOG = LogManager.getLogger();
 
   // The devp2p specification says only accept packets up to 1280, but some
   // clients ignore that, so we add in a little extra padding.
   private static final int MAX_PACKET_SIZE_BYTES = 1600;
-  private static final long PEER_REFRESH_INTERVAL_MS = MILLISECONDS.convert(30, TimeUnit.MINUTES);
```
mbaxter: Moved this constant into `PeerDiscoveryController` where it is used

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
```diff
@@ -277,20 +304,19 @@ public void onMessage(final Packet packet, final DiscoveryPeer sender) {
...
+    if (!isPeerPermittedToSendMessages(sender)) {
+      LOG.trace("Dropping packet from disallowed peer ({})", sender.getEnodeURLString());
       return;
     }
 
     // Load the peer from the table, or use the instance that comes in.
     final Optional<DiscoveryPeer> maybeKnownPeer = peerTable.get(sender);
     final DiscoveryPeer peer = maybeKnownPeer.orElse(sender);
     final boolean peerKnown = maybeKnownPeer.isPresent();
-    final boolean peerBlacklisted = peerBlacklist.contains(peer);
```
shemnon: Without the blacklist peers that are "bad" will still hit our peer table.  Is this acceptable per what we need to do?  Will we advertise that we've seen "blackbeard" even if we refuse to talk to him?  Should we be tracking peer-no-grata?  Or are we checking permissions at every time we would pull said peer out of the peer table?

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/RecursivePeerRefreshState.java
```diff
@@ -390,4 +379,9 @@ public void cancelTimeout() {
       timeoutCancelled.set(true);
     }
   }
+
+  @FunctionalInterface
+  public interface OutboundDiscoveryMessagingPermissions {
+    boolean isPermitted(Peer remotePeer);
```
shemnon: Maybe `isRemotePermitted`?  In case we add this interface to an uber-permissions object with inbound permissions implicit on the local peer, so the names won't collide?

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/network/DefaultP2PNetwork.java
```diff
@@ -778,7 +761,6 @@ private void validate() {
       checkState(
           supportedCapabilities != null && supportedCapabilities.size() > 0,
           "Supported capabilities must be set and non-empty.");
-      checkState(peerBlacklist != null, "PeerBlacklist must be set.");
```
shemnon: do we handle null peerPermissions correctly?

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/permissions/PeerPermissions.java
```diff
@@ -0,0 +1,110 @@
...
+
+  private static class NoopPeerPermissions extends PeerPermissions {
+    @Override
+    public boolean isPermitted(final Peer peer) {
+      return true;
+    }
+  }
+
+  private static class CombinedPeerPermissions extends PeerPermissions {
+    private final List<PeerPermissions> permissions;
```
shemnon: Consider ImmutableList.  Since we are allowing a list to shortcut other CombinedPeerPermissions by collapsing it into this list we should make it explicit that the object we are shortcutting won't mutate.  There is an ImmutableList collector as part of Guava.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/permissions/PeerPermissionsBlacklist.java
```diff
@@ -0,0 +1,81 @@
...
+
+public class PeerPermissionsBlacklist extends PeerPermissions {
+  private static int DEFAULT_INITIAL_CAPACITY = 20;
+
+  private final Set<BytesValue> blacklist;
+
+  private PeerPermissionsBlacklist(final int initialCapacity, final OptionalInt maxSize) {
+    if (maxSize.isPresent()) {
+      blacklist =
+          LimitedSet.create(initialCapacity, maxSize.getAsInt(), Mode.DROP_LEAST_RECENTLY_ACCESSED);
```
shemnon: Consider Guava's Cache instead: https://github.com/google/guava/wiki/CachesExplained - it doesn't do Least Recently Accessed 100% but it does have better concurrency then being wrapped in a synchronized set.

Under the covers Java's sets are Maps with the key being the entry and I can't remember what the value is.

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -214,7 +215,34 @@ void setBootnodes(final List<String> values) {
...
+                    return BytesValue.fromHexString(v, EnodeURL.NODE_ID_SIZE);
+                  })
+              .collect(Collectors.toList());
+    } catch (final IllegalArgumentException e) {
+      throw new ParameterException(
+          commandLine, "Invalid ids supplied to '--banned-node-ids'. " + e.getMessage());
+    }
+  }
+
+  private static void validateNodeIdString(final String nodeId) {
```
shemnon: could this be part of ENodeURL?

pantheon/src/test/java/tech/pegasys/pantheon/cli/PantheonCommandTest.java
```diff
@@ -1007,18 +1001,48 @@ public void bootnodesOptionMustBeUsed() {
...
 
-    assertThat(stringListArgumentCaptor.getValue().toArray()).isEqualTo(nodes);
+    assertThat(bytesValueCollectionCollector.getValue().toArray()).isEqualTo(nodes);
 
     assertThat(commandOutput.toString()).isEmpty();
     assertThat(commandErrorOutput.toString()).isEmpty();
   }
 
+  @Test
+  public void callingWithBannedNodeidsOptionButNoValueMustDisplayErrorAndUsage() {
```
shemnon: Can these be moved to 986?  It makes the diff more comprehensible.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
```diff
@@ -252,10 +256,33 @@ public void start() {
...
+
+  private void handlePermissionsUpdate(
+      final boolean addRestrictions, final Optional<List<Peer>> affectedPeers) {
+    if (!addRestrictions) {
+      // Nothing to do if permissions were relaxed
+      return;
+    }
+
+    // If we have an explicit list of peers, drop each peer from our discovery table
+    affectedPeers.ifPresent(peers -> peers.forEach(this::dropPeer));
```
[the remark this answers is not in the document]
mbaxter: The current interface just has a blanket rule where peers are fully allowed or fully disallowed.  So if the permissions were updated to add restrictions and a list of peers were provided, it means those peers are no longer allowed.  The follow-up PR will make these checks more granular such that we'll re-evaluate at this point.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
mbaxter: The blacklist basically transformed into `peerPermissions` which gets checked in the method  `isPeerPermittedToSendMessages` above on line 307.  So we're still filtering out the bad peers here.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/network/DefaultP2PNetwork.java
mbaxter: I think so - `peerPermissions` gets initialized with a default value, and the setter does a null check so that it can't get overwritten with a null value.

pantheon/src/test/java/tech/pegasys/pantheon/cli/PantheonCommandTest.java
mbaxter: The bannedNodeIds test was sandwiched in between some bootnode tests, so moved things around a bit for a more coherent test layout 🤷‍♀

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/RecursivePeerRefreshState.java
mbaxter: I've got a follow-up PR that will rework these interfaces.  Right now I'm moving towards an API like:  isPermitted(localNode, remotePeer, action).

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/permissions/PeerPermissionsBlacklist.java
mbaxter: I may follow-up on this suggestion in another PR.  But skipping for now.

PR #1447, opened 2019-05-15
Reduce synchronization in PendingTransactions

ajsutton:
## PR description
Reduce the amount of work `PendingTransactions` does inside synchronized blocks to reduce contention.  

* The hash to transaction map is now a `ConcurrentHashMap` allowing queries like `getTransactionByHash` to be performed without any need for synchronization.
* Iterating to remove old transactions or find local transactions to send to peers can now be done without synchronized blocks except when an old transaction is actually being removed.
* Adds and removes still require synchronization but notifying listeners is now done outside the synchronized block.


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/transactions/PendingTransactions.java
```diff
@@ -103,28 +104,19 @@ public PendingTransactions(
...
-                  transaction -> transaction.getAddedToPoolAt().isBefore(removeTransactionsBefore))
-              .collect(toList());
-      transactionsToRemove.forEach(transaction -> removeTransaction(transaction.getTransaction()));
-    }
+    final Instant removeTransactionsBefore =
+        clock.instant().minus(maxTransactionRetentionHours, ChronoUnit.HOURS);
+
+    pendingTransactions.values().stream()
+        .filter(transaction -> transaction.getAddedToPoolAt().isBefore(removeTransactionsBefore))
+        .forEach(transaction -> removeTransaction(transaction.getTransaction()));
```
shemnon: It would be nice if there was some way to assert that the underlying spliterator had the CONCURRENT characteristic before performing.  While this is correct (because it is from a ConcurrentHashMap) I am concerned that future alterations may miss this nuance.

ajsutton: If the map ever gets changed back to a non-concurrent map, all the places it's accessed from will be problematic.  This is actually the place that kind of mistake will be most obvious because the unit tests will fail with `ConcurrentModificationException`.  The others will just be subtle bugs in the face of multiple threads.

PR #1449, opened 2019-05-15
[PAN-2647] Validate Private Transaction nonce before submitting to Transaction Pool

iikirilov:
## PR description
Validates the Private Transaction nonce before submitting to the Transaction Pool.

Returns `PRIVATE_NONCE_TOO_LOW` or `INCORRECT_PRIVATE_NONCE` if the nonce is not valid

## Fixed Issue(s)
[PAN-2647]

[PAN-2647]: https://pegasys1.atlassian.net/browse/PAN-2647

PR #1455, opened 2019-05-16
[PAN-2672] Return specific and useful error for enclave issues

Puneetha17:
## PR description

This PR gets the error returned from the `Orion` using regex and propagates is back to the user. This solution is specific to `Orion` and not any other enclave.


ethereum/core/src/integration-test/java/tech/pegasys/pantheon/ethereum/mainnet/precompiles/privacy/PrivacyPrecompiledContractIntegrationTest.java
```diff
@@ -171,4 +170,36 @@ public void testSendAndReceive() throws IOException {
...
+  public void testNoPrivateKeyError() throws RuntimeException {
+    List<String> publicKeys = testHarness.getPublicKeys();
+    publicKeys.add("noPrivateKey");
+
+    String s = new String(VALID_PRIVATE_TRANSACTION_RLP_BASE64, UTF_8);
+    SendRequest sc = new SendRequest(s, publicKeys.get(0), publicKeys);
+    try {
+      enclave.send(sc);
+    } catch (IOException e) {
+      if (!e.getMessage().contains("EnclaveDecodePublicKey")) {
```
iikirilov: Test errors with `@Test` annotation - https://junit.org/junit4/faq.html#atests_7

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/privacy/EeaSendRawTransaction.java
```diff
@@ -116,8 +115,8 @@ private Transaction handlePrivateTransaction(final PrivateTransaction privateTra
     try {
       return privateTransactionHandler.handle(
           privateTransaction, () -> getNonce(privateTransaction.getSender()));
-    } catch (final IOException e) {
-      throw new InvalidJsonRpcRequestException("Unable to handle private transaction", e);
+    } catch (final Exception e) {
+      throw new InvalidJsonRpcRequestException(e.getMessage());
```
iikirilov: this was changed in https://github.com/PegaSysEng/pantheon/pull/1449

enclave/src/main/java/tech/pegasys/pantheon/enclave/Enclave.java
```diff
@@ -76,8 +78,19 @@ private Request buildPostRequest(
     try (Response response = client.newCall(request).execute()) {
       return objectMapper.readValue(response.body().string(), responseType);
     } catch (IOException e) {
-      LOG.error("Enclave failed to execute ", request);
-      throw new IOException("Enclave failed to execute post", e);
+      LOG.error("Enclave failed to execute {}", request);
+      String regex = "\\{(.*?)(\"([^\"]*)\")\\}";
```
iikirilov: Is there any way to avoid regex? Ideally, the response should be serialized to an object.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/privacy/EeaSendRawTransaction.java
Puneetha17: Will update once the PR is merged

ethereum/core/src/integration-test/java/tech/pegasys/pantheon/ethereum/mainnet/precompiles/privacy/PrivacyPrecompiledContractIntegrationTest.java
Puneetha17: I want to test the specific error message and not just if the exception is throw.

iikirilov: I don't agree. This should return an ErrorResponse object. Any exception should be related to connection issues. If the `send()` returns a response (even if it is an error) I think it should be kept in an `ErrorResponse` object

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/precompiles/privacy/PrivacyPrecompiledContract.java
```diff
@@ -102,7 +101,7 @@ public BytesValue compute(final BytesValue input, final MessageFrame messageFram
     ReceiveResponse receiveResponse;
     try {
       receiveResponse = enclave.receive(receiveRequest);
-    } catch (IOException e) {
+    } catch (Exception e) {
       LOG.debug("Enclave probably does not have private transaction with key {}.", key, e);
```
ekellstrand: Existing code, but should this log at a higher level error? or is this a normal, common use case.  I would expect that if it's normal/expected, then logic would pick those out and log at debug. but for everything else, error level and maybe rethrow?

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/privacy/PrivateTransactionHandler.java
```diff
@@ -59,15 +58,15 @@ public PrivateTransactionHandler(
...
     LOG.trace("Handling private transaction {}", privateTransaction.toString());
     final SendRequest sendRequest = createSendRequest(privateTransaction);
     final SendResponse sendResponse;
     try {
       LOG.trace("Storing private transaction in enclave");
       sendResponse = enclave.send(sendRequest);
-    } catch (IOException e) {
-      LOG.error("Failed to store private transaction in enclave", e);
+    } catch (Exception e) {
+      LOG.error("Failed to store private transaction in enclave {}", e.getMessage());
```
ekellstrand: LOG.error should include traceback.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/privacy/PrivateTransactionHandler.java
```diff
@@ -59,15 +58,15 @@ public PrivateTransactionHandler(
...
       throw e;
```
ekellstrand: Should this throw a new exception that wraps e and includes the log message?

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/privacy/EeaSendRawTransaction.java
ekellstrand: Rethrow should include the whole traceback, or add a comment if there is a reason we don't want to do that.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/response/JsonRpcError.java
```diff
@@ -101,7 +101,24 @@
   CANT_CONNECT_TO_LOCAL_PEER(-32100, "Cannot add local node as peer."),
 
   // Invalid input errors
-  ENODE_ID_INVALID(-32000, EnodeURL.INVALID_NODE_ID_LENGTH);
+  ENODE_ID_INVALID(-32000, EnodeURL.INVALID_NODE_ID_LENGTH),
+
+  // Enclave errors
+  NODE_MISSING_PEER_URL(-50200, "NodeMissingPeerUrl"),
```
ekellstrand: Question:  Is there a reason we use the same error code for every error?  Would it make it easier to debug if everything used unique codes?

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/precompiles/privacy/PrivacyPrecompiledContract.java
iikirilov: no rethrow here - this is how the precompile should behave.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/privacy/PrivateTransactionHandler.java
iikirilov: Potentially this can be done, but that is for another PR. The exception is caught in `EeaSendRawTransaction` and converted to a JSON-RPC response using `e.getMessage()`

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/response/JsonRpcError.java
iikirilov: There is no documentation regarding what the error code for Enclave related errors should be in the JSON-RPC spec for ethereum clients.

PR #1463, opened 2019-05-17
Events Plugin - Add initial "NewBlock" event message

shemnon:
## PR description

* It returns for any block we would broadcast to other peers, when
  we would broadcast them.
* It returns a JSON String containing hask, number, and timestamp
* This event data is not set in stone, it may change in type or content.
* Acceptance tests and unit tests got a re-work away from the assumption
  that there is only one plugin type.


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockBroadcaster.java
```diff
@@ -16,21 +16,34 @@
...
+  public long subscribePropagateNewBlocks(final Consumer<Block> callback) {
+    return newBlockSubscribers.subscribe(callback);
+  }
+
+  public void unsubscribePropagateNewBlocks(final long id) {
+    newBlockSubscribers.unsubscribe(id);
+  }
+
   public void propagate(final Block block, final UInt256 totalDifficulty) {
+    newBlockSubscribers.forEach(listener -> listener.accept(block));
```
mbaxter: I don't think you want to add this functionality here.  `Blockchain.observeBlockAdded` already supports observation of new block events.

shemnon: This event needs to be fired between when we validate the header and when we validate the entire block.

The stack for this code is in part
```
Breakpoint reached
	  at tech.pegasys.pantheon.ethereum.eth.sync.BlockBroadcaster.propagate(BlockBroadcaster.java:46)
	  at tech.pegasys.pantheon.ethereum.eth.sync.BlockPropagationManager.broadcastBlock(BlockPropagationManager.java:248)
	  at tech.pegasys.pantheon.ethereum.eth.sync.BlockPropagationManager.lambda$validateAndProcessPendingBlock$6(BlockPropagationManager.java:306)
	  at ...
```

And the stack for using the `Blockchain.observeBlockAdded` event is
```
...
	  at tech.pegasys.pantheon.ethereum.eth.sync.BlockPropagationManager.runImportTask(BlockPropagationManager.java:323)
	  at tech.pegasys.pantheon.ethereum.eth.sync.BlockPropagationManager.validateAndProcessPendingBlock(BlockPropagationManager.java:307)
	  at tech.pegasys.pantheon.ethereum.eth.sync.BlockPropagationManager.lambda$importOrSavePendingBlock$5(BlockPropagationManager.java:297)
	  at tech.pegasys.pantheon.ethereum.eth.sync.BlockPropagationManager$$Lambda$1088.740466537.get(Unknown Source:-1)
...
```

Like 307 for BlockPropagationManager runs the full block validation, which is later than the requirements we have for when this information should be pushed to the events plugin.  We don't want full block validation for this event just header (e.g. proof of work) validation.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockBroadcaster.java
```diff
@@ -16,21 +16,34 @@
...
+import java.util.function.Consumer;
+
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
 public class BlockBroadcaster {
   private static final Logger LOG = LogManager.getLogger();
 
   private final EthContext ethContext;
+  private final Subscribers<Consumer<Block>> newBlockSubscribers = new Subscribers<>();
```
mbaxter:
```suggestion
  private final Subscribers<Consumer<Block>> blockPropagatedSubscribers = new Subscribers<>();
```

pantheon/src/main/java/tech/pegasys/pantheon/services/PantheonEventsImpl.java
```diff
@@ -0,0 +1,61 @@
...
+  private static Logger LOG = LogManager.getLogger();
+
+  private final BlockBroadcaster blockBroadcaster;
+
+  public PantheonEventsImpl(final BlockBroadcaster blockBroadcaster) {
+    this.blockBroadcaster = blockBroadcaster;
+  }
+
+  @Override
+  public Object addBlockAddedListener(final Consumer<String> blockJSONListener) {
```
mbaxter:
```suggestion
  public Object observeNewBlockPropagated(final Consumer<String> blockJSONListener) {
```

pantheon/src/main/java/tech/pegasys/pantheon/services/PantheonEventsImpl.java
```diff
@@ -0,0 +1,61 @@
...
+    LOG.info("Adding Listener");
+    return blockBroadcaster.subscribePropagateNewBlocks(
+        block -> dispatchNewBlockMessage(block, blockJSONListener));
+  }
+
+  @Override
+  public void removeBlockAddedObserver(final Object listenerIdentifier) {
```
mbaxter:
```suggestion
  public void removeNewBlockPropagatedObserver(final Object listenerIdentifier) {
```

pantheon/src/main/java/tech/pegasys/pantheon/services/PantheonEventsImpl.java
```diff
@@ -0,0 +1,61 @@
...
+    LOG.info("Removing Listener");
+    if (listenerIdentifier instanceof Long) {
+      blockBroadcaster.unsubscribePropagateNewBlocks((Long) listenerIdentifier);
+    }
+  }
+
+  private void dispatchNewBlockMessage(final Block block, final Consumer<String> callback) {
```
mbaxter:
```suggestion
  private void dispatchNewBlockPropagatedMessage(final Block block, final Consumer<String> callback) {
```

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/BlockBroadcaster.java
mbaxter: Ok, that makes sense.  I think we just need to change the naming to make it clear what this event is.  In the future if we handle NEW_BLOCK_HASHES / NEW_BLOCK messages differently, we may want to expose more events to get all of the new blocks on the network.  Most clients don't propagate out NEW_BLOCK messages in response to receiving NEW_BLOCK_HASHES from the network, for example.

pantheon/src/main/java/tech/pegasys/pantheon/services/PantheonEventsImpl.java
mbaxter: (optional) Consider adding a functional interface in place of `Consumer<String>`.

shemnon: Listener is more prevalent in our code base than observer.  So I added the Propagated but kept the Listener.

pantheon/src/main/java/tech/pegasys/pantheon/services/PantheonEventsImpl.java
shemnon: Listener is more prevalent in our code base than observer.  So I added the Propagated but kept the Listener.

pantheon/src/main/java/tech/pegasys/pantheon/services/PantheonEventsImpl.java
shemnon: Listener is more prevalent in our code base than observer.  So I added the Propagated but kept the Listener.

pantheon/src/main/java/tech/pegasys/pantheon/services/PantheonEventsImpl.java
shemnon: Done.

pantheon/src/main/java/tech/pegasys/pantheon/services/PantheonEventsImpl.java
```diff
@@ -0,0 +1,53 @@
...
+
+public class PantheonEventsImpl implements PantheonEvents {
+  private final BlockBroadcaster blockBroadcaster;
+
+  public PantheonEventsImpl(final BlockBroadcaster blockBroadcaster) {
+    this.blockBroadcaster = blockBroadcaster;
+  }
+
+  @Override
+  public Object addNewBlockPropagatedLister(final NewBlockPropagatedListener listener) {
```
mbaxter:
```suggestion
  public Object addNewBlockPropagatedListener(final NewBlockPropagatedListener listener) {
```
Typo

build.gradle
```diff
@@ -105,7 +105,7 @@ allprojects {
     java {
       // This path needs to be relative to each project
       target fileTree('.') {
-        include '**/*.java'
+        include '**/src/*/java/**/*.java'
```
mbaxter: what are these changes for??

shemnon: meant for another PR.  Spotless was catching intellij generated files from .../out/classes/....  By limiting src to having to live in src directories generated and transient java files don't get style checked.

PR #1464, opened 2019-05-18
PAN-2711 - Don't require 'to' in 'blocks' queries

shemnon:
## PR description

`{blocks(from:7777777) {number}}` should work, instead it is requiring a `to:` argument when it should be implied as latest


ethereum/graphqlrpc/src/test/java/tech/pegasys/pantheon/ethereum/graphqlrpc/EthGraphQLRpcHttpBySpecTest.java
```diff
@@ -108,14 +109,14 @@ private void graphQLRPCCall(final String name) throws IOException {
...
-      assertThat(resp.code()).isEqualTo(expectedStatusCode);
-
       final JsonObject expectedRespBody = spec.getJsonObject("response");
       final String resultStr = resp.body().string();
 
       final JsonObject result = new JsonObject(resultStr);
       assertThat(result).isEqualTo(expectedRespBody);
+
+      final int expectedStatusCode = spec.getInteger("statusCode");
+      assertThat(resp.code()).isEqualTo(expectedStatusCode);
```
NicolasMassart: Why moving this at the end?

shemnon: So when an error occurs the response body shows up in the junit report.

PR #1471, opened 2019-05-20
[PAN-2614] Expand permissioning interface

mbaxter:
## PR description
This PR consolidates all p2p permission-like logic behind a single interface.  To facilitate this change, the `PeerPermissions` interface has been expanded to provide more granular control, allowing the `NodePermissioningController` logic to sit behind this interface via the `NodePermissioningAdapter`.


ethereum/p2p/src/test/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/PeerDiscoveryControllerTest.java
```diff
@@ -973,83 +972,205 @@ public void shouldNotAddPeerInNeighborsPacketWithoutBonding() {
   }
 
   @Test
-  public void shouldNotBondWithNonPermittedPeer() {
```
mbaxter: Just adding a note on these changes since the diff is confusing.  

The tests `shouldNotBondWithNonPermittedPeer` and `shouldNotRespondToPingFromNonWhitelistedDiscoveryPeer` were removed.  More comprehensive tests that include these cases were added to `PeerDiscoveryAgentTest`.   The new tests here are not related to the previous tests that were removed.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
```diff
@@ -752,6 +748,12 @@ public Builder tableRefreshIntervalMs(final long tableRefreshIntervalMs) {
       return this;
     }
 
+    public Builder cleanPeerTableIntervalMs(final long cleanPeerTableIntervalMs) {
```
shemnon: This method appears to not have test coverage, all the other builder methods do.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/network/DefaultP2PNetwork.java
```diff
@@ -587,43 +583,44 @@ public void start() {
...
+          .map(Optional::get)
+          .forEach(conn -> conn.disconnect(DisconnectReason.REQUESTED));
+    } else {
+      checkAllConnections();
+    }
+  }
+
+  private void checkAllConnections() {
+    if (!rlpxPermissions.isPresent()) {
+      return;
```
shemnon: This branch has no coverage.  Is this a case we expect to see in normal operations?  Would we ever realistically not have rlpxPermissions?

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/network/NodePermissioningAdapter.java
```diff
@@ -0,0 +1,130 @@
...
+    this.nodePermissioningController = nodePermissioningController;
+    this.bootnodes = bootnodes;
+    this.blockchain = blockchain;
+
+    // TODO: These events should be more targeted
+    blockchainListenId =
+        blockchain.observeBlockAdded((evt, chain) -> dispatchUpdate(true, Optional.empty()));
+    nodePermissioningListenId =
+        this.nodePermissioningController.subscribeToUpdates(
+            () -> dispatchUpdate(true, Optional.empty()));
```
shemnon: This event never gets fired in tests.  Is this something that the Dapp will likely exercise?

lucassaldanha: The permissioninng dapp doesn't play a role in this flow. Our `NodePermissioningController` needs to subscribe to chain updates so we can verify if a new block must trigger a re-evaluation of the rules (e.g. the block contains a transaction that removes a node from the whitelist).

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/network/NodePermissioningAdapter.java
```diff
@@ -0,0 +1,130 @@
...
+
+  private boolean allowOutboundBonding(final Peer localNode, final Peer remotePeer) {
+    boolean outboundMessagingAllowed = outboundIsPermitted(localNode, remotePeer);
+    if (!nodePermissioningController.getSyncStatusNodePermissioningProvider().isPresent()) {
+      return outboundMessagingAllowed;
+    }
+
+    // We're using smart-contract based permissioning
+    // If we're out of sync, only allow bonding to our bootnodes
+    final SyncStatusNodePermissioningProvider syncStatus =
```
lucassaldanha: All node permissioning providers are supposed to be under the `NodePermissioningController` object. The controller will take care of permissioning rules and the precedence order of them. We shouldn't be using it outside the `NodePermissioningController`.

Any reason for using it here?

mbaxter: I was trying to account for this code that's currently in the `PeerDiscoveryController`: https://github.com/PegaSysEng/pantheon/blob/8e27ca73d35d9d46b9eb9c549ccb78cfdd8d6fbe/ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/PeerDiscoveryController.java#L222-L238

mbaxter: Which is removed here: https://github.com/PegaSysEng/pantheon/pull/1471/files#diff-11104ef3adb5bd629a56cc0044b0cc1aL222

mbaxter: Also - your comment reminded me to remove the now-dead code here: https://github.com/PegaSysEng/pantheon/pull/1471/commits/af58dba7a6ddf0da6cc95c71c407e7269f14c8ef

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/PeerDiscoveryController.java
mbaxter: I just added the method for completeness of the builder, it isn't actually used.  Do you think we need a test for this?

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/network/DefaultP2PNetwork.java
mbaxter: The fact that this permissions object is wrapped in an `Optional` is unfortunate, and hopefully that will be going away in an upcoming PR.  I had tried initializing it in the constructor of `DefaultP2PNetwork` as a final unwrapped type that gets access to `DefaultP2PNetwork.getLocalEnode` via a `Supplier` argument in the `PeerRlpxPermissions` constructor.  But that caused issues with tests that needed to mock the `EnodeURL` returned by `DefaultP2pNetwork.getLocalEnode`.  For now, `rlpxPermissons` is created after the node starts and has a definite local enode.  

In this particular spot, this code will run when permissions are updated, which could happen before the node starts up.  Although, looks like I can simplify this a little by passing in the unwrapped permissions object - I'll push up that change.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/network/NodePermissioningAdapter.java
mbaxter: We have acceptance tests that will hit this branch.  The `NodePermissioningController` is in a separate package, so I don't want to pull in a real instance and try figuring out how to get that to broadcast an update within the p2p test suite.  Testing for that object should be covered in the `permissioning` package.

PR #1477, opened 2019-05-21
Have ThreadPantheonNodeRunner support plugin tests.

shemnon:
## PR description

* Do plugin lifecycle in ThreadPantheonNodeRunner
* Sniff out rendezvous directory for test during register call, it changes.


plugins/src/test/java/tech/pegasys/pantheon/plugins/TestPantheonEventsPlugin.java
```diff
@@ -62,8 +65,7 @@ private void onBlockAnnounce(final String json) {
     final int blockCount = blockCounter.incrementAndGet();
     LOG.info("I got a new block! (I've seen {}) - {}", blockCount, json);
     try {
-      final File callbackFile =
-          new File(System.getProperty("pantheon.plugins.dir", "plugins"), "newBlock." + blockCount);
+      final File callbackFile = new File(callbackDir, "newBlock." + blockCount);
```
mbaxter: Should each plugin get its own directory??

shemnon: I'll do that when we add another plugin.

PR #1485, opened 2019-05-22
PAN-2715 - return block not found reasons in error

shemnon:
## PR description

When looking for a specific block and failing we should explain why
in the error message.
Also, clean up test names and grouping

## Fixed Issue(s)

PAN-2715


ethereum/graphql/src/test/resources/tech/pegasys/pantheon/ethereum/graphql/eth_getBlock_wrongParams.json
```diff
@@ -0,0 +1,25 @@
+{
+  "request": "{block (number: \"0x03\", hash : \"0xc8df1f061abb4d0c107b2b1a794ade8780b3120e681f723fe55a7be586d95ba6\") {number transactions{hash} timestamp difficulty totalDifficulty gasUsed gasLimit hash nonce ommerCount logsBloom mixHash ommerHash extraData stateRoot receiptsRoot transactionCount transactionsRoot}} ",
```
NicolasMassart: I don't know how you can add a comment in there but I wonder what the invalid param is exactly in this specific request.
Also would there be a way to indicate it in the error message ?

PR #1487, opened 2019-05-23
Updated to add WS token content

MadelineMurray:
Updated to add WS token content


Errorific:
Just noticed all the examples beyond the changes say 'https'. It's just http, no ssl.

docs/Pantheon-API/Authentication.md
```diff
@@ -65,11 +65,25 @@ options to specify the [credentials file](#credentials-file).
...
-To obtain an authentication token, make a request to the `/login` endpoint with your username and password. 
+To obtain an authentication token, make a request to the `/login` endpoint with your username and password. Specify the 
+HTTPS port or the WS port to obtain a token to authenticate over HTTPS or WS respectively. A different token is required 
+for HTTPS and WS. 
 
 !!! example
-    ```bash tab="curl HTTPS request"
-    curl -X POST --data '{"username":"username1","password":"pegasys"}' <JSON-RPC-https-endpoint:port>/login
+    ```bash tab="Obtain Token for HTTPS"
+    curl -X POST --data '{"username":"username1","password":"pegasys"}' <JSON-RPC-https-endpoint:https-port>/login
```
Errorific: "JSON-RPC-https-endpoint" should be "JSON-RPC-hostname" everywhere, the endpoint is the hostname + the port together rather than separated.

NicolasMassart: I agree an endpoint should be host+port so endpoint:port doesn't seem right.

MadelineMurray:
> Just noticed all the examples beyond the changes say 'https'. It's just http, no ssl.

@Errorific - I used https in the examples because of this note further up: 

Authenticated requests must be made over HTTPS. HTTPS is encrypted which prevents eavesdropping on the connection to obtain the JWT token from the requests. 

But they should be http?

docs/Pantheon-API/Authentication.md
MadelineMurray: Updated

docs/Pantheon-API/Authentication.md
```diff
@@ -7,7 +7,7 @@ Pantheon uses the username and password to authenticate users and [JWT tokens](h
 
 !!! important 
     Authenticated requests must be made over HTTPS. HTTPS is encrypted which prevents eavesdropping on the connection
-    to obtain the JWT token from the requests.
+    to obtain the JWT token from the requests. Run Pantheon behind a reverse proxy or load balancer to perform the SSL.
```
Errorific: Still doesn't feel right. Think this is a better version of the whole paragraph.
```Authenticated requests should be made over HTTPS to prevent interception of authentication credentials and authenticated tokens. As Pantheon cannot provide a HTTPS connection natively it is recommended that production deployments be run behind a network layer that can provide SSL termination.```

NicolasMassart: I'm not comfortable with "As Pantheon cannot provide a HTTPS connection natively". When I read it I feels like pantheon lacks some security features. Actually it's just not its role to handle this transport security, so it's not a lack of security. I would then propose something like "As Pantheon role is not to provide transport security at the application level, it is recommended that you handle secure HTTPS connection for your production environment using a network layer that can provide SSL capabilities like a reverse proxy" Of course having a better sentence and a correct grammar is required, but you have the idea. The goal is to be as precise as Chris but without letting  users think pantheon is bad because it doesn't do SSL. It's not its job.

MadelineMurray: I changed it to 'does not' rather than 'can not' -> it's a deliberate choice not to do it in Pantheon.

PR #1489, opened 2019-05-23
[PAN-2560] Add LocalNode class

mbaxter:
## PR description
Add `LocalNode` class to hold information about the local node that is not immediately available, but becomes available as the node starts up.  This class is convenient because it allows us to pass a `final` dependency to classes that need local node data but are constructed before the node is fully up and running.


ethereum/p2p/src/test/java/tech/pegasys/pantheon/ethereum/p2p/network/netty/DeFramerTest.java
```diff
@@ -81,13 +82,19 @@
...
   private final InetSocketAddress remoteAddress = new InetSocketAddress("127.0.0.1", remotePort);
 
-  private final PeerInfo peerInfo =
-      new PeerInfo(
-          5,
-          "abc",
-          Arrays.asList(Capability.create("eth", 63)),
-          30303,
-          BytesValue.fromHexString("0x01"));
+  private final int p2pVersion = 5;
```
shemnon: 👍 to making the names of the parameters described in the code

PR #1492, opened 2019-05-23
[PIE-1580] Metrics for smart contract permissioning actions

Errorific:
plumbing some metrics into the permissioning controllers so we can measure their performance.


ethereum/permissioning/src/main/java/tech/pegasys/pantheon/ethereum/permissioning/NodeSmartContractPermissioningController.java
```diff
@@ -60,11 +66,30 @@ private static BytesValue hashSignature(final String signature) {
...
    */
   public NodeSmartContractPermissioningController(
-      final Address contractAddress, final TransactionSimulator transactionSimulator) {
+      final Address contractAddress,
+      final TransactionSimulator transactionSimulator,
+      final MetricsSystem metricsSystem) {
     this.contractAddress = contractAddress;
     this.transactionSimulator = transactionSimulator;
+
+    this.checkCounter =
```
lucassaldanha: If we have the counter for permitted and unpermitted, do we really need the total counter or can we just infer from the other two metrics?
I'm not aware of the impact of the metrics in system resources so maybe we are better playing safe?

ethereum/permissioning/src/main/java/tech/pegasys/pantheon/ethereum/permissioning/TransactionSmartContractPermissioningController.java
```diff
@@ -61,11 +67,30 @@ private static BytesValue hashSignature(final String signature) {
...
    */
   public TransactionSmartContractPermissioningController(
-      final Address contractAddress, final TransactionSimulator transactionSimulator) {
+      final Address contractAddress,
+      final TransactionSimulator transactionSimulator,
+      final MetricsSystem metricsSystem) {
     this.contractAddress = contractAddress;
     this.transactionSimulator = transactionSimulator;
+
+    this.checkCounter =
```
lucassaldanha: Same thing about the total checks counter.

lucassaldanha:
I'm not sure how the names of the metrics are put together later, but I wonder if we could create some hierarchical namespace to make it easier to understand.

e.g. we could have a "parent" counter for node permissioning and one for account permissioning. In the case of node permissioning, we could then have the metrics for each provider, and for account permissioning the metrics for the local config and onchain checks.

Something like:

`permissioning.node.permitted` = total of permitted communication in the NodePermissioningController
`permissioning.node.unpermitted` = total of permitted communication in the NodePermissioningController
`permissioning.node.local_config.permitted` = total of permitted communication in the NodeLocalConfigPermissioningProvider
`permissioning.node.smart_contract.permitted` = total of permitted communication in the NodeSmartContractPermissioningProvider
...

ethereum/permissioning/src/main/java/tech/pegasys/pantheon/ethereum/permissioning/NodeSmartContractPermissioningController.java
Errorific: There's an outcome where the permissioning check crashes (no contract, bad transaction etc) which increments the total counter but not the other 2. I'm assuming the metrics are no heavier than adding log lines so I decided to be distinct about all the things we might be checking.

PR #1500, opened 2019-05-28
[PAN-1625] Clique AT mining continues if validator offline

jframe:
## PR description
Acceptance test to verify that given a 3 node clique network, if a node is taken offline then mining can continue as there are still sufficient validators.


acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/condition/clique/ExpectBlockHasProposer.java
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  private Address proposerAddress(final Node node) {
+    final Block block = node.execute(eth.block());
+    final BlockHeader blockHeader = createBlockHeader(block);
+    final CliqueExtraData cliqueExtraData = CliqueExtraData.decode(blockHeader);
+    return cliqueExtraData.getProposerAddress();
+  }
+
+  private BlockHeader createBlockHeader(final Block block) {
```
rain-on: suspect this should actually be in a helper function somewhere (not stuck in AT, but testsupport somewhere?) - i.e. the ability to extract a blockHeader object from a block is SO useful...

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/jsonrpc/Clique.java
```diff
@@ -50,6 +53,14 @@ public Clique(final EthTransactions eth, final CliqueTransactions clique) {
     this.clique = clique;
   }
 
+  public List<PantheonNode> validators(final PantheonNode[] nodes) {
```
rain-on: does the name represent what this function does? I.e. what does it do? Sort the params? Why is that a Clique thing?

Also - can probably be static

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/jsonrpc/Clique.java
```diff
@@ -98,6 +109,10 @@ private int cliqueBlockPeriod(final PantheonNode node) {
     return Arrays.stream(validators).map(PantheonNode::getAddress).sorted().toArray(Address[]::new);
   }
 
+  public Condition proposerAtChainHeadEqual(final PantheonNode proposer) {
```
rain-on: maybe not a clique thing?

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/clique/CliqueMiningAcceptanceTest.java
```diff
@@ -72,4 +73,31 @@ public void shouldStallMiningWhenInsufficientValidators() throws IOException {
     minerNode1.verify(net.awaitPeerCount(0));
     minerNode1.verify(clique.noNewBlockCreated(minerNode1));
   }
+
+  @Test
+  public void shouldStillMineWhenANonProposerNodeFailsAndHasSufficientValidators()
+      throws IOException {
+    final PantheonNode[] nodes = {
```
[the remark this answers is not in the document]
jframe: So I know the order of the nodes and tell the last node to stop. It's what we doing elsewhere but still not nice. Should think of a better way of doing this.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/clique/CliqueMiningAcceptanceTest.java
```diff
@@ -72,4 +73,31 @@ public void shouldStallMiningWhenInsufficientValidators() throws IOException {
     minerNode1.verify(net.awaitPeerCount(0));
     minerNode1.verify(clique.noNewBlockCreated(minerNode1));
   }
+
+  @Test
+  public void shouldStillMineWhenANonProposerNodeFailsAndHasSufficientValidators()
```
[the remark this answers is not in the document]
jframe: Don't we absolutely certainty, best I have is copying the approach used in ibft tests and selecting the last validator when sorted by address. Perhaps better would be better would be determine a non-proposer immediately before shutting it down and use the last node in the list. But that might be getting a bit complex.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/clique/CliqueMiningAcceptanceTest.java
```diff
@@ -72,4 +73,31 @@ public void shouldStallMiningWhenInsufficientValidators() throws IOException {
...
+      pantheon.createCliqueNode("miner1"),
+      pantheon.createCliqueNode("miner2"),
+      pantheon.createCliqueNode("miner3")
+    };
+    final List<PantheonNode> validators = clique.validators(nodes);
+    final PantheonNode validator1 = validators.get(0);
+    final PantheonNode validator2 = validators.get(1);
+    final PantheonNode validator3 = validators.get(2);
+
+    cluster.start(validators);
```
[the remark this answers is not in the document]
jframe: It is a bit odd. I think there might be a variant that means I don't have to supply list of nodes to start. Will look into this.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/clique/CliqueMiningAcceptanceTest.java
```diff
@@ -72,4 +73,31 @@ public void shouldStallMiningWhenInsufficientValidators() throws IOException {
...
+    cluster.waitUntil(wait.chainHeadHasProgressedByAtLeast(validator1, 1, 85));
+
+    cluster.stopNode(validator3);
+    validator1.verify(net.awaitPeerCount(1));
+    validator2.verify(net.awaitPeerCount(1));
+
+    cluster.waitOnActiveNodesUntil(wait.chainHeadHasProgressedByAtLeast(validator1, 1));
+    cluster.verifyOnActiveNodes(clique.proposerAtChainHeadEqual(validators.get(0)));
```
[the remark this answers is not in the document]
jframe: done

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/condition/clique/ExpectBlockHasProposer.java
jframe: not quite a useful as might first appear, it's creating a Pantheon blockHeader from the a web3j block so probably only useful for AT. But still think it useful enough so have moved it into a util class.
done.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/clique/CliqueMiningAcceptanceTest.java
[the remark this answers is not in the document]
rain-on: how do you know which order they are assigned addresses? Or is the point that the validators function does that?

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/jsonrpc/Clique.java
jframe: Think it is, the condition needs to decode the clique specific extra data

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/clique/CliqueMiningAcceptanceTest.java
```diff
@@ -72,4 +72,23 @@ public void shouldStallMiningWhenInsufficientValidators() throws IOException {
...
+    final PantheonNode minerNode3 = pantheon.createCliqueNode("miner3");
+    cluster.start(minerNode1, minerNode2, minerNode3);
+
+    cluster.waitUntil(wait.chainHeadHasProgressedByAtLeast(minerNode1, 1, 85));
+
+    cluster.stopNode(minerNode3);
+    cluster.verifyOnActiveNodes(net.awaitPeerCount(1));
+
+    cluster.waitOnActiveNodesUntil(wait.chainHeadHasProgressedByAtLeast(minerNode1, 1));
+    cluster.verifyOnActiveNodes(clique.proposerAtChainHeadEqual(minerNode1));
```
ajsutton: You can't make this guarantee - it would be entirely valid for either minerNode1 or minerNode2 to be the most recent proposer.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/clique/CliqueMiningAcceptanceTest.java
```diff
@@ -72,4 +72,23 @@ public void shouldStallMiningWhenInsufficientValidators() throws IOException {
...
+
+    cluster.waitOnActiveNodesUntil(wait.chainHeadHasProgressedByAtLeast(minerNode1, 1));
+    cluster.verifyOnActiveNodes(clique.proposerAtChainHeadEqual(minerNode2));
```
ajsutton: Similarly, you can't guarantee that minerNode2 will have proposed the next block here - the chain keeps progressing outside the control of the AT so just because the chain progressed by at least 1 doesn't mean it didn't progress by 2 or more and so the proposer may not be who you expect.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/clique/CliqueMiningAcceptanceTest.java
jframe: Yeah your right I can't make that guarantee and it isn't trying to, the method is poorly named. What method is actually doing is verifying a block is created with the given proposer. It doesn't need to the current. I'll think of a better name otherwise it is confusing.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/clique/CliqueMiningAcceptanceTest.java
jframe: Definitely right there, we have no control other the chain. And it could really be either node. Similarly to above the intention isn't to make the hard guarantee. Bad naming really, as the name suggests it expecting the block at chain head to have the specified proposer. But the condition is actually checking a block at some point with the expected proposer. I'll think of better naming of the method at least to make that clear.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/condition/clique/ExpectBlockHasProposer.java
```diff
@@ -0,0 +1,49 @@
...
+import tech.pegasys.pantheon.ethereum.core.BlockHeader;
+import tech.pegasys.pantheon.tests.acceptance.dsl.condition.Condition;
+import tech.pegasys.pantheon.tests.acceptance.dsl.node.Node;
+import tech.pegasys.pantheon.tests.acceptance.dsl.transaction.eth.EthTransactions;
+
+import org.web3j.protocol.core.methods.response.EthBlock.Block;
+
+public class ExpectBlockHasProposer implements Condition {
+  private final EthTransactions eth;
+  private Address proposer;
```
CjHare: nit; `expectedProposer` might be a better fit here as it's more explicit, which is going to be valuable with the method `proposerAddress()`

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/jsonrpc/Clique.java
```diff
@@ -98,6 +99,10 @@ private int cliqueBlockPeriod(final PantheonNode node) {
     return Arrays.stream(validators).map(PantheonNode::getAddress).sorted().toArray(Address[]::new);
   }
 
+  public Condition aBlockIsCreatedByProposer(final PantheonNode proposer) {
```
CjHare: Consistency; drop the `a` prefix == `blockIsCreatedByProposer`

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/condition/clique/ExpectBlockHasProposer.java
jframe: done

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/jsonrpc/Clique.java
jframe: done

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/clique/CliqueMiningAcceptanceTest.java
```diff
@@ -72,4 +72,21 @@ public void shouldStallMiningWhenInsufficientValidators() throws IOException {
     minerNode1.verify(net.awaitPeerCount(0));
     minerNode1.verify(clique.noNewBlockCreated(minerNode1));
   }
+
+  @Test
+  public void shouldStillMineWhenSufficientValidators() throws IOException {
```
CjHare: nit; `shouldStillMineWhenSufficientValidators` -> `shouldStillMineWhithSufficientValidators`

jframe: went something a bit more wordy but hopefully clearer, shouldStillMineWhenANodeFailsAndHasSufficientValidators

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/clique/CliqueMiningAcceptanceTest.java
```diff
@@ -72,4 +72,21 @@ public void shouldStallMiningWhenInsufficientValidators() throws IOException {
     minerNode1.verify(net.awaitPeerCount(0));
     minerNode1.verify(clique.noNewBlockCreated(minerNode1));
   }
+
+  @Test
+  public void shouldStillMineWhenANodeFailsAndHasSufficientValidators() throws IOException {
```
CjHare: `shouldStillMineAfterStoppedValidatorButWithQuroumOfValidators`?

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/condition/clique/ExpectedBlockHasProposer.java
```diff
@@ -0,0 +1,49 @@
...
+import tech.pegasys.pantheon.ethereum.core.BlockHeader;
+import tech.pegasys.pantheon.tests.acceptance.dsl.condition.Condition;
+import tech.pegasys.pantheon.tests.acceptance.dsl.node.Node;
+import tech.pegasys.pantheon.tests.acceptance.dsl.transaction.eth.EthTransactions;
+
+import org.web3j.protocol.core.methods.response.EthBlock.Block;
+
+public class ExpectedBlockHasProposer implements Condition {
+  private final EthTransactions eth;
+  private Address proposer;
```
CjHare: `expectedProposer`?

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/clique/CliqueMiningAcceptanceTest.java
jframe: i like that. done.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/condition/clique/ExpectedBlockHasProposer.java
jframe: done

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/clique/CliqueMiningAcceptanceTest.java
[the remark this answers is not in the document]
jframe: This is being looked as part another PR

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/jsonrpc/Clique.java
jframe: not used anymore

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/clique/CliqueMiningAcceptanceTest.java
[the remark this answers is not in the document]
jframe: validators isn't being used anymore

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/clique/CliqueMiningAcceptanceTest.java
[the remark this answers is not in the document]
jframe: not using this approach anymore

PR #1502, opened 2019-05-28
expand ~ (tilde) to home directory in arguments

RatanRSur:
## PR description
When arguments are crafted in the form --arg=~/some/dir, the shell does not
expand the tilde. Since picocli does not expand the tilde by default,
this leads to inconsistencies.


shemnon:
My concern is that we are bringing in a shell specific conventions into what should be platform neutral code.  Note that this doesn't support the `~username/...` mode.

We can fix the tests to remove the ~ from complete_config.toml and the asserts.

RatanRSur:
What is `~username/...` mode?

shemnon:
It expands to the home directory of the user in question: `ls ~Guest` will get you the Guest home directory on mac, same with any other user name, just tilde is shortcut for "this user."  More common in multi user systems or large NFS systems where user homes have a common path across machines.

RatanRSur:
Ah I didn't know about that. It does seem like a bit of a trap to go down the unix-quirk compatibility rabbit hole. Better left to the user to figure out what their shell is doing and work around it.

shemnon:
I still think we should fix the test.  Re-opening so I can try and push a commit on this

shemnon:
Updates are locked out, so I'll do it in a new PR.

PR #1507, opened 2019-05-29
PAN-2445: Onchain account permissioning

lucassaldanha:
## PR description
* Added TransactionFilter param to ProtocolSpecBuilder
* Updated MainnetProtocolSpec to use tx filter
* Created methods chain to allow adding a tx filter into all protocol specs defined in a protocol schedule
* Added tracing logs for account permissioning checks


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/ProtocolSpec.java
```diff
@@ -243,4 +244,8 @@ public MiningBeneficiaryCalculator getMiningBeneficiaryCalculator() {
   public PrecompileContractRegistry getPrecompileContractRegistry() {
     return precompileContractRegistry;
   }
+
+  public void addTransactionFilter(final TransactionFilter transactionFilter) {
```
Errorific: This and calling chain above it should be setTransactionFilter

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/MutableProtocolSchedule.java
```diff
@@ -69,4 +71,10 @@ public String listMilestones() {
         .map(spec -> spec.getSpec().getName() + ": " + spec.getBlock())
         .collect(Collectors.joining(", ", "[", "]"));
   }
+
+  @Override
+  public void addTransactionFilterToTransactionValidators(
```
Errorific: Should be setTransactionFilter...

PR #1509, opened 2019-05-29
Log error if clique or ibft have 0 validators in genesis

rain-on:


pantheon/src/main/java/tech/pegasys/pantheon/controller/CliquePantheonControllerBuilder.java
```diff
@@ -121,6 +126,22 @@ protected MiningCoordinator createMiningCoordinator(
         genesisConfig.getConfigOptions(), nodeKeys, privacyParameters);
   }
 
+  @Override
+  protected boolean validateContext(final ProtocolContext<CliqueContext> context) {
+    final Optional<BlockHeader> genesisBlockHeader =
+        context.getBlockchain().getBlockHeader(GENESIS_BLOCK_NUMBER);
+    if (!genesisBlockHeader.isPresent()) {
+      LOG.error("Genesis block cannot be retriefed from chain.");
```
ajsutton:
```suggestion
      LOG.error("Genesis block cannot be retrieved from chain.");
```

ajsutton: I don't think you actually need this check - the blockchain guarantees the genesis block is always available so you could just call `get()` directly.

pantheon/src/main/java/tech/pegasys/pantheon/controller/PantheonControllerBuilder.java
```diff
@@ -299,6 +301,10 @@ protected abstract MiningCoordinator createMiningCoordinator(
 
   protected abstract ProtocolSchedule<C> createProtocolSchedule();
 
+  protected boolean validateContext(final ProtocolContext<C> context) {
+    return true;
```
ajsutton: Remove the return value - nothing appears to use it.

rain-on: yeah .... oops .... was thinking that if the genesis was invalid, that may we'd stop execution ... but, think the log is plenty...

PR #1517, opened 2019-05-31
Dependency Version Updates

shemnon:
## PR description

Release plugin 2.8.0 fails on JDK-12 and Zulu in docker containers.


shemnon:
It appears that 2.8.0 version of the release plugin presumes git is available in the build environment. 
 Even to just configure the project, not to run any release specific tasks.

Errorific:
I'll add something to my queue to add git to the build images and rebump the version of the plugin up.

ajsutton:
I've created a ticket for this in Yeti's backlog. It will be a good way to get the new hires up to speed with how CI is setup.

PR #1519, opened 2019-06-02
[PAN-2547] Modified JSON-RPC subscription processing to avoid blockin…

mark-terry:
…g important threads.

## PR description
Modified how some JSON-RPC subscriptions are being process to avoid blocking important threads (synchronizer etc.)

## Fixed Issue(s)

#PAN-2547


ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/websocket/subscription/blockheaders/NewBlockHeadersSubscriptionService.java
```diff
@@ -38,20 +40,23 @@ public NewBlockHeadersSubscriptionService(
 
   @Override
   public void onBlockAdded(final BlockAddedEvent event, final Blockchain blockchain) {
-    final List<NewBlockHeadersSubscription> subscribers =
-        subscriptionManager.subscriptionsOfType(
-            SubscriptionType.NEW_BLOCK_HEADERS, NewBlockHeadersSubscription.class);
+    final ExecutorService executor = Executors.newSingleThreadExecutor();
```
ajsutton: We shouldn't create a new executor for every block - need to have this `ExecutorService` managed as part of the component life cycle - started when JSON-RPC starts and shutdown when it's shutdown.  Probably simplest to use vertx - either by jumping to one of it's worker queues or onto its event queue.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/websocket/subscription/syncing/SyncingSubscriptionService.java
```diff
@@ -32,15 +34,18 @@ public SyncingSubscriptionService(
   }
 
   private void sendSyncingToMatchingSubscriptions(final SyncStatus syncStatus) {
-    final List<Subscription> syncingSubscriptions =
-        subscriptionManager.subscriptionsOfType(SubscriptionType.SYNCING, Subscription.class);
+    final ExecutorService executor = Executors.newSingleThreadExecutor();
```
ajsutton: Same thing - this should probably be the same `ExecutorService` or again go via vertx.

ethereum/jsonrpc/src/test/java/tech/pegasys/pantheon/ethereum/jsonrpc/websocket/subscription/blockheaders/NewBlockHeadersSubscriptionServiceTest.java
```diff
@@ -84,7 +85,7 @@ public void shouldSendMessageWhenBlockAdded() {
 
     simulateAddingBlock();
 
-    verify(subscriptionManager)
+    verify(subscriptionManager, timeout(100))
```
ajsutton: We really want to avoid using multiple threads in tests - it' makes everything much more complex.  If the `ExecutorService` was passed in (or if vertx is being used, have an interface that abstracts it) then we could use a stub that executes on the same thread rather than deferring to another thread and avoid multiple threads in tests.

If for some reason we do get stuck with multiple threads 100ms is way too short a timeout.  Needs to be more like 10 seconds.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/websocket/subscription/blockheaders/NewBlockHeadersSubscriptionService.java
```diff
@@ -38,20 +38,28 @@ public NewBlockHeadersSubscriptionService(
...
+                        : blockWithTransactionHash(newBlockHash);
 
-      subscriptionManager.sendMessage(subscription.getId(), newBlock);
-    }
+                subscriptionManager.sendMessage(subscription.getId(), newBlock);
+              }
+
+              future.complete();
+            },
+            result -> {});
```
ajsutton: If an exception is thrown from the action, does it wind up being logged or does this result handle wind up silently ignoring it?

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/websocket/subscription/blockheaders/NewBlockHeadersSubscriptionService.java
```diff
@@ -38,20 +38,28 @@ public NewBlockHeadersSubscriptionService(
 
   @Override
   public void onBlockAdded(final BlockAddedEvent event, final Blockchain blockchain) {
-    final List<NewBlockHeadersSubscription> subscribers =
-        subscriptionManager.subscriptionsOfType(
-            SubscriptionType.NEW_BLOCK_HEADERS, NewBlockHeadersSubscription.class);
+    subscriptionManager
+        .getVertx()
+        .<Void>executeBlocking(
```
ajsutton: It would probably be better if instead of `SubscriptionManager` exposing a `getVertx` and then having these classes be directly dependent on vertx,  just expose a method to jump to a worker thread.  That could be a simple `executeOnWorkerThread` that takes a runnable or a really nice API would be to combine the jump to a different thread with looking up the subscribers.  So:
```
subscriptionManager.notifySubscribers(SubscriptionType.NEW_BLOCK_HEADERS, NewBlockHeadersSubscription.class, subscribers -> {
   final Hash new BlockHash = event.getBlock().getHash()
   ...
});
```

That would also avoid having to duplicate this code to jump to another thread for every subscription service.

ethereum/jsonrpc/src/test/java/tech/pegasys/pantheon/ethereum/jsonrpc/websocket/subscription/syncing/SyncingSubscriptionServiceTest.java
```diff
@@ -38,6 +40,7 @@
   @Mock private SubscriptionManager subscriptionManager;
   @Mock private Synchronizer synchronizer;
   private SyncStatusListener syncStatusListener;
+  private final Vertx vertx = Vertx.vertx();
```
ajsutton: You need to close `vertx` in a tear down, but see above for a way to avoid having these classes use vertx directly.  Then you could just have a `when(subscriptionManager.notifySubscribers(...)).thenAnswer(...)` where the answer would invoke the callback immediately so you don't need to deal with multiple threads in the tests.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/websocket/subscription/blockheaders/NewBlockHeadersSubscriptionService.java
mark-terry: Currently ignored, will add logging.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/websocket/subscription/SubscriptionManager.java
```diff
@@ -160,4 +161,22 @@ public void sendMessage(final Long subscriptionId, final JsonRpcResult msg) {
...
+      final Consumer<List<T>> runnable) {
+    vertx.executeBlocking(
+        future -> {
+          final List<T> syncingSubscriptions = subscriptionsOfType(subscriptionType, clazz);
+          runnable.accept(syncingSubscriptions);
+          future.complete();
+        },
+        result -> {
+          if (result.failed()) {
+            LOG.error("Failed to notify subscribers. Cause: {}", result.cause().getMessage());
```
ajsutton: Rather than logging just the message, include the full stack trace.
```suggestion
            LOG.error("Failed to notify subscribers.", result.cause());
```

PR #1527, opened 2019-06-05
Updated contract name and added trusted bootnode behaviour

MadelineMurray:
Updated contract name and added trusted bootnode behaviour


macfarla:
There is also the env var INGRESS_CONTRACT_ADDRESS which is now NODE_INGRESS_CONTRACT_ADDRESS

PR #1528, opened 2019-06-06
Fix estimate gas RPC failing for clique when no blocks have been created

jframe:
## PR description
Fixes issue when using Clique, if the chain height is still at 0 then the estimateGas rpc fails with a internal server error.

This change special cases block 0 so we return address 0 for the proposer in this case instead of failing with an IllegalArgumentException as the genesis block does not have proposer seals.


consensus/clique/src/main/java/tech/pegasys/pantheon/consensus/clique/CliqueBlockHashing.java
```diff
@@ -23,6 +23,7 @@
 import java.util.function.Supplier;
 
 public class CliqueBlockHashing {
+  private static final Address ADDRESS_ZERO = Address.extract(Hash.ZERO);
```
ajsutton: I suspect we should just introduce `Address.ZERO = Address.fromHexString('0x0')` in the `Address` class.

consensus/clique/src/main/java/tech/pegasys/pantheon/consensus/clique/CliqueBlockHashing.java
```diff
@@ -48,6 +49,9 @@ public static Hash calculateDataHashForProposerSeal(
    */
   public static Address recoverProposerAddress(
       final BlockHeader header, final CliqueExtraData cliqueExtraData) {
+    if (header.getNumber() == 0) {
```
ajsutton:
```suggestion
    if (header.getNumber() == BlockHeader.GENESIS_BLOCK_NUMBER) {
```

Please don't ever hard code 0 as the genesis block number. One day we're going to have to support variable genesis block numbers.

consensus/clique/src/main/java/tech/pegasys/pantheon/consensus/clique/CliqueBlockHashing.java
jframe: done

consensus/clique/src/main/java/tech/pegasys/pantheon/consensus/clique/CliqueBlockHashing.java
jframe: done

consensus/clique/src/main/java/tech/pegasys/pantheon/consensus/clique/CliqueBlockHashing.java
```diff
@@ -48,6 +47,9 @@ public static Hash calculateDataHashForProposerSeal(
    */
   public static Address recoverProposerAddress(
       final BlockHeader header, final CliqueExtraData cliqueExtraData) {
+    if (header.getNumber() == BlockHeader.GENESIS_BLOCK_NUMBER) {
```
rain-on: nit: Personally, I'd move this inside the "isPresent" check - just to minimise the conditionals executed on every check ... its probably a minor benefit at best.

jframe: done

PR #1534, opened 2019-06-06
[PAN-2010] format filter id as quantity

RatanRSur:
## PR description
Prevents intermittent `eth_getFilterLogs` failure when the `filterId`
began with a hex 0. Previously leading 0's were not trimmed so depending
on the randomly generated value, a filter could appear to not exist in
the `FilterRegistry`

## Fixed Issue(s)

https://pegasys1.atlassian.net/secure/RapidBoard.jspa?rapidView=28&modal=detail&selectedIssue=PAN-2010


acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/web3j/generated/EventEmitter.java
```diff
@@ -146,18 +146,15 @@ protected EventEmitter(
...
-              }
+            log -> {
+              Contract.EventValuesWithLog eventValues =
+                  extractEventParametersWithLog(STORED_EVENT, log);
+              StoredEventResponse typedResponse = new StoredEventResponse();
+              typedResponse.log = log;
+              typedResponse._to = (String) eventValues.getNonIndexedValues().get(0).getValue();
+              typedResponse._amount =
+                  (BigInteger) eventValues.getNonIndexedValues().get(1).getValue();
+              return typedResponse;
```
RatanRSur: this was a suggested change by intellij, not sure if we wanted the explicitness of the complete anonymous class decl.

RatanRSur: From intellij:
Reports all anonymous classes which can be replaced with lambda expressions.
Note that if an anonymous class is converted into a stateless lambda, the same lambda object can be reused by Java runtime during subsequent invocations. On the other hand, when an anonymous class is used, separate objects are created every time. Thus, applying the quick-fix can cause the semantics change in rare cases, e.g. when anonymous class instances are used as HashMap keys.
Lambda syntax is not supported under Java 1.7 or earlier JVMs.

ajsutton: +1 to using lambdas when possible.  So much easier to read.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/results/Quantity.java
```diff
@@ -48,6 +48,10 @@ public static String create(final long value) {
     return uint256ToHex(UInt256.of(value));
   }
 
+  public static String create(final byte[] value) {
+    return create(new BigInteger(1, value));
+  }
+
   public static String create(final BigInteger value) {
     return uint256ToHex(UInt256.of(value));
```
mbaxter: (Optional) Looks like this class could use a bit of refactoring while we're here.  

We should get rid of `create(final BigInteger value)` and replace usages with the more direct conversion in `format(final BigInteger input)`.  We should also make the method names consistent - either name them all `create` or `format`.

PR #1537, opened 2019-06-07
Add pivot block metrics

ajsutton:
## PR description
Add two metrics:
 * Current pivot block number (allowing monitoring dashboards to report how much further the fast sync chain download has to go)
 * Number of times a pivot block has been selected (allowing you to detect when a world state download stalled and a new pivot block was required).


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/fastsync/FastSyncActions.java
```diff
@@ -44,6 +47,8 @@
   private final EthContext ethContext;
   private final SyncState syncState;
   private final MetricsSystem metricsSystem;
+  private final Counter pivotBlockSelectionCounter;
+  private final AtomicLong pivotBlockGauge = new AtomicLong(0);
```
ajsutton: I actually want a gauge which I can set the value for instead of having to pass in a supplier.  Prometheus metrics support this (in fact it's all it supports natively) but not sure if we should complicate the metrics API to expose that or not.

Normally the pivot block number is only stored in `FastSyncState` which is immutable and just passed as a method argument so we can't create a gauge that reads the value directly from there.

shemnon: What would be nicer is if we could de-register the gauge or make it optional sensitive so the value doesn't report when we are not fast-syncing.  If I ever get back to my labeled gauge work I'll try to remember that as well.

ajsutton: Very true.

PR #1539, opened 2019-06-07
Update ethstats doc

NicolasMassart:
## PR description
- update Network monitor name to EthStats Lite as it's the correct branding
- fixed [WebSocket](https://en.wikipedia.org/wiki/WebSocket) case

## Fixed Issue(s)

fixes PIE-1656

### DIFF
--- a/docs/Alethio/Lite-Network-Monitor.md
+++ b/docs/Alethio/Lite-Network-Monitor.md
@@ -3,21 +3,21 @@ description: Alethio EthStats Lite Network Monitor
 
 # Alethio EthStats Lite Network Monitor
 
-Use the [Alethio EthStats Lite Network Monitor](https://github.com/Alethio?utf8=%E2%9C%93&q=ethstats&type=&language=javascript)
+Use [EthStats Lite](https://github.com/Alethio?utf8=%E2%9C%93&q=ethstats&type=&language=javascript)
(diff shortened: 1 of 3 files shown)


docs/Alethio/Lite-Network-Monitor.md
```diff
@@ -3,21 +3,21 @@ description: Alethio EthStats Lite Network Monitor
 
 # Alethio EthStats Lite Network Monitor
 
-Use the [Alethio EthStats Lite Network Monitor](https://github.com/Alethio?utf8=%E2%9C%93&q=ethstats&type=&language=javascript)
+Use [EthStats Lite](https://github.com/Alethio?utf8=%E2%9C%93&q=ethstats&type=&language=javascript)
```
MadelineMurray: There's still a few instances of 'Network Monitor' on this page that need to be updated to EthStats Lite

NicolasMassart: I wanted to keep the term only once in the title to make sure people know what it's about then call it by the EthStats lite name. Should I add parentheses ?

NicolasMassart: Updated other instances of the name. Just kept it with parentheses in the title and in the description for search engines

PR #1543, opened 2019-06-10
Only increment the added transaction counter if we actually added the transaction

ajsutton:
## PR description
The added transaction counters were far too high because it was counting every transaction we tried to add to the pool, including ones that were already present or that had the same sender and nonce as another but didn't increase the gas limit.


rojotek:
Approved — a test would probably be a good idea.

PR #1546, opened 2019-06-11
[PAN-2009] - Fix cluster clean start after stop in Acceptance tests

usmansaleem:
## PR description
Allow a cluster to be restarted after a stop.

 - Clear bootnodes on cluster start
 - Cleaner logic to pass bootnodes list only to non-boot nodes in cluster startNode
 - Do not shutdown pantheonNodeRunner from stop method
 - Additional call to stop method from close method in Cluster
 - Fix getBootnodes to setBootnodes in NodeConfiguration
 - Initialise PluginContextImpl once per node in ThreadPantheonNodeRunner
    (allows clean start after a close)
 - Stop plugins in stopNode and shutdown methods of ThreadPantheonNodeRunner
 - Add new acceptance test for a cluster to clean start after stop


acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/ProcessPantheonNodeRunner.java
```diff
@@ -46,14 +46,18 @@
...
+  private ExecutorService outputProcessorExecutor = Executors.newCachedThreadPool();
 
   ProcessPantheonNodeRunner() {
     Runtime.getRuntime().addShutdownHook(new Thread(this::shutdown));
   }
 
   @Override
   public void startNode(final PantheonNode node) {
+    if (outputProcessorExecutor == null || outputProcessorExecutor.isShutdown()) {
+      outputProcessorExecutor = Executors.newCachedThreadPool();
```
ajsutton: Why do we need to support creating a new executor service?  Shouldn't it only be shutdown after `close` is called?  I'd have expected `stop` to leave it running and the cluster can be restarted but `close` should be a very final operation and the cluster shouldn't be used after it's called.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/ThreadPantheonNodeRunner.java
```diff
@@ -62,6 +62,19 @@
...
 
+  private PantheonPluginContextImpl buildPluginContext(final PantheonNode node) {
+    PantheonPluginContextImpl pantheonPluginContext = new PantheonPluginContextImpl();
+    final Path pluginsPath = node.homeDirectory().resolve("plugins");
+    final File pluginsDirFile = pluginsPath.toFile();
+    if (!pluginsDirFile.isDirectory()) {
+      pluginsDirFile.mkdirs();
+      pluginsDirFile.deleteOnExit();
+    }
+    System.setProperty("pantheon.plugins.dir", pluginsPath.toString());
```
ajsutton: Why are we setting system properties here?  Different nodes are going to overwrite the property with their own conflicting values.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/pubsub/NewPendingTransactionAcceptanceTest.java
```diff
@@ -70,7 +70,7 @@ public void transactionRemovedByChainReorganisationMustPublishEvent() throws Exc
 
     final Condition atLeastLighterForkBlockNumber = blockchain.blockNumberMustBeLatest(minerNode);
 
-    cluster.stop();
+    cluster.close();
```
ajsutton: Why `close` instead of `stop` here?  I'd have expected `close` to only be called as part of tear down.

pantheon/src/main/java/tech/pegasys/pantheon/services/PantheonPluginContextImpl.java
```diff
@@ -121,9 +121,9 @@ public void registerPlugins(final Path pluginsDir) {
 
   public void startPlugins() {
     checkState(
-        state == Lifecycle.REGISTERED,
+        state == Lifecycle.REGISTERED || state == Lifecycle.STOPPED,
```
ajsutton: I'm a bit suspect about this one.  I'm not sure I would have expected the exact same instance of pantheon or its plugins to be able to be restarted.  I'd have expected the `ThreadedPantheonNodeRunner` to create a new instance to start when its restarted.

ajsutton: If we really do need this then we need to update the documentation in the plugin API repo to include this change to the life cycle.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/pubsub/NewPendingTransactionAcceptanceTest.java
usmansaleem: This method will be removed in next PR ... I'll revert these changes so that it doesn't get reflect in this PR.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/ThreadPantheonNodeRunner.java
usmansaleem: It was just refactored from earlier logic where we are setting it every time we called start node.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/ProcessPantheonNodeRunner.java
usmansaleem: Agreed, this was a last minute addition following the similar pattern as ThreadP*NodeRunner. I will remove this call as well.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/ThreadPantheonNodeRunner.java
ajsutton: oh yeah, I see why that's required.  That's really awful but I don't have a better option.

pantheon/src/main/java/tech/pegasys/pantheon/services/PantheonPluginContextImpl.java
usmansaleem: let me revisit this one.

usmansaleem: @ajsutton I've reverted changes in this class and incorporated suggestions in ThreadedPan..Runner class.

PR #1550, opened 2019-06-12
Make MetricCategories more flexible

ajsutton:
## PR description
Artemis have expressed interest in using our metrics library since they'll need to export to Prometheus but don't necessarily want to be tied to it and want to avoid the very static-heavy approach the Prometheus client library normally uses.

Currently there are a few Pantheon specific details that have leaked into the metrics library - the hardcoded `pantheon_` prefix and `MetricCategories` being an enum full of categories specifically for Pantheon.

Having `MetricCategories` as an enum also prevents future plugins from adding their own categories if we add a metrics service to our plugin APIs.

So there are four major changes in this PR (once we've agreed on the final design we're heading towards I can break it up into smaller PRs to simplify reviews):
1. Move MetricConfiguration to use a builder pattern instead of being mutable and using setters. This was actually unrelated but it just annoyed me...
2. Make `MetricCategory` an interface instead of an enum.  This required getting rid of most usages of `EnumSet` in favour of a plain `Set`.  There are now two enums which implement `MetricCategory`, `StandardMetricCategory` (JVM and PROCESS) which should stay in the metrics package and `PantheonMetricCategory` which needs to find a home outside of the metric package since it's Pantheon specific (not sure where yet).
3. `PantheonCommand` builds up a map of allowed metric category names to their instances.  Currently that's just a `HashMap` which all the values from `StandardMetricCategory` and `PantheonMetricCategory` are added into.  A plugin API could be exposed to allow plugins to register their own categories in the future.
4. Removed the dependency on the `util`.  This was only used in production code to check the network port was valid so I just inlined it.  The tests still have a dependency on it for one method but that won't affect other projects trying to import the metrics library.

Open questions/ugly bits:
 * Where should `PantheonMetricCategory` actually live.  It needs to be available to most of Pantheon. May need to just create a `:metrics:pantheon` module specifically for it, although its slightly tempting to shove it into `util`...
 * The code in `PantheonCommand` to register the metric categories doesn't feel particularly clean. Might be able to extract a class to clean it up.
 * The `MetricSystem` doesn't actually check that the `MetricCategory` supplied when creating a counter/gauge/etc was actually registered back in `PantheonCommand`, although if it wasn't it will just never be enabled.
 * The default list of metrics won't include anything from plugins which seems wrong, although not something that matters until we actually expose a metrics API to plugins.


metrics/core/src/main/java/tech/pegasys/pantheon/metrics/MetricCategory.java
```diff
@@ -12,45 +12,9 @@
...
-  }
-
-  public String getName() {
-    return name;
-  }
-
-  public boolean isPantheonSpecific() {
-    return pantheonSpecific;
-  }
+  boolean isApplicationSpecific();
```
mbaxter: Instead of having a `boolean` here + the `applicationPrefix` configured on the `MetricsSystem`, what about just storing an `Optional<String> applicationPrefix` on `MetricsCategory`?

ajsutton: I was skeptical at first but have come around.  This does simplify things and given an application would typically have one enum for its categories it actually works out pretty nicely.

ajsutton:
Agreed, I'll split it up into a couple of separate PRs to make things clearly - it took a bit of playing to work out what the final destination needed to look like.

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -650,6 +653,11 @@ public void parse(
     commandLine.registerConverter(Wei.class, (arg) -> Wei.of(Long.parseUnsignedLong(arg)));
     commandLine.registerConverter(PositiveNumber.class, PositiveNumber::fromString);
 
+    final MetricCategoryConverter metricCategoryConverter = new MetricCategoryConverter();
+    metricCategoryConverter.addCategories(PantheonMetricCategory.class);
+    metricCategoryConverter.addCategories(StandardMetricCategory.class);
```
shemnon: I may want to expose this via the plugin APIs, but we should wait until some plugin wants it.

ajsutton: Yep agreed.

PR #1553, opened 2019-06-12
[PIE-1612] [PIE-1613] liveness and readiness probe stub endpoints

abdelhamidbakhta:
## PR description
This PR aims to provide responding endpoints to monitor health of Pantheon node when deployed in a Kubernetes context. No real business logic has been implemented in the health check. We consider the node ready as soon as the JSON RPC server is up.

- implement stub readiness endpoint
- implement stub liveness endpoint
- prepare unit tests

## Fixed Issue(s)
#PIE-1612
#PIE-1613


ethereum/jsonrpc/src/test/java/tech/pegasys/pantheon/ethereum/jsonrpc/JsonRpcHttpServiceTest.java
```diff
@@ -2032,6 +2033,22 @@ public void ethGetStorageAtInvalidParameterStorageIndex() throws Exception {
     }
   }
 
+  @Test
+  public void assertThatLivenessProbeWorks() throws Exception {
+    try (final Response resp =
+        client.newCall(buildGetRequest(HealthService.LIVENESS_PATH)).execute()) {
```
NicolasMassart: you main not use the constants defined in the code as values used in the tests. This would prevent you from detecting any issue on the path change for instance. Hardcoding the `/liveness` in the test seems better to me so we can check that we are not going to ship a breaking change for the user, and path change, for whatever reason, is a breaking change.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/health/ReadinessService.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+package tech.pegasys.pantheon.ethereum.jsonrpc.health;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class ReadinessService extends HealthService {
+  private static final Logger LOG = LogManager.getLogger();
+
+  private static final int READY_STATUS_CODE = 200;
```
NicolasMassart: Codes seems to be HTTP codes. Could it be HttpResponseStatus instead of setting you own int directly ?
```suggestion
  private static final int READY_STATUS_CODE = HttpResponseStatus.OK.code();
```
or even store the READY_STATUS_CODE as a HttpResponseStatus type.
Also why not having common response codes (like SUCCESS_STATUS_CODE and FAILURE_STATUS_CODE thaat could apply to both readiness and liveness) handled at the parent abstract class level instead of somewhat duplicating them?

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/health/HealthService.java
```diff
@@ -0,0 +1,56 @@
...
+        return healthCheck.isHealthy();
+      }
+    };
+  }
+
+  public int healthyStatusCode() {
+    return HEALTHY_STATUS_CODE;
+  }
+
+  public int nonHealthyStatusCode() {
```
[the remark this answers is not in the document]
abdelhamidbakhta: Initially it was to override status codes in some cases. But i'll remove them.

PR #1559, opened 2019-06-13
Update privacy quickstart docs

antonydenyer:
## PR description
You need to build the images locally first before running the quick start.


NicolasMassart:
@antonydenyer could you specify on which config (os, vm,...) you encountered this issue with Docker-compose not automatically building the Orion image ?

antonydenyer:
Very strange, just tried again with no images/containers and it does indeed build everything if required. I think I must have gotten into a state where docker thought it had the images but didn't and needed to rebuild/fetch them.

PR #1566, opened 2019-06-14
add build-and-run-pantheon-script

RatanRSur:


RatanRSur:
Nico's comments:
- I thought that this script could be used as an alternative, but not as a replacement.
- The command has to be directly usable, so don’t provide `./run-pantheon <args...>` but one with the most used args, and perhaps these should be default values so you can just run `./run-pantheon`
- Renaming is to run-pantheon perhaps should be better to say `build-and-run` as we know we are in pantheon and i think it’s important to have the build thing clear. You could provide some comments inside the script so people could directly know what it does without the need to read the doc site.

RatanRSur:
- I considered leaving the step by step in but I think that most people don't care about the fact that it's unzipping a `.tar.gz` etc. and if they did they could inspect the script
- <args...> can be empty so it is directly usable as `run-pantheon`, is that what you're talking about?
- I like the rename to `build-and-run` but the reason I left in pantheon was because it would clarify things if run outside of the pantheon directory. It also helps in searching backwards in your bash/zsh history because you can search for pantheon and get all the times you ran it, not just the ones from this script. Would `build-and-run-pantheon` be too long?
- I'm definitely open to adding comments! Will do that now.

NicolasMassart:
- I'm adding @MadelineMurray for the content review part.
- if args can be empty, then provide the command description and an example with no args, we can do tabs in our doc, see what's already made for CLI reference for instance.
- `build-and-run-pantheon`, why not.
- can you start by adding some description in this PR first so we directly know what to review, how to test it and what Jira issue it's related to ?

hmijail:
As commented in the slack thread, it would probably be simpler (and more educative for those who look) to use `./gradlew installDist` instead of all the hand-crafted version checking and untarring...

PR #1573, opened 2019-06-18
Extract MetricsCategory interface

ajsutton:
## PR description
Extract a `MetricCategory` interface instead of having all categories in a single enum.

The vast majority of this is just the rename of the enum to `PantheonMetricCategory`.

Builds on https://github.com/PegaSysEng/pantheon/pull/1572 but getting it lined up...


metrics/core/src/main/java/tech/pegasys/pantheon/metrics/MetricsSystem.java
```diff
@@ -56,6 +56,6 @@ default void createLongGauge(
   Stream<Observation> streamObservations(MetricCategory category);
 
   default Stream<Observation> streamObservations() {
-    return Stream.of(MetricCategory.values()).flatMap(this::streamObservations);
+    return Stream.of(PantheonMetricCategory.values()).flatMap(this::streamObservations);
```
ajsutton: This still ties us to `PantheonMetricCategory` but will be cleaned up in a follow-up PR. Trying to focus just on the rename in this one.

metrics/core/src/main/java/tech/pegasys/pantheon/metrics/prometheus/PrometheusMetricsSystem.java
```diff
@@ -56,7 +59,8 @@
   private final Map<String, LabelledMetric<tech.pegasys.pantheon.metrics.OperationTimer>>
       cachedTimers = new ConcurrentHashMap<>();
 
-  private final EnumSet<MetricCategory> enabledCategories = EnumSet.allOf(MetricCategory.class);
+  private final Set<MetricCategory> enabledCategories =
```
ajsutton: Need to use a `Set` instead of `EnumSet` now.

metrics/core/src/test/java/tech/pegasys/pantheon/metrics/prometheus/PrometheusMetricsSystemTest.java
```diff
@@ -14,32 +14,32 @@
...
 import java.util.Comparator;
-import java.util.EnumSet;
 
 import org.junit.Test;
 
 public class PrometheusMetricsSystemTest {
 
   private static final Comparator<Observation> IGNORE_VALUES =
-      Comparator.comparing(Observation::getCategory)
+      Comparator.<Observation, String>comparing(o -> o.getCategory().getName())
```
ajsutton: Category is no longer an enum so the comparator has to be based off the name instead of the enum ordinal (which is probably a good thing anyway). Not sure why the generics is being difficult but it really wants to be told explicitly.

ajsutton:
Actually, this was a step too far - the change to an interface requires all the detail.  Will just do a simple rename without the enum as the next step.

PR #1575, opened 2019-06-18
[PAN-2786] Stop Transaction Pool Queue from Growing Unbounded

abdelhamidbakhta:
## PR description
- update 'MAX_PENDING_TRANSACTIONS' from 4096 to 1024
- use a retention period in seconds instead of hours (default to 60 seconds)
- update documentation
- update tests
## Fixed Issue(s)

fixes #PAN-2786


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/transactions/PendingTransactions.java
```diff
@@ -105,11 +111,18 @@ public PendingTransactions(
...
+        pendingTransactions.values().stream()
+            .filter(
+                transaction -> transaction.getAddedToPoolAt().isBefore(removeTransactionsBefore));
+    final AtomicInteger txCounter = new AtomicInteger(0);
+    transactionsToRemove.forEach(
+        transaction -> {
+          removeTransaction(transaction.getTransaction());
+          txCounter.incrementAndGet();
+        });
+    LOG.debug("{} transactions evicted from the pool.", txCounter.get());
```
RatanRSur: nice touch adding logging

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/transactions/PendingTransactions.java
```diff
@@ -51,10 +56,11 @@
  */
 public class PendingTransactions {
 
-  public static final int MAX_PENDING_TRANSACTIONS = 4096;
-  public static final int DEFAULT_TX_RETENTION_HOURS = 13;
+  private static final Logger LOG = LogManager.getLogger();
+  public static final int MAX_PENDING_TRANSACTIONS = 1024;
+  public static final int DEFAULT_TX_RETENTION_SECONDS = 60;
```
RatanRSur: How did you decide on these constants?

abdelhamidbakhta: Arbitrarily. The idea is to reduce the max pending transactions so i divided the max size by 4. And for the retention period it was mentioned in the JI description. Adrian suggested to evict transactions older than 1 minute. By the way the two are configurable. I have worked on the default values because often they are not set.

mbaxter: I think this change will be problematic for local transaction propagation.  It can take us a while to find peers, so a 1 minute time window could mean local transactions get dropped before we have a chance to broadcast them to anyone.

abdelhamidbakhta: You are right. Should we have different eviction policies for local and remote transactions ?

RatanRSur: If there are no peers yet, there are only local transactions, correct? Therefore, instead of having different eviction policies, could we make time-based eviction only occur after the `MAX_PENDING_TRANSACTIONS` has been reached?

abdelhamidbakhta: It seems that i misunderstood Adrian's suggestion. It was more about processing incoming messages from the network that add transactions to the pool. We can still use the eviction mechanism to fix this but we should probably also limit the number of messages we handle asynchronously.

ajsutton:
Yeah this is not the change we want. :) The transactions *in* `PendingTransactions` are great, please don't reduce the number of retention period. Doing so would actually make it worse as we're then more likely to get incoming messages that aren't still in the pool which then requires us to validate them.

We need to limit the queue size on the transactions executor we use to process incoming transactions messages from peers.

abdelhamidbakhta:
Not relevant.

PR #1578, opened 2019-06-18
[PAN-2630] Synchronizer should disconnect the sync target peer on invalid block data

abdelhamidbakhta:
## PR description
If we receive invalid blocks from the network, we should disconnect the sync target peer for guiding us to those bad blocks.  
- check if headers are sequential
- if sequential, check if they form a chain, if not, disconnect the peer


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/PipelineChainDownloader.java
```diff
@@ -121,6 +123,10 @@ public synchronized void cancel() {
       // Allowing the normal looping logic to retry after a brief delay.
       return scheduler.scheduleFutureTask(() -> completedFuture(null), PAUSE_AFTER_ERROR_DURATION);
     }
+    if (ExceptionUtils.rootCause(error) instanceof InvalidBlockException) {
+      syncTarget.peer().disconnect(DisconnectMessage.DisconnectReason.BREACH_OF_PROTOCOL);
```
mbaxter: Can we add a test that hits this case?

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/messages/BlockHeadersMessage.java
```diff
@@ -25,7 +25,7 @@
 import java.util.Arrays;
 import java.util.List;
 
-public final class BlockHeadersMessage extends AbstractMessageData {
+public class BlockHeadersMessage extends AbstractMessageData {
```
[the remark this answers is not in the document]
abdelhamidbakhta: It was for mocking purpose. I found another way to test it without mocking so i will undo this change.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/PipelineChainDownloader.java
abdelhamidbakhta: Done.

PR #1580, opened 2019-06-19
[PAN-2811] Be more lenient with discovery message deserialization

mbaxter:
## PR description
[EIP-8](https://eips.ethereum.org/EIPS/eip-8) specifies that devp2p discovery message deserialization should be lenient in order to make protocol upgrades easier.  This PR updates discovery packet deserialization to be more lenient by:
* ignoring extra fields
* ignoring the version in PING packets

### DIFF
--- a/ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/FindNeighborsPacketData.java
+++ b/ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/FindNeighborsPacketData.java
@@ -52,7 +52,7 @@ public class FindNeighborsPacketData implements PacketData {
     in.enterList();
     final BytesValue target = in.readBytesValue();
     final long expiration = in.readLongScalar();
-    in.leaveList(true);
+    in.leaveListLenient();
     return new FindNeighborsPacketData(target, expiration);
   }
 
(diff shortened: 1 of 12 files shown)


ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/discovery/internal/FindNeighborsPacketData.java
```diff
@@ -52,7 +52,7 @@ public static FindNeighborsPacketData readFrom(final RLPInput in) {
     in.enterList();
     final BytesValue target = in.readBytesValue();
     final long expiration = in.readLongScalar();
-    in.leaveList();
+    in.leaveList(true);
```
RatanRSur: (optional) make leaveList parameter an enum. Much easier to understand what
`in.leaveList(RLPInput.IgnoreRest)` does as opposed to `in.leaveList(true)`

shemnon: I'd say either @RatanRSur's suggestion or at least leave a comment that (true) means to ignore the rest of the list, for all the `leaveList(true)` calls.

ajsutton: Third option would be to have two methods - the usual `leaveList` and the new `leaveListSkippingRemainingElements`

mbaxter: Changed this to `leaveListLenient()`

PR #1586, opened 2019-06-20
[PAN-2786] Stop Transaction Pool Queue from Growing Unbounded

abdelhamidbakhta:
## PR description
The time based policy has been removed. A follow up PR will be created to include a timeout in `TransactionsMessageProcessor` and to skip message if the message has expired.
- implement a custom bounded queue
- ~~use a time based policy with keep alive configuration~~
- ~~implement eviction process based on the policy~~
- add metrics
- expose a method in `MonitoredExecutors` to create a working queue with a maximum capacity
- update `EthScheduler` to use a limited working queue for the `txWorkerExecutor`


ajsutton:
Wouldn’t this cause exceptions to be thrown or the  vertx/netty thread to block once the executor queue fills up?  We’d want to discard the oldest messages when the queue is full because they’re likely full off outdated transactions and we want to process the most recent. 

We also want to ensure there’s good logging and metrics to indemnify when this happens since losing transactions is a bad thing for a lot of cases. 

Finally the queue size of 128 is far too small. Take a look at the queue sizes and incoming message rate for the MainNet Fast i3.large box - it gets up to 500 incoming messages/sec and only hits trouble when the queue size is in the millions. We probably don’t want to let it get that big as the messages are likely out of date by then anyway but it needs to be a few thousand at least I’d say. Having an idea of how long messages were in that queue would help to set the size appropriately.

abdelhamidbakhta:
> Wouldn’t this cause exceptions to be thrown or the vertx/netty thread to block once the executor queue fills up? We’d want to discard the oldest messages when the queue is full because they’re likely full off outdated transactions and we want to process the most recent.
> 
> We also want to ensure there’s good logging and metrics to indemnify when this happens since losing transactions is a bad thing for a lot of cases.
> 
> Finally the queue size of 128 is far too small. Take a look at the queue sizes and incoming message rate for the MainNet Fast i3.large box - it gets up to 500 incoming messages/sec and only hits trouble when the queue size is in the millions. We probably don’t want to let it get that big as the messages are likely out of date by then anyway but it needs to be a few thousand at least I’d say. Having an idea of how long messages were in that queue would help to set the size appropriately.

Yes indeed it will cause exceptions to be thrown once the executor queue fills up.
Ok for the default size too small, i will increase it to few thousand. A next step will be to have this number configurable in a follow up PR.
Regarding discarding oldest messages, where is the best place to put this logic ? In the `EthScheduler` or in the 'TransactionsMessageHandler'. If we do this in `TransactionsMessageHandler` it will require to expose the list of pending tasks from the `EthScheduler`. And we probably would have to use a custom `BlockingQueue` to implement the logic based on the timestamp.

ajsutton:
I’d say use a bounded queue with a really big size (a million or two), include the time stamp of when the message was first received in the queues task to execute and then when the executor runs the task the first thing it does is check it it’s too old and bail out if so.

The speed difference between checking a time stamp and actually validating a transaction is so great that this mechanism will allow catching up on backlogs very quickly without anything too complex. And if for some reason we’re getting so many incoming messages that we can’t even do a long comparison on each fast enough then the client has no hope, we’ll hit the queue size limit and throw loads of exceptions. The client is unlikely to cope at that point but at least we have a lot of error messages telling us why. Definitely worth checking it throws an exception if the queue is full instead of blocking.


But the idea of using a custom queue is possibly a good one.  If we did that, it would be easy for it to detect that the size limit is reached and remove from the front of the list when adding an item to the end.  Logging and monitoring could all be inside the custom queue implementation then too.

abdelhamidbakhta:
> I’d say use a bounded queue with a really big size (a million or two), include the time stamp of when the message was first received in the queues task to execute and then when the executor runs the task the first thing it does is check it it’s too old and bail out if so.
> 
> The speed difference between checking a time stamp and actually validating a transaction is so great that this mechanism will allow catching up on backlogs very quickly without anything too complex. And if for some reason we’re getting so many incoming messages that we can’t even do a long comparison on each fast enough then the client has no hope, we’ll hit the queue size limit and throw loads of exceptions. The client is unlikely to cope at that point but at least we have a lot of error messages telling us why. Definitely worth checking it throws an exception if the queue is full instead of blocking.
> 
> But the idea of using a custom queue is possibly a good one. If we did that, it would be easy for it to detect that the size limit is reached and remove from the front of the list when adding an item to the end. Logging and monitoring could all be inside the custom queue implementation then too.

I have implemented a custom queue that use a time based policy when adding an element at full capacity. A keep alive parameter is used to evict transactions that have been in the working queue too much time.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/manager/MockExecutorService.java
```diff
@@ -55,6 +56,10 @@ public void runPendingFuturesInSeparateThreads(final ExecutorService executorSer
     currentTasks.forEach(task -> executorService.execute(task::run));
   }
 
+  public List<ExecutorTask<?>> getTasks() {
```
mbaxter: I don't think you need to expose this?  You can just use the existing `getPendingFuturesCount()` or `getFutures()`.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/manager/EthSchedulerShutdownTest.java
```diff
@@ -85,8 +87,8 @@ public void shutdown_txWorkerShutsDown() throws InterruptedException {
     final MockEthTask task1 = new MockEthTask(1);
     final MockEthTask task2 = new MockEthTask();
 
-    ethScheduler.scheduleTxWorkerTask(task1::executeTask);
-    ethScheduler.scheduleTxWorkerTask(task2::executeTask);
+    ethScheduler.scheduleTxWorkerTask(BoundedTimedTask.fromRunnable(task1::executeTask));
```
mbaxter: Why not move this wrapping into `scheduleTxWorkerTask`?

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/bounded/TimedTask.java
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package tech.pegasys.pantheon.ethereum.eth.manager.bounded;
+
+import java.time.Duration;
+import java.time.Instant;
+
+public interface TimedTask extends Runnable {
```
mbaxter: (Optional) Consider merging this with the concrete implementation `BoundedTimedTask`.  If we don't anticipate multiple implementations, it's simpler to just have a single concrete class.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/bounded/BoundedTimedQueue.java
```diff
@@ -0,0 +1,163 @@
...
+      final TimedTask task = (TimedTask) runnable;
+      if (!isEligible(task)) {
+        LOG.debug("Task does not match time policy, rejecting task.");
+        return false;
+      }
+      final boolean offered = super.offer(task);
+      if (offered) {
+        return true;
+      }
+      // Try to evict tasks based on time policy.
```
mbaxter: So, we'll only start evicting when we hit 1,000,000 items?  Not sure how well that will work out in practice - seems like at that point we might be overwhelmed?  And then we may end up blocking for a while as we empty out the task queue.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/bounded/BoundedTimedQueue.java
```diff
@@ -0,0 +1,163 @@
...
+    this.keepAliveUnit = keepAliveUnit;
+    this.evictionPollTimeout = evictionPollTimeout;
+    this.evictionPollTimeoutUnit = evictionPollTimeoutUnit;
+    this.metricsSystem = metricsSystem;
+    this.metricName = this.getClass().getSimpleName().toLowerCase(Locale.US).replace('-', '_');
+    this.totalEvictedTaskCounter =
+        this.metricsSystem.createCounter(
+            PantheonMetricCategory.EXECUTORS,
+            metricName + "_rejected_tasks_total",
+            "Total number of tasks rejected by this working queue.");
```
ajsutton: We probably need to distinguish between different instances of these queues. e.g. if in the future we also used a bounded queue for other types of messages, we'd want to know how many transaction messages were dropped vs how many GET_DATA requests.

We already have a `metricName` passed into `MonitoredExecutors` which can be passed down for this.  Following the same pattern as is used there the metric probably becomes `metricName + "_dropped_tasks_total"`.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/bounded/BoundedTimedQueue.java
```diff
@@ -0,0 +1,163 @@
...
+
+  private final long keepAlive;
+  private final TemporalUnit keepAliveUnit;
+  private final long evictionPollTimeout;
+  private final TimeUnit evictionPollTimeoutUnit;
+  private final MetricsSystem metricsSystem;
+  private final String metricName;
+  private final Counter totalEvictedTaskCounter;
+
+  public BoundedTimedQueue() {
```
ajsutton: This constructor is unused and can be removed.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/bounded/BoundedTimedQueue.java
```diff
@@ -0,0 +1,163 @@
...
+      }
+      final boolean offered = super.offer(task);
+      if (offered) {
+        return true;
+      }
+      // Try to evict tasks based on time policy.
+      int evictedTaskCounter = processEviction();
+      if (evictedTaskCounter == 0) {
+        // No task has been evicted, rejecting new task.
+        LOG.warn(
```
ajsutton: We need some kind of rate limiting for log messages when messages are dropped - otherwise we'll try to print hundreds of thousands of log messages which will quickly become the performance bottleneck and make things worse.

I'm not entirely sure what the right policy is for this is - we do want to make it visible that something is going wrong, but definitely need to avoid logging for every dropped message.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/bounded/TimedTask.java
ajsutton: +1 mostly in Pantheon we avoid interfaces with a single implementation (unless we need to deliberately narrow the exposed API).  That said, this interface will disappear completely when we remove the timing aspects from this PR.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/EthScheduler.java
```diff
@@ -62,12 +65,33 @@ public EthScheduler(
       final int txWorkerCount,
       final int computationWorkerCount,
       final MetricsSystem metricsSystem) {
+    this(
+        syncWorkerCount,
+        txWorkerCount,
+        BoundedTimedQueue.Default.CAPACITY,
```
ajsutton: I'd make the capacity explicitly transaction specific - we should be selecting a custom queue size that suits the use case every time we create a bounded queue, not just using a general default.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/bounded/BoundedTimedQueue.java
```diff
@@ -0,0 +1,163 @@
...
+   */
+  private boolean isEligible(final TimedTask task) {
+    return !isTooOld(task);
+  }
+
+  private boolean isTooOld(final TimedTask task) {
+    return task.age().get(keepAliveUnit) >= keepAlive;
+  }
+
+  public static final class Default {
```
ajsutton: This internal class is redundant - typically in Pantheon we'd just make these public static fields on the `BoundedQueue` class.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/bounded/BoundedTimedQueue.java
```diff
@@ -0,0 +1,163 @@
...
+import java.time.temporal.ChronoUnit;
+import java.time.temporal.TemporalUnit;
+import java.util.Locale;
+import java.util.concurrent.LinkedBlockingDeque;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class BoundedTimedQueue extends LinkedBlockingDeque<Runnable> {
```
ajsutton: Can we have some unit tests covering this class please?

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/manager/MockExecutorService.java
abdelhamidbakhta: Done.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/manager/EthSchedulerShutdownTest.java
abdelhamidbakhta: Done.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/bounded/BoundedTimedQueue.java
abdelhamidbakhta: I see your point. However how to decide the number of messages to drop if we remove the time criteria. I mean when there is no more space in the queue the current eviction process remove tasks from the head until a non expired task is found. Should we use a fixed number of messages to drop ?

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/bounded/BoundedTimedQueue.java
abdelhamidbakhta: Done.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/bounded/BoundedTimedQueue.java
abdelhamidbakhta: Done.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/EthScheduler.java
abdelhamidbakhta: Done.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/bounded/BoundedTimedQueue.java
abdelhamidbakhta: Done.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/bounded/BoundedTimedQueue.java
```diff
@@ -0,0 +1,163 @@
...
+      final long evictionPollTimeout,
+      final TimeUnit evictionPollTimeoutUnit,
+      final MetricsSystem metricsSystem) {
+    super(capacity);
+    this.keepAlive = keepAlive;
+    this.keepAliveUnit = keepAliveUnit;
+    this.evictionPollTimeout = evictionPollTimeout;
+    this.evictionPollTimeoutUnit = evictionPollTimeoutUnit;
+    this.metricsSystem = metricsSystem;
+    this.metricName = this.getClass().getSimpleName().toLowerCase(Locale.US).replace('-', '_');
```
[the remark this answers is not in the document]
abdelhamidbakhta: Done.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/bounded/BoundedTimedQueue.java
abdelhamidbakhta: Done.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/bounded/BoundedTimedQueue.java
abdelhamidbakhta: Discussed internally. We decided to remove the time based policy. We make a room for a new element when we reach max capacity.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/bounded/BoundedQueue.java
```diff
@@ -0,0 +1,61 @@
...
+            PantheonMetricCategory.EXECUTORS,
+            metricName + "_dropped_tasks_total",
+            "Total number of tasks rejected by this working queue.");
+  }
+
+  @Override
+  public boolean offer(final Runnable task) {
+    try {
+      if (!super.offer(task)) {
+        poll(evictionPollTimeout, evictionPollTimeoutUnit);
```
ajsutton: We shouldn't need to use `poll` and have a timeout here.  If the queue is full there must be an item to remove and if there isn't something is very, very wrong.  So we can just use `remove()` and get rid of the related config options.

There's also a race condition here - if two threads are calling `offer` at the same time, one may be rejected, remove an item to make room but then the other thread fills the new spot, which results in this one being rejected incorrectly.

I suspect this is the place where we want to use a tight loop - as long as `capacity` is greater than 0, eventually we'll be able to remove an item and fill the available slot before another thread does.  So something like:

```
while (!super.offer(task)) {
  remove();
}
return true;
```

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/manager/bounded/BoundedQueueTest.java
```diff
@@ -0,0 +1,35 @@
...
+import tech.pegasys.pantheon.metrics.noop.NoOpMetricsSystem;
+
+import java.util.concurrent.TimeUnit;
+
+import org.junit.Test;
+
+public class BoundedQueueTest {
+
+  @Test
+  public void offerShouldMakeARoomAndAcceptNewElementAtFullCapacity() {
```
ajsutton: We should probably have some tests for cases where it doesn't need to remove an item.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/manager/MockExecutorService.java
```diff
@@ -139,7 +140,7 @@ public void execute(final Runnable command) {
     submit(command);
   }
 
-  private static class ExecutorTask<T> {
+  public static class ExecutorTask<T> {
```
ajsutton: nit: I don't think this needs to be made public now.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/manager/EthSchedulerTest.java
```diff
@@ -204,4 +207,19 @@ public void timeout_cancelsTaskWhenResultIsCancelled() {
...
+  @Test
+  public void scheduleTxWorkerTask_completesWhenScheduledTaskCompletes() {
+    final MockRunnable runnable = Mockito.mock(MockRunnable.class);
+    final MockExecutorService mockTxWorker = ethScheduler.txWorkerExecutor();
+    ethScheduler.scheduleTxWorkerTask(runnable);
+    MatcherAssert.assertThat(mockTxWorker.getFutures(), hasSize(1));
+    verify(runnable).run();
+  }
+
+  static class MockRunnable implements Runnable {
```
ajsutton: This class is redundant - just mock `Runnable.class` directly.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/manager/EthSchedulerTest.java
```diff
@@ -204,4 +207,19 @@ public void timeout_cancelsTaskWhenResultIsCancelled() {
     assertThat(task.isSucceeded()).isFalse();
     assertThat(task.isCancelled()).isTrue();
   }
+
+  @Test
+  public void scheduleTxWorkerTask_completesWhenScheduledTaskCompletes() {
```
ajsutton: This name is misleading - there's no `Future` returned so nothing to complete.  It seems to just be testing that the mock executor immediately calls `run`.  As far as I can tell we're really just testing stubs so could just remove this test.

Unless I've missed something?

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/EthScheduler.java
```diff
@@ -45,9 +47,9 @@
   private static final Logger LOG = LogManager.getLogger();
 
   private final Duration defaultTimeout = Duration.ofSeconds(5);
-
   private final AtomicBoolean stopped = new AtomicBoolean(false);
   private final CountDownLatch shutdown = new CountDownLatch(1);
+  private static final int txWorkerCapacity = 1000000;
```
ajsutton: nit: constants should use all uppercase names
```suggestion
  private static final int TX_WORKER_CAPACITY = 1000000;
```

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/manager/bounded/BoundedQueue.java
abdelhamidbakhta: Ok it makes sense. Done.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/manager/MockExecutorService.java
abdelhamidbakhta: Done.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/manager/EthSchedulerTest.java
abdelhamidbakhta: Yeah this test is not relevant anymore.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/manager/EthSchedulerTest.java
abdelhamidbakhta: Removed.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/manager/bounded/BoundedQueueTest.java
abdelhamidbakhta: Test added.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/manager/DeterministicEthScheduler.java
```diff
@@ -48,6 +48,10 @@ public DeterministicEthScheduler(final TimeoutPolicy timeoutPolicy) {
             (MockExecutorService) this.computationExecutor);
   }
 
+  public MockExecutorService txWorkerExecutor() {
```
ajsutton: nit: Is this now unused?

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/manager/bounded/BoundedQueueTest.java
```diff
@@ -0,0 +1,42 @@
...
+    final BoundedQueue queue = new BoundedQueue(size, "test", new NoOpMetricsSystem());
+    for (int i = 0; i < size; i++) {
+      assertTrue(queue.offer(() -> {}));
+      assertEquals(i + 1, queue.size());
+    }
+  }
+
+  @Test
+  public void offerShouldMakeARoomAndAcceptNewElementAtFullCapacity() {
+    final BoundedQueue queue = new BoundedQueue(1, "test", new NoOpMetricsSystem());
```
ajsutton: Can we update this to have a queue size of 2 or 3 so that we can actually assert that the oldest entry was removed when it filled up?  The default instance equality of `Runnable` will be enough to check which are still in the queue if you store them in variables:
```
Runnable task1 = () -> {};
Runnable task2 = () -> {};
...
assertThat(queue).contains(task1);
assertThat(queue).doesNotContain(task2);
```

And one more little nit - we try to use AssertJ's `Assertion` instead of the Unit `assertTrue`/`assertEquals` as it reads better and gives much better failure messages.

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/manager/DeterministicEthScheduler.java
abdelhamidbakhta: yes it is. i remove it

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/manager/bounded/BoundedQueueTest.java
abdelhamidbakhta: Ok. i do it and i update to use AssertJ.

abdelhamidbakhta:
> LGTM. Just a couple of little nits that would be good to tidy up. Thanks for your patience on this one, timezones have made it take a long time - sorry.

No worries. Thank you for the time you spent on it.

PR #1597, opened 2019-06-24
[PAN-2832] Support setting config options via environment variables

ajsutton:
## PR description
Supports setting config options via environment variables.  Each CLI config option has an equivalent environment variable which is all upper-case, - becomes _ and a `PANTHEON_` prefix is added.  So `--miner-coinbase` can be set via `PANTHEON_MINER_COINBASE`.

When the same option is set via multiple means, Pantheon will use the priority order: CLI arg, env var, config file.


pantheon/src/test/java/tech/pegasys/pantheon/cli/EnvironmentVariableDefaultProviderTest.java
```diff
@@ -0,0 +1,53 @@
...
+    assertThat(provider.defaultValue(OptionSpec.builder("--env-var", "--env-var-set").build()))
+        .isEqualTo("abc");
+  }
+
+  @Test
+  public void shouldNotReturnValueForShortOptions() {
+    environment.put("PANTHEON_H", "abc");
+    assertThat(provider.defaultValue(OptionSpec.builder("-h").build())).isNull();
+  }
+}
```
joshuafernandes: :+1:

PR #1599, opened 2019-06-24
[PAN-2871] Columnar rocksdb

RatanRSur:
## PR description
This PR builds upon the mark/sweep PoC to introduce a multi-column database for different types of node storage needs, which allows for iteration over those types.

Changes from PoC:
* `ColumnarRocksDbStorageTest.java`
  * Extended existing tests which test **single** segment of multi-segment db
  * Added multi-segment specific tests, including iteration
* changed visibility of `createSegmentedProvider`
* misc. updates to reflect changes in master since the PoC
  * `MetricsCategory` -> `PantheonMetricsCategory`
  * use new static `Subscribers` constructor
* rename `KeyValueStorageWorldStateStorage` to `WorldStateKeyValueStorage` for consistency
## Fixed Issue(s)

https://pegasys1.atlassian.net/secure/RapidBoard.jspa?rapidView=28&modal=detail&selectedIssue=PAN-2817


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/storage/keyvalue/RocksDbStorageProvider.java
```diff
@@ -12,22 +12,83 @@
...
+      return createSegmentedProvider(rocksDbConfiguration, metricsSystem);
+    } else {
+      return createUnsegmentedProvider(rocksDbConfiguration, metricsSystem);
+    }
+  }
+
+  private static StorageProvider createUnsegmentedProvider(
+      final RocksDbConfiguration rocksDbConfiguration, final MetricsSystem metricsSystem)
+      throws IOException {
     Files.createDirectories(rocksDbConfiguration.getDatabaseDir());
```
ajsutton: We probably should have an info log here to indicate that we're using the original format.  Maybe even a warning that an old format is being used (not at all sure about that...)

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/storage/keyvalue/RocksDbStorageProvider.java
```diff
@@ -12,22 +12,83 @@
...
+  private enum RocksDbSegment implements Segment {
+    BLOCKCHAIN(new byte[] {1}),
+    WORLD_STATE(new byte[] {2}),
+    PRIVATE_TRANSACTIONS(new byte[] {3}),
+    PRIVATE_STATE(new byte[] {4}),
+    PRUNING_STATE(new byte[] {5});
+
+    private final byte[] id;
+
+    RocksDbSegment(final byte[] id) {
```
ajsutton: nit: would be tidier if we made this accept `final byte... id` then the array would be created automatically and the enum declarations would just be `BLOCKCHAIN(1), WORLD_STATE(2)` etc.  Cheating but kind of nice. :)

services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/RocksDbConfiguration.java
```diff
@@ -108,6 +115,15 @@ public String getLabel() {
...
     int backgroundThreadCount;
 
+    @CommandLine.Option(
+        names = {"--Xrocksdb-columns-enabled"},
+        hidden = true,
+        defaultValue = "true",
+        paramLabel = "<BOOLEAN>",
+        description =
+            "Whether to separate chain and world state into separate RocksDB columns (default: ${DEFAULT-VALUE})")
+    boolean useColumns = true;
```
ajsutton: Since we haven't done the versioning as part of this PR (good choice), we need to set the default to `false`. :)

Once we do have database versioning, we should change this to allow specifying a version rather than just toggling columns on/off.  That will be quite a useful toggle for testing in the future I suspect.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/storage/keyvalue/RocksDbStorageProvider.java
RatanRSur: The warning is already an acceptance criterion of the database versioning task

services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/ColumnarRocksDbKeyValueStorage.java
```diff
@@ -0,0 +1,326 @@
...
+
+  private final OperationTimer readLatency;
+  private final OperationTimer removeLatency;
+  private final OperationTimer writeLatency;
+  private final OperationTimer commitLatency;
+  private final Counter rollbackCount;
+  private final Statistics stats;
+  private final Map<String, ColumnFamilyHandle> columnHandlesByName;
+
+  public static SegmentedKeyValueStorage<ColumnFamilyHandle> create(
```
shemnon: Should this just be `ColumnarRocksDbKeyValueStorage` for the return value?

services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/ColumnarRocksDbKeyValueStorage.java
```diff
@@ -0,0 +1,326 @@
...
+          metricsSystem
+              .createLabelledTimer(
+                  PantheonMetricCategory.KVSTORE_ROCKSDB,
+                  "commit_latency_seconds",
+                  "Latency for commits to RocksDB.",
+                  "database")
+              .labels(rocksDbConfiguration.getLabel());
+
+      if (metricsSystem instanceof PrometheusMetricsSystem) {
+        RocksDBStats.registerRocksDBMetrics(stats, (PrometheusMetricsSystem) metricsSystem);
```
shemnon: Do the rocks db stats need to be made aware of the column data store?  i.e. will the keys we fish for match up?

services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/ColumnarRocksDbKeyValueStorage.java
```diff
@@ -0,0 +1,326 @@
...
+      }
+
+      metricsSystem.createLongGauge(
+          PantheonMetricCategory.KVSTORE_ROCKSDB,
+          "rocks_db_table_readers_memory_bytes",
+          "Estimated memory used for RocksDB index and filter blocks in bytes",
+          () -> {
+            try {
+              return db.getLongProperty("rocksdb.estimate-table-readers-mem");
```
shemnon: Off topic for another PR - can we get the estimated DB size from another property?  Even if it is a different metric that size on disk?

services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/InMemoryKeyValueStorage.java
```diff
@@ -39,11 +41,43 @@
...
+    final Lock lock = rwLock.writeLock();
+    lock.lock();
+    try {
+      hashValueStore.clear();
+    } finally {
+      lock.unlock();
+    }
+  }
+
+  public Set<BytesValue> keySet() {
```
shemnon: Do we want a snapshot of the keyset or one that morphs as the underlying datastore morphs?  That would invalidate iteration too if the state changes.

services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/RocksDbKeyValueStorage.java
```diff
@@ -154,6 +156,41 @@ public Transaction startTransaction() throws StorageException {
...
+      throw new StorageException(e);
+    }
+    return removedNodeCounter;
+  }
+
+  @Override
+  public void clear() {
+    try (final RocksIterator rocksIterator = db.newIterator()) {
+      rocksIterator.seekToFirst();
+      if (rocksIterator.isValid()) {
```
shemnon: I think a guard blocks would look cleaner here, instead of a nested check.

```
      if (!rocksIterator.isValid()) {
        return;
      }
      final byte[] firstKey = rocksIterator.key();
      rocksIterator.seekToLast();
      if (!rocksIterator.isValid()) {
        return;
      }
      db.deleteRange(firstKey, rocksIterator.key());
```

services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/ColumnarRocksDbKeyValueStorage.java
RatanRSur: I don't know but there doesn't seem to be a mention of columns here: https://github.com/facebook/rocksdb/wiki/Statistics. Is there an example statistic you had in mind?

services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/InMemoryKeyValueStorage.java
RatanRSur: It would be consistent with the other kvstores to have snapshot iteration, which I believe this is. Any motivating examples for the alternative?

services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/ColumnarRocksDbKeyValueStorage.java
shemnon: I was thinking out loud to see if stuff like caching was done per column.  If the docs don't mention it then I don't think it is done.

services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/InMemoryKeyValueStorage.java
shemnon: If we want a point in time we will need to copy what is returned from `keySet` instead of just wrapping it in unmodifiableSet.  The wrapper only prevents methods from adding/removing, not from the set being wrapped changing, which is what we would see here.

services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/ColumnarRocksDbKeyValueStorage.java
shemnon: PAN-2853, public and in the backlog.  Probably a good first issue for an external contributor.

PR #1603, opened 2019-06-25
Revert Reason

pinges:
## PR description
This will include the revert reason associated with revert evm operation as an additional field in the transaction receipt when retrieved via the eth_getTransactionReceipt rpc call.

## Fixed Issue(s)

If available, add the revert reason thrown inside a contract to the Transaction Receipt returned via a rpc calls.


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/TransactionReceipt.java
```diff
@@ -201,6 +239,10 @@ public TransactionReceiptType getTransactionReceiptType() {
     return transactionReceiptType;
   }
 
+  public String getRevertReason() {
+    return revertReason.isPresent() ? revertReason.get() : null;
```
jframe: can be simpler using revertReason.orElse(null)

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/MainnetProtocolSpecs.java
```diff
@@ -258,12 +258,14 @@ private MainnetProtocolSpecs() {}
 
   private static TransactionReceipt frontierTransactionReceiptFactory(
       final TransactionProcessor.Result result, final WorldState worldState, final long gasUsed) {
-    return new TransactionReceipt(worldState.rootHash(), gasUsed, result.getLogs());
+    return new TransactionReceipt(
+        worldState.rootHash(), gasUsed, result.getLogs(), result.getRevertReason());
```
ajsutton: This is probably good belt and braces stuff, but for the record - there are no revert reasons in frontier, the REVERT opcode was only added in Byzantium.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/TransactionReceipt.java
```diff
@@ -137,15 +168,22 @@ public static TransactionReceipt readFrom(final RLPInput input) {
       // TODO consider validating that the logs and bloom filter match.
       final LogsBloomFilter bloomFilter = LogsBloomFilter.readFrom(input);
       final List<Log> logs = input.readList(Log::readFrom);
+      final Optional<String> revertReason;
+      if (input.isEndOfCurrentList()) {
+        revertReason = Optional.empty();
+      } else {
+        final byte[] bytes = input.readBytesValue().getArrayUnsafe();
+        revertReason = Optional.of(new String(bytes, StandardCharsets.UTF_8));
```
ajsutton: We need a separate method which checks if a revert reason was available and parses it.  When we receive receipts over the wire, it's a protocol violation if they have the extra revert reason included and should be rejected.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/TransactionReceipt.java
```diff
@@ -115,7 +144,9 @@ public void writeTo(final RLPOutput out) {
     out.writeLongScalar(cumulativeGasUsed);
     out.writeBytesValue(bloomFilter.getBytes());
     out.writeList(logs, Log::writeTo);
-
+    if (withRevertReason && revertReason.isPresent()) {
+      out.writeBytesValue(BytesValue.wrap(revertReason.get().getBytes(StandardCharsets.UTF_8)));
```
rain-on: If someone has asked for the reason, but it doesn't exist, should we include an empty  string?
I.e. Client requested the RevertReason explicitly, probably should give it to them (even if it doesn't exist).

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/TransactionReceipt.java
```diff
@@ -103,6 +124,14 @@ private TransactionReceipt(
    * @param out The RLP output to write to
    */
   public void writeTo(final RLPOutput out) {
+    writeTo(out, false);
+  }
+
+  public void writeToWithReason(final RLPOutput out) {
```
rain-on: does this need to exist? - or do we just expose the following function?
If we keep this, maybe rename to writeToWithRevertReason (i.e. include the word revert).

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/TransactionReceipt.java
rain-on: Could that be done at the level above this? i.e. a TransactionReceipt having or not having a revert reason is ok - however if it comes via a message _then_ it is illegal ... could the message receiver/handler raise the exception if the TR in the message contains a Revert?

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/TransactionReceipt.java
rain-on: This function feels like a mismatch with the constructor - i.e. we can pass in an Optional Revert, but we give back a concrete ... or is this normal? (I can honestly say I don't know).

ethereum/core/src/test-support/java/tech/pegasys/pantheon/ethereum/core/BlockDataGeneratorWithReason.java
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package tech.pegasys.pantheon.ethereum.core;
+
+import java.util.Arrays;
+import java.util.Optional;
+
+public class BlockDataGeneratorWithReason extends BlockDataGenerator {
```
rain-on: Could this just be done with Mocking?

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/core/TransactionReceiptTest.java
```diff
@@ -25,7 +25,22 @@ public void toFromRlp() {
     final BlockDataGenerator gen = new BlockDataGenerator();
     final TransactionReceipt receipt = gen.receipt();
     final TransactionReceipt copy =
-        TransactionReceipt.readFrom(RLP.input(RLP.encode(receipt::writeTo)));
+        TransactionReceipt.readFrom(
+            RLP.input(
+                RLP.encode(
+                    receipt
+                        ::writeToWithReason))); // STEFAN: needs to write the "reason" if available.
```
rain-on: assuming the comment is to be removed?

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/TransactionReceipt.java
pinges: I've changed that so the conversion from Optional<String> to String happens in the json rpc area.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/TransactionReceipt.java
ajsutton: As long as we do reject the message it's probably ok.  Seems cleaner to me to do it here - ultimately just by passing in a boolean to say if the revert reason is allowed.  But I'm not overly fussed.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/MainnetProtocolSpecs.java
pinges: Fixed

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/TransactionReceipt.java
pinges: Done

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/TransactionReceipt.java
pinges: renamed

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/core/TransactionReceiptTest.java
pinges: done

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/condition/eth/EthConditions.java
```diff
@@ -50,4 +52,22 @@ public Condition sendRawTransactionExceptional(
     return new ExpectEthSendRawTransactionException(
         transactions.sendRawTransactionTransaction(transactionData), expectedMessage);
   }
+
+  public Condition expectSuccessfulTransactionReceiptWithReason(
```
CjHare: Why not create a wrapper object for the configuration (`hostName` & `jsonRpcHttpSocketPort`) and use Varargs?

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/PantheonNode.java
```diff
@@ -239,6 +240,15 @@ private String getDiscoveryPort() {
     }
   }
 
+  @Override
+  public Optional<Integer> getJsonRpcHttpPort() {
```
CjHare: How is the JsonRpcHttpPort optional? ...there's always a Json RPC port when JSON-RPC is enabled. Otherwise `null` seems like a more appropriate return value _i.e. dumb programmer problem_

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/PantheonNode.java
```diff
@@ -291,6 +300,10 @@ private NodeRequests nodeRequests() {
     return nodeRequests;
   }
 
+  public Web3jService web3jService() {
```
CjHare: Super not keen on breaking this encapsulation, is there really no other way to achieve this?

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/transaction/eth/EthGetTransactionReceiptRawResponseTransaction.java
```diff
@@ -0,0 +1,44 @@
...
+
+import tech.pegasys.pantheon.tests.acceptance.dsl.transaction.NodeRequests;
+import tech.pegasys.pantheon.tests.acceptance.dsl.transaction.Transaction;
+import tech.pegasys.pantheon.tests.acceptance.dsl.transaction.eth.EthRawRequestFactory.EthGetTransactionReceiptRawResponse;
+import tech.pegasys.pantheon.tests.acceptance.dsl.transaction.eth.EthRawRequestFactory.TransactionReceiptRaw;
+
+import java.io.IOException;
+import java.util.Optional;
+
+public class EthGetTransactionReceiptRawResponseTransaction
```
CjHare: This is inconsistent with the other Transactions, drop the `Response` and duplicated `Transaction`, making the class named `EthGetTransactionReceiptRaw`

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/web3j/RevertReason.sol
```diff
@@ -0,0 +1,39 @@
...
+// compile with:
+// solc RevertReason.sol --bin --abi --optimize --overwrite -o .
+// then create web3j wrappers with:
+// web3j solidity generate -b ./generated/RevertReason.bin -a ./generated/RevertReason.abi -o ../../../../../ -p tech.pegasys.pantheon.tests.web3j.generated
+contract RevertReason {
+
+    uint public value = 0;
+
+    function revertWithRevertReason() public {
+        if (true) {
```
CjHare: Why not use the `assert` and `require` operations for these two methods instead of the redundant if/else?

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/web3j/RevertReasonAcceptanceTest.java
```diff
@@ -0,0 +1,68 @@
...
+  private PantheonNode minerNode;
+
+  @Before
+  public void setUp() throws Exception {
+    minerNode = pantheon.createMinerNode("miner-node-withRevertReason");
+    cluster.start(minerNode);
+  }
+
+  @Test
+  public void mustRevertWithRevertReason() throws Exception {
```
CjHare: @Test don't usually `throws`, why do these need to?

jframe: done. it no longer needs to throw

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/transaction/eth/EthGetTransactionReceiptRawResponseTransaction.java
jframe: done

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/PantheonNode.java
jframe: done. no longer needed

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/PantheonNode.java
jframe: done. no longer done this way

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/condition/eth/EthConditions.java
jframe: done. no longer needed

ethereum/core/src/test-support/java/tech/pegasys/pantheon/ethereum/core/BlockDataGeneratorWithReason.java
jframe: Could be but theres a BlockDataGenerator that already exists. And the code is nicer using the data generator rather than adding mock. I've also added a receipt(revertReason) method to the existing BlockDataGenerator remove need for this extra BlockDataGenerator class.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/web3j/RevertReason.sol
jframe: simplified the test contract to just revert

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/TransactionReceipt.java
jframe: Don't think an empty string makes sense as an empty string is a valid revert reason. And returning a null for the revert reason as it does now makes more sense as there is no revert reason value to return.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/condition/eth/ExpectSuccessfulEthGetTransactionReceiptWithReason.java
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright 2018 ConsenSys AG.
```
CjHare: Shouldn't this be `2019`?

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/condition/eth/ExpectSuccessfulEthGetTransactionReceiptWithReason.java
```diff
@@ -0,0 +1,49 @@
...
+    this.expectedRevertReason = expectedRevertReason;
+  }
+
+  @Override
+  public void verify(final Node node) {
+    WaitUtils.waitFor(
+        () -> assertThat(revertReasonMatches(node, expectedRevertReason)).isPresent());
+  }
+
+  private Optional<TransactionReceiptRaw> revertReasonMatches(
```
CjHare: nit; Having this return a `boolean` would provide a cleaner signature and nicer reading code in the one place it is used

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/condition/eth/ExpectSuccessfulEthGetTransactionReceiptWithoutReason.java
```diff
@@ -0,0 +1,44 @@
+/*
+ * Copyright 2018 ConsenSys AG.
```
CjHare: Looks like the year is off in all of the headers?

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/condition/eth/ExpectSuccessfulEthGetTransactionReceiptWithoutReason.java
```diff
@@ -0,0 +1,44 @@
...
+      final EthGetTransactionReceiptRaw transaction) {
+    this.transaction = transaction;
+  }
+
+  @Override
+  public void verify(final Node node) {
+    WaitUtils.waitFor(() -> assertThat(revertReasonIsEmpty(node)).isPresent());
+  }
+
+  private Optional<String> revertReasonIsEmpty(final Node node) {
```
CjHare: nit; return a `boolean` instead of an Optional

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/transaction/CallSmartContractFunctionWithRevert.java
```diff
@@ -0,0 +1,56 @@
...
+import java.math.BigInteger;
+import java.util.Collections;
+
+import org.web3j.abi.FunctionEncoder;
+import org.web3j.abi.datatypes.Function;
+import org.web3j.crypto.Credentials;
+import org.web3j.protocol.core.methods.response.EthSendTransaction;
+import org.web3j.tx.RawTransactionManager;
+
+public class CallSmartContractFunctionWithRevert implements Transaction<EthSendTransaction> {
```
CjHare: What in this class links to `WithRevert`? ...as it looks like a generic smart contract function invocation

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/transaction/eth/EthRawRequestFactory.java
```diff
@@ -0,0 +1,54 @@
...
+package tech.pegasys.pantheon.tests.acceptance.dsl.transaction.eth;
+
+import java.util.Collections;
+
+import org.web3j.protocol.Web3jService;
+import org.web3j.protocol.core.Request;
+import org.web3j.protocol.core.Response;
+import org.web3j.protocol.core.methods.response.TransactionReceipt;
+
+public class EthRawRequestFactory {
```
CjHare: This class is bunk. 

The one method in the class is calling `eth_getTransactionReceipt`, we have simply _extended_ that function to include an additional field. Why not put this new method in `Eth`?

pantheon/src/test/resources/everything_config.toml
```diff
@@ -97,4 +97,7 @@ privacy-precompiled-address=9
 tx-pool-retention-hours=999
 
 tx-pool-max-size=1234
-Xincoming-tx-messages-keep-alive-seconds=60
\ No newline at end of file
+Xincoming-tx-messages-keep-alive-seconds=60
+
+# Revert Reaspm
```
CjHare: Typo? `#Revert Reason`

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/transaction/CallSmartContractFunctionWithRevert.java
jframe: renamed to CallSmartContractFunction

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/transaction/eth/EthRawRequestFactory.java
jframe: Moved this rpc request method into the custom rpc factory

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/TransactionReceipt.java
```diff
@@ -42,25 +46,37 @@
...
   private final TransactionReceiptType transactionReceiptType;
+  private final Optional<String> revertReason;
 
   /**
    * Creates an instance of a state root-encoded transaction receipt.
    *
    * @param stateRoot the state root for the world state after the transaction has been processed
    * @param cumulativeGasUsed the total amount of gas consumed in the block after this transaction
    * @param logs the logs generated within the transaction
+   * @param revertReason whether storing the revert reason is for failed transactions
```
CjHare: Grammer.

`@param revertReason whether to store the revert reason for failed transactions`

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/TransactionReceipt.java
```diff
@@ -69,32 +85,40 @@ private TransactionReceipt(
    * @param status the status code for the transaction (1 for success and 0 for failure)
    * @param cumulativeGasUsed the total amount of gas consumed in the block after this transaction
    * @param logs the logs generated within the transaction
+   * @param revertReason the revert reason
```
CjHare: @param revertReason whether to store the revert reason for failed transactions

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/TransactionReceipt.java
CjHare: 👍 `null` return value for a `null` revert reason.

In this context `null` and an empty string would have different meanings i.e. no revert message vs an empty revert message.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/TransactionProcessor.java
```diff
@@ -95,6 +97,13 @@ default boolean isSuccessful() {
      * @return the validation result, with the reason for failure (if applicable.)
      */
     ValidationResult<TransactionInvalidReason> getValidationResult();
+
+    /**
+     * Returns the reason a transaction was reverted (if applicable).
```
CjHare: Missing word.

`Returns the reason why a transaction was reverted (if applicable).`

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/TransactionReceipt.java
jframe: This is actually the revert reason not a flag whether or not to store the revert reason. So have updated this and the below constructor to reflect this.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/TransactionReceipt.java
jframe: As above, the java doc was misleading have updated to the same as the java doc for the other constructor

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/TransactionProcessor.java
jframe: done

PR #1604, opened 2019-06-25
[PIE-1707] Implement a timeout in TransactionMessageProcessor

abdelhamidbakhta:
## PR description
In order to avoid the transaction pool to grow unbounded, a maximum capacity has been implemented in #1586. This follow up PR enable to skip processing a transaction if too old. 
- `processTransactionsMessage` now takes a `keepAlive` parameter
- don't process the message if expired
- add unit tests
- use a default timeout for transactions (1 minute)


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/transactions/TransactionsMessageProcessor.java
```diff
@@ -39,6 +42,17 @@ public TransactionsMessageProcessor(
   }
 
   void processTransactionsMessage(
+      final EthPeer peer,
+      final TransactionsMessage transactionsMessage,
+      final Instant startedAt,
+      final Duration keepAlive) {
+    // Check if message not expired.
+    if (startedAt.plus(keepAlive).isAfter(now())) {
```
NicolasMassart: Don't you provide any log or some feedback in case you skip the message?

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/transactions/TransactionsMessageProcessorTest.java
```diff
@@ -50,9 +57,31 @@ public void shouldMarkAllReceivedTransactionsAsSeen() {
...
+        peer1,
+        TransactionsMessage.create(asList(transaction1, transaction2, transaction3)),
+        now(),
+        ofMinutes(1));
     verify(transactionPool)
         .addRemoteTransactions(ImmutableSet.of(transaction1, transaction2, transaction3));
   }
+
+  @Test
+  public void shouldNotMarkAllReceivedTransactionsAsSeenIfExpired() {
```
NicolasMassart:
```suggestion
  public void shouldNotMarkReceivedExpiredTransactionsAsSeen() {
```

ethereum/eth/src/test/java/tech/pegasys/pantheon/ethereum/eth/transactions/TransactionsMessageProcessorTest.java
```diff
@@ -50,9 +57,31 @@ public void shouldMarkAllReceivedTransactionsAsSeen() {
...
+    messageHandler.processTransactionsMessage(
+        peer1,
+        TransactionsMessage.create(asList(transaction1, transaction2, transaction3)),
+        now().minus(ofMinutes(1)),
+        ofMillis(1));
+    verifyZeroInteractions(transactionTracker);
+  }
+
+  @Test
+  public void shouldNotAddReceivedTransactionsToTransactionPoolIfExpired() {
```
NicolasMassart:
```suggestion
  public void shouldNotAddReceivedExpiredTransactionsToTransactionPool() {
```

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/transactions/TransactionsMessageProcessor.java
abdelhamidbakhta: No because potentially we will drop thousands of messages and logging would be a bottleneck. A potential follow up PR would gather information and maintain a counter as a prometheus metric to see the total number of dropped messages.

abdelhamidbakhta: I will create a follow up PR after this one to keep it small.

abdelhamidbakhta: As discussed in #1586 we have to deal with up to 500 tx / second. We want to drop oldest messages because they are more likely to contain non relevant transactions. So we check if the message has expired prior to process it. Since we would probably have to skip thousands of messages we don't want to log each individual message skipped. A follow up PR will be created after this one to ensure we have proper feedback.

PR #1607, opened 2019-06-26
Added readiness and liveness endpoints

MadelineMurray:
Added readiness and liveness endpoints


docs/Pantheon-API/Using-JSON-RPC-API.md
```diff
@@ -71,7 +71,40 @@ Send individual requests as a JSON data package at each prompt:
...
 
-### API Methods Enabled by Default
+## Readiness and Liveness Endpoints 
+
+Pantheon provides readiness and liveness endpoints to confirm the Pantheon node status. Both return a
+`200 OK` status when ready or live and a `503 Service Unavailable` status if not ready or live. 
+ 
+By default, the readiness check requires a peer to be connected and the node to be within 2 blocks of the best
+known block. If [p2p communication is disabled](../Reference/Pantheon-CLI-Syntax.md#p2p-enabled), 
+no peers are required.
```
ajsutton: If there are no remote peers then by definition we have the best chain.  So effectively if p2p communication is disabled, the node is always considered ready.

docs/Pantheon-API/Using-JSON-RPC-API.md
```diff
@@ -71,7 +71,40 @@ Send individual requests as a JSON data package at each prompt:
...
+`200 OK` status when ready or live and a `503 Service Unavailable` status if not ready or live. 
+ 
+By default, the readiness check requires a peer to be connected and the node to be within 2 blocks of the best
+known block. If [p2p communication is disabled](../Reference/Pantheon-CLI-Syntax.md#p2p-enabled), 
+no peers are required. 
+
+Use the query parameters `minPeers` and `maxBlocksBehind` to adjust the number of peers required and number of blocks tolerance.
+
+```bash tab="Readiness Endpoint"
+http://<JSON-RPC-HTTP-endpoint:port>/readiness
```
ajsutton: Missing the actual `curl` command.  And consider including `-v` arg so the headers are output and you can see the status code.

docs/Pantheon-API/Using-JSON-RPC-API.md
MadelineMurray: Fixed

docs/Pantheon-API/Using-JSON-RPC-API.md
MadelineMurray: It's the endpoint so it's missing on purpose - could also use the endpoint in Postman or similar rather than curl.  Update the example name to make it clearer

PR #1608, opened 2019-06-26
Upload jars to bintray as part of releases

ajsutton:
## PR description
Testing upload of jars to bintray (targeting a test repo for now).


build.gradle
```diff
@@ -249,6 +277,54 @@ task deploy() {}
...
 subprojects {
 
+  if (file('src/main/java').directory) {
+    apply plugin: 'com.jfrog.bintray'
+    apply plugin: 'maven-publish'
+
+    publishing {
+      publications {
+        mavenJava(MavenPublication) {
+          groupId "tech.pegasys.${project.group}"
```
shemnon: (a) do we really want each sob jar to be it's own maven group?  Since they are all needed in pantheon and only used in pantheon and can't be severed from pantheon (except for two cases: metrics and plugins api) wouldn't one group make more sense? Could we all put them in to one group and let the artifact name do the work?

(b) regardless of (a) should the group name include something like "internal" or "unstable" to make it clear it's not a stable API?

build.gradle
```diff
@@ -249,6 +277,54 @@ task deploy() {}
...
+              }
+            }
+            scm {
+              connection = 'scm:git:git://github.com/PegaSysEng/pantheon.git'
+              developerConnection = 'scm:git:ssh://github.com/PegaSysEng/pantheon.git'
+              url = 'https://github.com/PegaSysEng/pantheon'
+            }
+          }
+        }
+      }
```
shemnon: I think if you add a `repository` block you may get the maven metadata

```gradle
  repositories {
    maven {
      // change URLs to point to your repos, e.g. http://my.org/repo
      def releasesRepoUrl = "$buildDir/repos/releases"
      def snapshotsRepoUrl = "$buildDir/repos/snapshots"
      url = version.endsWith('SNAPSHOT') ? snapshotsRepoUrl : releasesRepoUrl
    }
  }
```

build.gradle
ajsutton: Everything in `tech.pegasys.pantheon` makes sense to me.

Including 'internal' probably makes sense.  We may need to find a way to override some of this info on a per-package basis (eg would be nice if metrics wasn't marked internal) but starting with everything internal and then some migrating out if/when we work out how to do that makes sense to me.

ajsutton:
Unfortunately there's no way to know for sure the release will still work until we actually try it next.  I have disabled the automatic publishing on every master build because that will conflict (both the release job and master job would be trying to publish the same release).  Will discuss the various options to avoid that and go from there.

build.gradle
ajsutton: No luck.  But I've tested with both gradle and maven depending on these publish jars and it works even without the metadata.

PR #1609, opened 2019-06-26
[PIE-1709] Implement Prometheus metric counter for skipped expired transaction messages.

abdelhamidbakhta:
## PR description
Follow up PR of #1604 
We want to provide Prometheus metrics when the [TransactionsMessageProcessor](https://github.com/PegaSysEng/pantheon/blob/master/ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/transactions/TransactionsMessageProcessor.java) skips some expired messages.
- a prometheus counter is created for skipped messages
- the counter is updated when a message is skipped.
- a log is provided every time a step is hit (default is 1000).
## Fixed Issue(s)

#PIE-1709


NicolasMassart:
Why counting in Pantheon and not just sending an event on the metrics API (received usually by Prometheus in our case) and let Prometheus count and build the aggregated data? It would remove the need to maintain a counter in Pantheon.

abdelhamidbakhta:
> Why counting in Pantheon and not just sending an event on the metrics API (received usually by Prometheus in our case) and let Prometheus count and build the aggregated data? It would remove the need to maintain a counter in Pantheon.

The interface actually use Prometheus API.

PR #1610, opened 2019-06-26
[PIE-1710] Expose a CLI option to configure the life time of transaction messages.

abdelhamidbakhta:
## PR description
We want to make configurable the life time of transactions. This life time is used to drop incoming transactions from the network if they are too old. 
- TX_KEEP_ALIVE is not anymore a constant and is set in the constructor of TransactionsMessageHandler
- the TransactionPoolFactory pass the option to the TransactionsMessageHandler
- add a cli option `--tx-pool-keep-alive-seconds`

## Fixed Issue(s)
#PIE-1710


pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -817,6 +825,7 @@ private void ensureAllNodesAreInWhitelist(
               new MiningParameters(coinbase, minTransactionGasPrice, extraData, isMiningEnabled))
           .maxPendingTransactions(txPoolMaxSize)
           .pendingTransactionRetentionPeriod(pendingTxRetentionPeriod)
+          .txMessageKeepAliveSeconds(txMessageKeepAliveSeconds)
```
mbaxter: (Optional) Might be time for a `TransactionPoolConfiguration` object.

abdelhamidbakhta: I will do this in a follow up pr.

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -566,6 +567,13 @@ void setBannedNodeIds(final List<String> values) {
       arity = "1")
   private final Integer pendingTxRetentionPeriod = PendingTransactions.DEFAULT_TX_RETENTION_HOURS;
 
+  @Option(
+      names = {"--tx-pool-keep-alive-seconds"},
+      paramLabel = MANDATORY_INTEGER_FORMAT_HELP,
+      description = "Keep alive of transactions in seconds (default: ${DEFAULT-VALUE})",
+      arity = "1")
+  private final Integer txMessageKeepAliveSeconds = TransactionPool.DEFAULT_TX_MSG_KEEP_ALIVE;
```
ajsutton: I don't think this should be a visible option - it's potentially very misleading for users and ideally wouldn't ever kick in since we should keep up with transaction messages.  I'd say it should just be a `-X` option.

abdelhamidbakhta: #1622

PR #1611, opened 2019-06-26
[PAN-2825] eea send raw transaction with privacy group id

iikirilov:
# PR description

Allows to send private transaction with privacy group id instead of privateFor/privateFrom

## Fixed Issue(s)
[PAN-2825]

[PAN-2825]: https://pegasys1.atlassian.net/browse/PAN-2825


ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/privacy/PrivateTransactionHandlerTest.java
```diff
@@ -107,26 +107,28 @@ public void setUp() throws Exception {
             mockEnclave(),
             Address.DEFAULT_PRIVACY,
             KEY_PAIR,
+            "A1aVtMxLCUHmBVHXoZzzBgPbW/wj5axDpW9X8l91SGo=",
```
ekellstrand: Can this get loaded from the key file so it's clear what this value is?

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/privacy/PrivateTransactionTest.java
```diff
@@ -50,6 +50,14 @@
           + "6e766966746a69697a706a52742b4854754642733d8a726573747269637465"
           + "64";
 
+  private static final String VALID_PRIVATE_TRANSACTION_RLP_PRIVACY_GROUP =
```
ekellstrand: Can this be generated so that it is clear how this is built?

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/privacy/EeaDeletePrivacyGroup.java
```diff
@@ -0,0 +1,66 @@
...
+
+    LOG.trace("Deleting a privacy group with privacyGroupId {} and from {}", privacyGroupId, from);
+
+    DeletePrivacyGroupRequest deletePrivacyGroupRequest =
+        new DeletePrivacyGroupRequest(privacyGroupId, from);
+    String response;
+    try {
+      response = enclave.deletePrivacyGroup(deletePrivacyGroupRequest);
+    } catch (Exception e) {
+      LOG.error("Failed to fetch transaction from Enclave with error " + e.getMessage());
```
ekellstrand: Any reason these 2 log lines aren't combined?

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/privacy/PrivateTransactionTest.java
iikirilov: No - because I am testing here that the generation will produce these fixed values. If I generate these values and test if the generator produces the generated values (which it obviously will) I am not really testing anything. If someone messes with the generator then we will not catch the error :)

PR #1613, opened 2019-06-27
PIE-1717: Add TransactionValidationParam to TxProcessor

lucassaldanha:
## PR description
- Updated TransactionValidationParams with checkLocalPermissioning flag
- Added TransactionValidationParam to TxProcessor


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/TransactionValidationParams.java
```diff
@@ -31,10 +35,55 @@ public boolean checkOnchainPermissions() {
...
-  public static class Builder {
+  public boolean checkLocalPermissions() {
+    return checkLocalPermissions;
+  }
+
+  public static TransactionValidationParams transactionSimulator() {
+    return new Builder()
+        .checkLocalPermissions(false)
+        .checkOnchainPermissions(false)
+        .allowFutureNonce(false)
```
Errorific: Should future nonce be true for the simulator?

PR #1614, opened 2019-06-27
[Hackathon] Blake2b f compression precompile 

iikirilov:
# PR description

Introduces the Blake2b F compression function as a precompile in Pantheon as specified in [https://github.com/ethereum/EIPs/pull/2129](https://github.com/ethereum/EIPs/pull/2129)

Using the test vectors from [https://github.com/keep-network/go-ethereum/blob/f-precompile/core/vm/contracts_test.go#L350](https://github.com/keep-network/go-ethereum/blob/f-precompile/core/vm/contracts_test.go#L350)

Implementation of Blake2b F compression function ported from https://github.com/keep-network/blake2 (BSD license)

## Fixed Issue(s)

[https://github.com/PegaSysEng/BountiedWork/issues/19](https://github.com/PegaSysEng/BountiedWork/issues/19)


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/precompiles/BLAKE2fPrecompileContract.java
```diff
@@ -0,0 +1,278 @@
...
+import tech.pegasys.pantheon.ethereum.vm.MessageFrame;
+import tech.pegasys.pantheon.util.bytes.BytesValue;
+
+import java.util.Arrays;
+
+import com.google.common.primitives.Ints;
+import com.google.common.primitives.Longs;
+import org.bouncycastle.util.Pack;
+
+public class BLAKE2fPrecompileContract extends AbstractPrecompiledContract {
```
shemnon: This name implies Compression Function F is on the same level as Blake2B and Blake2s.  `BLAKE2bCompressionFunctionFPrecompileContract` is a mouthful, but more accurate than `Blake2f`.  So I'de like to see it changed.  If solidity has a nice name for it we can use that one as `<foo>PrecompileContract`, so keep an eye out.

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/mainnet/precompiles/BLAKE2fPrecompileContractTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+  public BLAKE2fPrecompileContractTest() {}
+
+  private MessageFrame messageFrame = mock(MessageFrame.class);
+
+  @Parameterized.Parameters
+  public static Object[][] parameters() {
+    return new Object[][] {
+      {
+        "6a09e667f2bd8948bb67ae8484caa73b3c6ef372fe94f82ba54ff53a5f1d36f1510e527fade682d19b05688c2b3e6c1f1f83d9abfb41bd6b5be0cd19137e2179000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f202122232425262728292a2b2c2d2e2f303132333435363738393a3b3c3d3e3f0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000c",
```
shemnon: Where do these test cases come from?  Add a link please.  If you create them yourself that is fine, we should fine someone else's test vectors and add them as well, with a link.

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/mainnet/precompiles/BLAKE2fPrecompileContractTest.java
```diff
@@ -0,0 +1,93 @@
...
+import tech.pegasys.pantheon.ethereum.vm.MessageFrame;
+import tech.pegasys.pantheon.util.bytes.BytesValue;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+@RunWith(Parameterized.class)
+public class BLAKE2fPrecompileContractTest {
+  private final BLAKE2fPrecompileContract contract =
```
shemnon: Can we get some negative test cases as well? size checking, input/output that won't match, etc.  Probably in a separate test file since this one is parameterized.

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/mainnet/precompiles/BLAKE2fPrecompileContractTest.java
iikirilov: The test vectors are linked in the PR description above. They come from the go implementation :) Do you mean a comment with the link in the code?

shemnon: Yes, a comment providing a link to the code.  This will be squash merged so patch history doesn't move as well.

shemnon: And for the code impl itself, a link in the javadoc would be good.

iikirilov:
This PR has passed the deadline for the hackathon period but it is still incomplete.

1. The precompile was not integrated with the precompile registry.
2. The EIP has not been finalized so some implementation details/test vectors may change.
3. The gas calculation was not tested.

I have the capacity to follow up on this should that be requested by the reviewer.

shemnon:
The gas cost appears to be specified in ethereum/EIPs#2129. While this is still open the got some positive feedback on the All Core Devs call. Can you finish the gas cost calculations and function registration based on these values?

iikirilov:
@shemnon you will find that the gas requirement method already reflects this. Should this be abstracted to the `FrontierGasCalculator`?

shemnon:
No need to bring it into frontier.  You listed some other TODOs - 
1. The precompile was not integrated with the precompile registry.
2. The EIP has not been finalized so some implementation details/test vectors may change.
3. The gas calculation was not tested.
4. The name of the precompile was not agreed.

For 1 this task is outstanding.  
For 2 I am of the opinion the changes won't be dramatic, and likely there will be no more changes.  
For 3 can you add some simple gas calculation tests based on the existing test vectors
For 4 lets call it "BLAKE2BF", if a different name comes up it's a trivial change we can handle later.

iikirilov:
I have addressed the above. Waiting for the tests cases to be updated and we should be good to go.

PR #1615, opened 2019-06-27
[PAN-2850] Create a transaction pool configuration object

abdelhamidbakhta:
## PR description
Regroup `txPoolMaxSize`, `pendingTxRetentionPeriod`, `txMessageKeepAliveSeconds` CLI options into one `TransactionPoolConfiguration` object.


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/transactions/TransactionPoolFactory.java
```diff
@@ -30,16 +30,17 @@ public static TransactionPool createTransactionPool(
       final ProtocolContext<?> protocolContext,
       final EthContext ethContext,
       final Clock clock,
-      final int maxPendingTransactions,
       final MetricsSystem metricsSystem,
       final SyncState syncState,
-      final int maxTransactionRetentionHours,
       final Wei minTransactionGasPrice,
```
mbaxter: (Discussion) What about folding `minTransactionGasPrice` into `TransactionPoolConfiguration`?  Or maybe the `TransactionPool` also gets access to `MiningParams` so that we don't end up duplicating data on 2 different config objects?

abdelhamidbakhta: Yes it makes sense.

abdelhamidbakhta: I mean the second approach where TransactionPool also gets access to MiningParams

abdelhamidbakhta: Actually the `PantheonControllerBuilder` passes the `minTransactionPrice` to the transactionPool from the `MiningParams`
```java
    final TransactionPool transactionPool =
        TransactionPoolFactory.createTransactionPool(
            protocolSchedule,
            protocolContext,
            ethProtocolManager.ethContext(),
            clock,
            metricsSystem,
            syncState,
            miningParameters.getMinTransactionGasPrice(),
            transactionPoolConfiguration)
```
SInce it uses only one parameter from the MiningParams i don't think we need to give it access to the whole object now. What do you think ?

mbaxter: 🤷‍♀ Just feels odd to have a config object and then have an additional dangling option.  But there's some weirdness with all 3 options really (as is + my 2 alternative suggestions).  I think I'm leaning towards passing in the `MiningParams` - it makes it more explicit that the transactionPool's behavior is modified by the mining configuration.  But I'll leave it up to you.

PR #1622, opened 2019-07-01
[PIE-1736] Implement incoming transaction messages CLI option as an unstable command.

abdelhamidbakhta:
## PR description
`--tx-pool-keep-alive-seconds` should be a `-X` option.
- `tx-pool-keep-alive-seconds` is renamed to `incoming-tx-messages-keep-alive-seconds`
- `incoming-tx-messages-keep-alive-seconds` is an unstable command not visible for the user (`--Xincoming-tx-messages-keep-alive-seconds`)

Related to #1610


pantheon/src/test/java/tech/pegasys/pantheon/cli/PantheonCommandTest.java
```diff
@@ -2612,38 +2608,31 @@ public void errorIsRaisedIfStaticNodesAreNotWhitelisted() throws IOException {
...
-    parseCommand("--tx-pool-keep-alive-seconds", String.valueOf(txMessageKeepAliveSeconds));
-
-    verify(mockControllerBuilder)
-        .transactionPoolConfiguration(transactionPoolConfigurationArgumentCaptor.capture());
-    assertThat(transactionPoolConfigurationArgumentCaptor.getValue().getTxMessageKeepAliveSeconds())
-        .isEqualTo(txMessageKeepAliveSeconds);
-
+    parseCommand(
+        "--Xincoming-tx-messages-keep-alive-seconds", String.valueOf(txMessageKeepAliveSeconds));
+    verify(mockTransactionPoolConfigurationBuilder).build();
```
mbaxter: You're not actually checking that the value 999 is set.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/transactions/TransactionPoolConfiguration.java
```diff
@@ -79,7 +81,15 @@ public static Builder builder() {
...
-    private int txMessageKeepAliveSeconds = TransactionPool.DEFAULT_TX_MSG_KEEP_ALIVE;
+
+    @CommandLine.Option(
+        names = {"--Xincoming-tx-messages-keep-alive-seconds"},
+        paramLabel = "<INTEGER>",
+        hidden = true,
+        description =
+            "Keep alive of incoming transaction messages in seconds (default: ${DEFAULT-VALUE})",
+        arity = "1")
+    private Integer txMessageKeepAliveSeconds = TransactionPool.DEFAULT_TX_MSG_KEEP_ALIVE;
```
mbaxter: Can this be a primitive `int` instead of an `Integer`?

abdelhamidbakhta: It does not work with an int. 
See: https://github.com/PegaSysEng/pantheon/blob/285f9f8589725355bc728b45402a7d5ee4562bde/pantheon/src/test/java/tech/pegasys/pantheon/cli/PantheonCommandTest.java#L675
With an int the else is used and it tries to get a string and it fails.

pantheon/src/test/java/tech/pegasys/pantheon/cli/PantheonCommandTest.java
abdelhamidbakhta: Yes. We have not yet the mechanism to do this kind of assertion with unstable options. I suggest we think about it in another PR because it is out of scope of this PR.

abdelhamidbakhta: We found a way to test it using a real object instead of the mock

PR #1627, opened 2019-07-02
[PIE-1741] Refactor CLI

abdelhamidbakhta:
## PR description
- refactor packages
- remove `PantheonExceptionHandler` inner class and create a separate class
- introduce subcommand package


pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -857,7 +833,7 @@ private PantheonCommand controller() {
     return this;
   }
 
-  PantheonController<?> buildController() {
+  public PantheonController<?> buildController() {
```
mbaxter: You've got a lot of visibility changes here - looks like at least a few of these are unnecessary.  Can you double-check the visibility changes?

abdelhamidbakhta: I have double checked. They are all necessary because sub commands are not in  the same package as the parent command. I have only made public methods that are actually used by sub commands.

PR #1631, opened 2019-07-03
PAN-2860 - Ignore discport during static node permissioning check

usmansaleem:
## PR description
Convert ENodeURI to URI by dropping query strings to allow comparison for permitted nodes in O(1) instead of O(n) in UpdateSyncStatusNodePermissioningProvider

## Fixed Issue(s)

fixes #PAN-2860


ethereum/permissioning/src/main/java/tech/pegasys/pantheon/ethereum/permissioning/node/provider/SyncStatusNodePermissioningProvider.java
```diff
@@ -116,4 +123,19 @@ public boolean isPermitted(final EnodeURL sourceEnode, final EnodeURL destinatio
   public boolean hasReachedSync() {
     return hasReachedSync;
   }
+
+  private static URI removeQueryFromURI(final URI uri) {
```
lucassaldanha: What about creating a method in the EnodeURL class called `toURIWithoutQueryParam` or something similar?
This will enable us to Convert from the EnodeURL straight into the URI without the query param and also won't require us building the URI with the query (inside the toURI) method just to strip it later.

usmansaleem: :+1:

PR #1638, opened 2019-07-03
[PAN-2819] [PAN-2820] Mark Sweep

RatanRSur:
## PR description
Brings in Mark Sweep Pruner and test associated with it.

Key differences from PoC:
* PrunerTest.java
* Waiting for a constant number of blocks before starting to mark to prevent wasted effort caused by reorgs

## Fixed Issue(s)
https://pegasys1.atlassian.net/secure/RapidBoard.jspa?rapidView=28&modal=detail&selectedIssue=PAN-2820
https://pegasys1.atlassian.net/secure/RapidBoard.jspa?rapidView=28&modal=detail&selectedIssue=PAN-2819


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/worldstate/Pruner.java
```diff
@@ -0,0 +1,129 @@
...
+                .setDaemon(true)
+                .setPriority(Thread.MIN_PRIORITY)
+                .setNameFormat("StatePruning-%d")
+                .build());
+    this.retentionPeriodInBlocks = retentionPeriodInBlocks;
+    this.transientForkThreshold = transientForkThreshold;
+  }
+
+  public void start() {
+    // TODO: Don't attempt pruning while a fast sync is in progress.
```
RatanRSur: @ajsutton Is this still an issue given that `start` is called only when it's a full-sync?

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/worldstate/MarkSweepPruner.java
```diff
@@ -0,0 +1,185 @@
...
+    worldStateStorage.removeNodeAddedListener(nodeAddedListenerId); // Just in case.
+    nodeAddedListenerId = worldStateStorage.addNodeAddedListener(this::markNewNodes);
+  }
+
+  // Note chainHeadStateRoot must be the state root of the current chain head.
+  // We can delay the actual sweep until a certain number of blocks in the future if we want to
+  // have a certain number of block history available.
+  // TODO: Need to ensure we only start marking when new world states aren't being persisted
+  // Once we add our node added listener persisting world states can continue but we can't start
+  // half way through persisting a world state.
```
RatanRSur: @ajsutton Is some of this outdated? As far as my understanding goes, the hash given to mark needn't be the chain head, is that right?

Any new thoughts on the persistence?

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/worldstate/Pruner.java
ajsutton: No, not an issue anymore.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/worldstate/MarkSweepPruner.java
ajsutton: The block we mark needs to be the current chain head at the time that we add the NodeAddedListener.  We can then wait a few blocks before we actually start marking to ensure we aren't on a short lived fork - it's adding the NodeAddedListener while the mark block is head that's critical.

Otherwise if head was 10 and we picked 5 as the mark block and added the NodeAddedListener, blocks 11 onwards would be marked by the NodeAddedListener and 5 would be marked by walking its trie but 6,7,8,9 would be missed.

ethereum/blockcreation/src/main/java/tech/pegasys/pantheon/ethereum/blockcreation/BlockMiner.java
```diff
@@ -80,6 +80,11 @@ public void run() {
         Thread.currentThread().interrupt();
       } catch (final Exception ex) {
         LOG.error("Block mining threw an unhandled exception.", ex);
+        try {
+          Thread.sleep(TimeUnit.SECONDS.toMillis(5));
+        } catch (final InterruptedException e) {
+          Thread.currentThread().interrupt();
```
ajsutton: Apologies if I somehow introduced this into the POC, but this seems unrelated to the mark/sweep process.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/worldstate/Pruner.java
```diff
@@ -0,0 +1,128 @@
...
+
+  private void handleNewBlock(final BlockAddedEvent event) {
+    if (!event.isNewCanonicalHead()) {
+      return;
+    }
+    final BlockHeader header = event.getBlock().getHeader();
+    if (state.compareAndSet(State.IDLE, State.TRANSIENT_FORK_OUTLIVING)) {
+      pruningStrategy
+          .prepare(); // implications for calling prepare here when blocks can come from multiple
+      // threads?
```
ajsutton: There can only be one thread at a time importing a block, and new block notifications are sent as part of that process, so only ever one thread at a time in this method but which thread it is may change (so we need to ensure changes are visible cross-thread).  So we'll need to make `nodeAddedListenerId` volatile in `MarkSweepPruner` but otherwise it will be safe.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/worldstate/MarkSweepPruner.java
```diff
@@ -0,0 +1,179 @@
...
+  private static final Logger LOG = LogManager.getLogger();
+  private static final BytesValue IN_USE = BytesValue.of(1);
+  private static final int MARKS_PER_TRANSACTION = 1000;
+  private final WorldStateStorage worldStateStorage;
+  private final KeyValueStorage markStorage;
+  private final Counter markedNodesCounter;
+  private final Counter markOperationCounter;
+  private final Counter sweepOperationCounter;
+  private final Counter sweptNodesCounter;
+  private long nodeAddedListenerId;
```
ajsutton: Because calls to `prepare` and `sweep` could come from any thread, we need to make this `volatile` to ensure the latest value is visible to the CPU running the thread those calls happen to come in on.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/worldstate/Pruner.java
```diff
@@ -0,0 +1,128 @@
...
+        });
+  }
+
+  private void execute(final Runnable action) {
+    try {
+      executor.execute(action);
+    } catch (final Throwable t) {
+      LOG.error("Pruning failed", t);
+      state.set(State.IDLE);
+      System.exit(1);
```
ajsutton: We definitely shouldn't be calling `System.exit`.  Failing pruning and going back to `IDLE` is probably enough.

I may have done this in the PoC so it was really obvious if pruning failed and to prevent it moving on and changing all the data which would prevent me investigating what happened further.

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/worldstate/PrunerTest.java
```diff
@@ -0,0 +1,91 @@
...
+import tech.pegasys.pantheon.ethereum.storage.keyvalue.KeyValueStoragePrefixedKeyBlockchainStorage;
+import tech.pegasys.pantheon.metrics.noop.NoOpMetricsSystem;
+import tech.pegasys.pantheon.services.kvstore.InMemoryKeyValueStorage;
+
+import java.util.List;
+
+import org.junit.Test;
+import org.mockito.Mock;
+
+public class PrunerTest {
```
ajsutton: We probably need to cover more cases than this.  At least covering the normal flow to check that it starts marking the right block and then sweeps when that's done. Also around things like when marking completes before the block retention has been reached and visa versa (so we never sweep if marking hasn't finished).

Sorry, won't be fun to test but it is important.

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/DefaultSynchronizer.java
```diff
@@ -164,6 +166,8 @@ private void handleFastSyncResult(final FastSyncState result, final Throwable er
 
   private void startFullSync() {
     fullSyncDownloader.start();
+    // TODO: enable when pruning cli implemented
+    // pruner.start();
```
ajsutton: I would omit this comment for now.  Or possibly just use a hard-coded feature toggle (aka a `private static boolean PRUNING_ENABLED = false`) which we'd later replace with a CLI feature toggle.

ethereum/trie/src/main/java/tech/pegasys/pantheon/ethereum/trie/StoredMerklePatriciaTrie.java
```diff
@@ -40,12 +42,19 @@
...
    * @param valueDeserializer A function for deserializing values from bytes.
+   * @param singleUseNodes whether nodes should unload data after being visited
    */
   public StoredMerklePatriciaTrie(
       final NodeLoader nodeLoader,
       final Function<V, BytesValue> valueSerializer,
-      final Function<BytesValue, V> valueDeserializer) {
-    this(nodeLoader, MerklePatriciaTrie.EMPTY_TRIE_NODE_HASH, valueSerializer, valueDeserializer);
+      final Function<BytesValue, V> valueDeserializer,
+      final boolean singleUseNodes) {
```
ajsutton: As a follow up, it would be worth seeing if we can replace this flag with something in `AllNodesVisitor` - e.g. after visiting each node it calls `Node.release` or something.  If you're iterating all nodes, you never want them to all stay in memory so it would be nice if that automatically took care of itself and then the change would be a lot more focussed as well.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/worldstate/Pruner.java
```diff
@@ -0,0 +1,128 @@
...
+    this.retentionPeriodInBlocks = retentionPeriodInBlocks;
+    this.transientForkThreshold = transientForkThreshold;
+  }
+
+  public void start() {
+    blockchain.observeBlockAdded((event, blockchain) -> handleNewBlock(event));
+  }
+
+  public void stop() {
+    executor.shutdownNow();
```
ajsutton: Need to wait for the executor to actually shut down - this will cause it to discard any task not started and interrupt any running threads, but there's no guarantee when those threads will actually notice the interrupt so need to wait for them to actually exit with `executor.awaitShutdown`.

services/kvstore/src/test/java/tech/pegasys/pantheon/services/kvstore/AbstractKeyValueStorageTest.java
```diff
@@ -62,6 +63,20 @@ public void put() throws Exception {
         Optional.of(BytesValue.fromHexString("0DEF")), store.get(BytesValue.fromHexString("0F")));
   }
 
+  @Test
+  public void mayContainKey() throws Exception {
```
ajsutton: Naming of this test is suspicious. There's no "may" in it anymore, we're getting a very explicit response to whether this key exists or not.

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/worldstate/PrunerTest.java
RatanRSur: Thoughts on new tests?

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/worldstate/PrunerTest.java
```diff
@@ -0,0 +1,165 @@
...
+    options = new BlockOptions().setBlockNumber(2L).setParentHash(forkBlock.getHash());
+    final Block reorgCausingBlock = gen.block(options);
+    receipts = gen.receipts(reorgCausingBlock);
+    blockchain.appendBlock(reorgCausingBlock, receipts);
+    verify(markSweepPruner).mark(canonicalHead.getHeader().getStateRoot());
+    verify(markSweepPruner, never()).sweep();
+  }
+
+  @Test
+  public void shouldThrowIllegalArgumentExceptionCorrectly() {
```
ajsutton: nit: A better name would include something about what's being tested, not just the expected outcome: e.g. `shouldRejectInvalidOptions`

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/worldstate/PrunerTest.java
ajsutton: Yeah I think these look good.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/worldstate/ExecutorServiceBuilder.java
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package tech.pegasys.pantheon.ethereum.worldstate;
+
+import java.util.concurrent.ExecutorService;
+
+public interface ExecutorServiceBuilder {
```
shemnon: A `Supplier<ExecutorService>` would be cleaner.  We don't need to do reflection or reification of the types.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/worldstate/MarkSweepPruner.java
```diff
@@ -0,0 +1,177 @@
...
+    nodeAddedListenerId = worldStateStorage.addNodeAddedListener(this::markNewNodes);
+  }
+
+  public void mark(final Hash rootHash) {
+    markOperationCounter.inc();
+    markStorage.clear();
+    createStateTrie(rootHash)
+        .visitAll(
+            node -> {
+              if (Thread.interrupted()) {
```
shemnon: Is checking the interrupt the best way to stop an in-process mark?  Should we consider a shared boolean instead?  Also, do we have a test case covering this abort scenario?  Is it one we expect often?  Or just during shutdown?

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/worldstate/Pruner.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    final long blockNumber = event.getBlock().getHeader().getNumber();
+    if (state.compareAndSet(State.IDLE, State.TRANSIENT_FORK_OUTLIVING)) {
+      pruningStrategy.prepare();
+      markBlockNumber = blockNumber;
+    }
+
+    if (blockNumber >= markBlockNumber + transientForkThreshold
+        && state.compareAndSet(State.TRANSIENT_FORK_OUTLIVING, State.MARKING)) {
```
shemnon: Style question: would a nested if for the compareAndSet communicate the side effects better than the short circuit &&?

```
if (blockNumber >= markBlockNumber + transientForkThreshold) {
  if (state.compareAndSet(State.TRANSIENT_FORK_OUTLIVING, State.MARKING)) {
  //...
  }
}
```

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/worldstate/Pruner.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+  }
+
+  private void handleNewBlock(final BlockAddedEvent event) {
+    if (!event.isNewCanonicalHead()) {
+      return;
+    }
+
+    final long blockNumber = event.getBlock().getHeader().getNumber();
+    if (state.compareAndSet(State.IDLE, State.TRANSIENT_FORK_OUTLIVING)) {
```
shemnon: In this stretch of if blocks is it ever possible to trigger two of them in the same call?  Would that be a bug?  If only one could fire should we `else if` them?  This could interact with my short-circuit boolean comment below.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/worldstate/Pruner.java
```diff
@@ -0,0 +1,134 @@
...
+          String.format(
+              "Expected transientForkThreshold >= 0 and retentionPeriodInBlocks >= transientForkThreshold. transientForkThreshold=%d, retentionPeriodInBlocks=%d",
+              transientForkThreshold, retentionPeriodInBlocks));
+    }
+    this.retentionPeriodInBlocks = retentionPeriodInBlocks;
+    this.transientForkThreshold = transientForkThreshold;
+  }
+
+  public void start() {
+    executor = executorServiceBuilder.build();
```
shemnon: Do we need a new executor every time?  Can we guarantee start will only be called once per pruner before stop is called?  Or do we need to null out the pruner and throw a "start called twice" error?  Once we call stop will the pruner instance ever be re-used?

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/worldstate/Pruner.java
```diff
@@ -0,0 +1,134 @@
...
+    blockchain.observeBlockAdded((event, blockchain) -> handleNewBlock(event));
+  }
+
+  public void stop() throws InterruptedException {
+    if (executor != null) {
+      executor.awaitTermination(30, TimeUnit.SECONDS);
+    }
```
shemnon: If we are stopping the executor we should clean up the blockchain listener too.

pantheon/src/main/java/tech/pegasys/pantheon/controller/PantheonControllerBuilder.java
```diff
@@ -199,6 +203,31 @@
...
     final MutableBlockchain blockchain = protocolContext.getBlockchain();
 
+    final Pruner pruner =
+        new Pruner(
+            new MarkSweepPruner(
+                protocolContext.getWorldStateArchive().getWorldStateStorage(),
+                storageProvider.createPruningStorage(),
+                metricsSystem),
+            protocolContext.getBlockchain(),
+            () ->
```
shemnon: I don't think we need to wrap this in an interface.  Especially since stop is only called at JVM shutdown.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/worldstate/Pruner.java
RatanRSur: It used to be `else if`'d but I changed them to be chained in order to allow it to cascade from one state to another in, for example, the case where the `TRANSIENT_FORK_OUTLIVING` period is 0.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/worldstate/Pruner.java
RatanRSur: It should be the case because we are only planning, as of now, to have it toggled through the command line

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/worldstate/MarkSweepPruner.java
RatanRSur: Since we're not looking to ever abort a mark ourselves, my understanding is the only thing we need to make sure to check is if we're interrupted. Does that make sense?

shemnon: That makes sense, we are only looking for JVM shutdown.  May be worth a code comment in that case.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/worldstate/Pruner.java
shemnon: That makes me wonder if the logic is too fragile.  Could we switch or if/else if on the current state and then check our other conditions and then set?

RatanRSur: To really make sure the logic is tight (we're exiting asap, not checking states we know we can't be in), we can use sometimes breakless cases in a switch to cascade through. I find that it has quite difficult to grasp the control flow though. The current version is doing unnecessary checks but we know that once we get one failure to state transition, the rest will fail too and it's much more readable.

```
    switch (state.get()) {
      case IDLE:
        {
          state.set(State.TRANSIENT_FORK_OUTLIVING);
          pruningStrategy.prepare();
          markBlockNumber = blockNumber;
        }
      case TRANSIENT_FORK_OUTLIVING:
        {
          if (blockNumber >= markBlockNumber + transientForkThreshold) {
            state.set(State.MARKING);
            markedBlockHeader = blockchain.getBlockHeader(markBlockNumber).get();
            mark(markedBlockHeader);
          } else break;
        }
      case MARKING_COMPLETE:
        {
          if (state.get() != State.MARKING_COMPLETE) // possible if we flow over from previous case
          break;
          if (blockNumber >= markBlockNumber + retentionPeriodInBlocks
              && blockchain.blockIsOnCanonicalChain(markedBlockHeader.getHash())) {
            state.set(State.SWEEPING);
            sweep();
          }
        }
    }
```

shemnon: Looks good.  
Style wise I would drop the braces for the case bodies.  Also I would include a `// nobreak` or `// fallthrough` line in the cases you are intending to fall through, for explicit readability.  And the `else break;` will require braces but spotless will catch it.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/worldstate/Pruner.java
```diff
@@ -0,0 +1,133 @@
...
+      final MarkSweepPruner pruningStrategy,
+      final Blockchain blockchain,
+      final ExecutorService executorService,
+      final long transientForkThreshold,
+      final long retentionPeriodInBlocks) {
+    this.pruningStrategy = pruningStrategy;
+    this.executorService = executorService;
+    this.blockchain = blockchain;
+    if (!(transientForkThreshold >= 0 && retentionPeriodInBlocks >= transientForkThreshold)) {
+      throw new IllegalArgumentException(
```
mbaxter: Just a note that this exception might be a bit aggressive.  We can throw if the values are < 0,  otherwise if one value is just overshadowing the other, I'd log a warning.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/worldstate/Pruner.java
RatanRSur: After talking with Meredith, we decided that switching back to the `if/else if` and simply not allowing the case where you cascade through multiple states in one go because it probably won't matter to the user and it makes the code a lot cleaner.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/worldstate/Pruner.java
```diff
@@ -0,0 +1,127 @@
...
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class Pruner {
+  private static final Logger LOG = LogManager.getLogger();
+
+  private final MarkSweepPruner pruningStrategy;
+  private final Blockchain blockchain;
+  private ExecutorService executorService;
```
shemnon: nit: can this be made final?  I think it is only set in the constructor.

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/worldstate/PrunerTest.java
```diff
@@ -0,0 +1,173 @@
...
+    appendBlockWithParent(blockchain, forkBlock);
+    verify(markSweepPruner).mark(initiallyCanonicalBlock.getHeader().getStateRoot());
+    verify(markSweepPruner, never()).sweep();
+    pruner.stop();
+  }
+
+  @Test
+  public void shouldRejectInvalidArguments() {
+    final Blockchain mockchain = mock(Blockchain.class);
+    assertThatThrownBy(() -> new Pruner(markSweepPruner, mockchain, mockExecutorService, -1, -2))
```
shemnon: nit: in a previous commit you covered multiple cases where only one arg was illegal (i.e. 0, -1 and what should have been -1, 0). I felt that test had more coverage.

RatanRSur: Yes, meredith and I realized that the case that I thought was illegal is actually possible, so we relaxed the cases where the arguments were illegal.

PR #1644, opened 2019-07-04
remove the docker gradle plugin, handle building docker with shell now

joshuafernandes:
## Fixed Issue(s)

- remove the docker gradle plugin, handle building docker with shell now


build.gradle
```diff
@@ -481,7 +480,7 @@ distZip {
 // simple for use in docker
 tasks.register("dockerDistUntar") {
   dependsOn distTar
-  def dockerBuildDir = "build/docker-pantheon/"
+  def dockerBuildDir = "kubernetes/"
```
ajsutton: We shouldn't spew generated files into our source tree so I suspect we want this back under `build` somewhere, even if we have to copy the contents of `kubernetes` over.

joshuafernandes: fixed, i'm using the original dir now

PR #1646, opened 2019-07-05
[PAN-2879] Support multiple private marker transactions in a block 

rojotek:
Ensure multiple privacy marker transactions can appear in the same block.


config/src/main/resources/dev.json
```diff
@@ -1,6 +1,6 @@
 {
   "config": {
-    "chainId": 2018,
+    "chainId": 1,
```
rain-on: not sure what number this should be - do we want to avoid low numbers due to conflicts with existing chains? AJS would be smarter than I...

PR #1650, opened 2019-07-06
[PAN-2846] Added eea_getPrivacyPrecompileAddress

MadelineMurray:
Added eea_getPrivacyPrecompileAddress


docs/Reference/Pantheon-API-Methods.md
```diff
@@ -3738,6 +3738,36 @@ None
...
+        
+    ```bash tab="wscat WS request"
+    {"jsonrpc":"2.0","method":"eea_getPrivacyPrecompileAddress","params":[], "id":1}
+    ```
+        
+    ```json tab="JSON result"
+    {
+        "jsonrpc": "2.0",
+        "id": 1,
+        "result": 126
```
iikirilov: I think this should be the full ethereum hex encoded address i.e. "0x00000.....7E"

Puneetha17: No I don't think.
We had discussed this in a call and decided to keep it `int` when we pass it as input, such that the its not confusing to the user. And so the output should be consistent with it and return as `int`.

iikirilov: Yes, you specify it as an integer int the command line - but when you want to use it, it is an ethereum address.

If we return the integer, the caller needs to convert the response into an ethereum address to use it. At the very least we should have instructions on how to interpret the output of this call.

MadelineMurray: When you say _when you want to use it, it is an ethereum address_, when and how would you be using it? 

I think if using it is going to require converting the response and doc on the fact that conversion is  needed, it probably makes more sense to have it return the address rather than the int. 

We did discuss it in the call and the reasoning for keeping it as an int was it's a precompile and all the other precompiles are indentified by ints (I think).

iikirilov: So from a user story point of view I would say "I want a new precompile and it will be the precompile at position `x`" where `x` is a number.

Then once that precompile is put in position `x` the story becomes "I want to use my new precompile at position `x` so I will send a transaction to `y`" where `y` is the ethereum address representing `x`.

MadelineMurray: OK - I think we really do need to update eea_getPrivacyPrecompileAddress to return the address in hex then.  I was wondering what the 7E address I was seeing in transaction objects was - I can see now it's 126 in hex.

docs/Reference/Pantheon-API-Methods.md
```diff
@@ -3738,6 +3738,36 @@ None
     The `EEA` API methods are not enabled by default for JSON-RPC. Use the [`--rpc-http-api`](Pantheon-CLI-Syntax.md#rpc-http-api) 
     or [`--rpc-ws-api`](Pantheon-CLI-Syntax.md#rpc-ws-api) options to enable the `EEA` API methods.
 
+### eea_getPrivacyPrecompileAddress
+
+Returns the address to which the [privacy pre-compiled contract](../Privacy/Private-Transaction-Processing.md) is mapped.
```
bgravenorst: Returns the address of the privacy pre-compiled contract.

MadelineMurray: Fixed

PR #1651, opened 2019-07-06
Added transaction validation content and image

MadelineMurray:
Added transaction validation content and image


docs/Using-Pantheon/Transactions/Transaction-Validation.md
```diff
@@ -0,0 +1,16 @@
...
+
+When transactions are submitted and added to a block, validations are performed as illustrated. 
+
+![Transaction Validation](../../images/transaction-validation.png)
+
+The set of transaction pool validations are repeated when the transaction is propagated. The same set of 
+validations are repeated when the block including the transaction is imported except the nonce must be 
+exactly right when importing the block. 
+
+When the transaction is added to a block an additional validation is performed to check the total gas for the transaction
```
ajsutton: check the transaction gas limit is less than the remaining gas limit for the block.  People commonly expect the transaction to be added if its total gas used is less than the remaining block gas limit but that's not true - the entire transaction gas limit has to fit in the remaining block gas.

MadelineMurray: Fixed and added an additional note to make this distinction clearer.

PR #1656, opened 2019-07-08
PAN-2324  Support changing log level at runtime

matkt:
## PR description

Add json rpc method to change the logs without restarting the client


ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/AdminChangeLogLevel.java
```diff
@@ -0,0 +1,54 @@
...
+  @Override
+  public String getName() {
+    return RpcMethod.ADMIN_CHANGE_LOG_LEVEL.getMethodName();
+  }
+
+  @Override
+  public JsonRpcResponse response(final JsonRpcRequest request) {
+    try {
+      final Level logLevel = parameters.required(request.getParams(), 0, Level.class);
+      LOG.debug(String.format("Setting logging level to %s", logLevel.name()));
```
shemnon: HMM... I think admin alteration of logging level should always be logged, regardless.  At debug we won't see info->warn->fatal->info.  This looks a bit counterintuitive but I think should work.

```suggestion
      LOG.log(Level.OFF, String.format("Setting logging level to %s", logLevel.name()));
```

I don't think there's hooks for preventing logging of the `OFF` level, but if there are then it should be done as `LOG.fatal` level.

matkt: If we do not want to depend on the logger we can maybe use "System.out.println" for this message?

shemnon: The target of the log may not always be the ConsoleLogger, so I think we should stick with messages on the logger.  We uses System.out in pantheon command as it is startup code and we have more assurance we are on the console at that point.

matkt: ok you're absolutely right. I add your modification.

PR #1658, opened 2019-07-09
[PAN-2733] [PAN-2514] Java 11 updates

shemnon:
## PR description

* Update the build to require and target Java 11
* Add JVM flags to suppress specific illegal access warnings
* Remove Java 8 accommodations


build.gradle
```diff
@@ -261,11 +267,7 @@ allprojects {
   // Normalise Xdoclint behaviour across JDKs (OpenJDK 8 is more lenient than Oracle JDK by default).
```
ajsutton: We can probably remove this comment now.

PR #1660, opened 2019-07-09
add Prerequisites and Quick-Start

lazaridiscom:
* Prerequisites should further clarify which JDKs are surpported (Oracle, OpenJDK,)
* Quick-Start provides the super-fast instructions


lazaridiscom:
further simplified. The goal is that an experience java-dev can get started immediately (starting build, go on reading).

PR #1659, opened 2019-07-09
Removing unused methods on KeyValueStorage

CjHare:
## PR description
These methods are unused. 
Applying the principle of YANGI and in preparation for supporting the plugin system for storage, this PR delete them.

PR #1661, opened 2019-07-09
AT- Removing unused methods on KeyValueStorage

CjHare:
These methods are unused.
Applying the principle of YANGI and in preparation for supporting the plugin system for storage, this PR delete them.


CjHare:
This is a duplicate of #1659 with the exception that the submodule for the reference tests is correct

RatanRSur:
I need `containsKey` and `clear` but I think you're good to go to remove Entry

CjHare:
@RatanRSur I've added back `containsKey` and `clear` ...for fun I ordered the methods alphabetically.

The only 'real' change now is removing the unused `EntryKey` from `KeyValueStorage`

CjHare:
👍

PR #1663, opened 2019-07-09
Factorise metrics code for KeyValueStorage database

abdelhamidbakhta:
## PR description
- introduce `RocksDBMetricsHelper `
- factorise code
- remove metrics instanciation in `RocksDbKeyValueStorage` and `ColumnarRocksDbKeyValueStorage`

### DIFF
--- a/services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/MonitorableKeyValueStorage.java
+++ b/services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/MonitorableKeyValueStorage.java
@@ -1,0 +1,150 @@
+/*
+ * Copyright 2019 ConsenSys AG.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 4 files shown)


services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/MonitorableKeyValueStorage.java
```diff
@@ -0,0 +1,150 @@
...
+import tech.pegasys.pantheon.metrics.prometheus.PrometheusMetricsSystem;
+import tech.pegasys.pantheon.metrics.rocksdb.RocksDBStats;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.rocksdb.RocksDBException;
+import org.rocksdb.Statistics;
+import org.rocksdb.TransactionDB;
+
+public class MonitorableKeyValueStorage {
```
shemnon: This has RocksDB specific parts inside of it, so I think RocksDB should be part of the name somehow.  And it doesn't actually provide the KeyValueStorage.  Perhaps `RockDBMetrics` or `RocksDBMetricsHelper`?

abdelhamidbakhta: Yes, sounds good, i rename it

PR #1664, opened 2019-07-10
New release docker image

joshuafernandes:
## Fixed Issue(s)

- Single docker image going forward using the one in the kubernetes folder, removing the old docker image
- Updated release process so we do the bintray and docker image in one step now


Jenkinsfile.release
```diff
@@ -9,12 +9,14 @@ void tryAddKnownHost(String hostUrl){
 
 pipeline {
     agent {
-        docker { image 'pegasyseng/pantheon-build:0.0.5-jdk8' }
+        docker { image 'pegasyseng/pantheon-build:0.0.5-jdk11' }
     }
     parameters {
         string(name: 'BRANCH_NAME', defaultValue: 'master', description: '[MANDATORY] The name of the branch to create the release from')
-        string(name: 'RELEASE_VERSION', defaultValue: '', description: '[OPTIONAL] When empty: defaults to the current project version')
+        string(name: 'RELEASE_VERSION', defaultValue: '1.x', description: '[MANDATORY] The version you are deploying')
```
ajsutton: `1.x` as the default seems misleading here - our versions would have three parts (1.x.x) and having a default will make it harder to detect if someone misses filling in this field.

Jenkinsfile.release
```diff
@@ -9,12 +9,14 @@ void tryAddKnownHost(String hostUrl){
...
         string(name: 'NEXT_VERSION', defaultValue: '', description: '[OPTIONAL] When empty: defaults to next patch version after current project version')
+        string(name: 'DOCKER_ADDITIONAL_TAGS', defaultValue: '', description: '[OPTIONAL] When empty: you get the release version as a tag only')
```
ajsutton: Might be a follow-up PR but can we avoid the need for this?  The tags would be `latest` and the 2-part form of the version (so for a 1.2.1 release it would be `latest,1.2`).  For RC releases I think we'd leave this blank so would probably need to make it a checkbox to indicate if this was a stable release or not.  Hopefully we could calculate everything else from that - maybe even just treat any version with a `-` as a non-stable release (e.g `1.2.0` is stable but `1.2.0-BETA` or `1.2.0-RC` are not and so don't get the additional tags).

This field confuses the heck of out everyone doing the release so would be good if we can simplify it.

Jenkinsfile.release
```diff
@@ -9,12 +9,14 @@ void tryAddKnownHost(String hostUrl){
...
+        booleanParam(name: 'PUSH_RELEASE', defaultValue: true, description: '[OPTIONAL] Push the docker images to dockerhub')
```
ajsutton: Is there a reason we'd release without releasing the docker images?

Errorific: The push was originally so we could trigger a test build without pushing it to servers. Might no longer make sense

Jenkinsfile.release
joshuafernandes: updated to x.y.z

Jenkinsfile.release
joshuafernandes: I've removed this param now

Jenkinsfile.release
joshuafernandes: Updated and removed the addditional_tags param. Behaviour is as follows:
- every build gets tagged 'develop'
- every release also gets an additional '<release_version>'
- release where a version contains no hyphen i.e its stable will get additional 'latest', and '<major,minor>' tags

PR #1665, opened 2019-07-10
[PAN-1683] Limit the fraction of wire connections initiated by peers

abdelhamidbakhta:
## PR description
To protect against eclipse attacks, we should not allow all of our wire connections to be initiated from network peers. Some fraction of wire connections should be initiated by our node by connecting to peers in our discovery peer table. This PR ensures the fraction abides the limit.
- `RlpxAgent` checks the `fractionRemoteConnectionsAllowed` in `handleIncomingConnection` method
- add a cli option to configure the fraction: `--fraction-remote-connections-allowed`
- introduce `Fraction` class to handle the conversion of the CLI option to a double and check if the value is between 0.0 and 1.0
- added explicit CI flag to enable the protection against eclipse attack: `--limit-remote-wire-connections-enabled`


ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/rlpx/RlpxAgent.java
```diff
@@ -374,6 +390,21 @@ private void handleIncomingConnection(final PeerConnection peerConnection) {
...
+        Math.toIntExact(
+            connectionsById.values().stream()
+                .filter(RlpxConnection::isActive)
+                .filter(RlpxConnection::initiatedRemotely)
+                .count());
+    if (remotelyInitiatedConnectionsCount == 0) {
+      return true;
+    }
+    final double fractionRemoteConnections =
+        (double) remotelyInitiatedConnectionsCount / (double) getConnectionCount();
```
shemnon: I think this should be against the max allowed connection count, not a ratio against the current set of connection.  @mbaxter, what's your opinion?

abdelhamidbakhta: yes i think it would be better, i update the computation

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/rlpx/RlpxAgent.java
```diff
@@ -329,6 +335,16 @@ private void handleIncomingConnection(final PeerConnection peerConnection) {
       return;
     }
 
+    // Disconnect if the fraction of wire connections initiated by peers is too high to protect
+    // against eclipse attacks
+    if (limitRemoteWireConnectionsEnabled && !doTheFractionOfRemoteConnectionsAllowsNewOne()) {
```
shemnon: for readability I would change the function to return a `true` if the ratio is too high, so we don't have a negation '!' in the boolean logic.  We also negate it inside the method.  Also, the name would need to change.

```suggestion
    if (limitRemoteWireConnectionsEnabled && remoteConnectionExceedsLimit()) {
```

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/rlpx/RlpxAgent.java
```diff
@@ -374,6 +390,21 @@ private void handleIncomingConnection(final PeerConnection peerConnection) {
     enforceConnectionLimits();
   }
 
+  private boolean doTheFractionOfRemoteConnectionsAllowsNewOne() {
```
shemnon:
```suggestion
  private boolean remoteConnectionExceedsLimit() {
```

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/rlpx/RlpxAgent.java
abdelhamidbakhta: Ok.

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/rlpx/RlpxAgent.java
```diff
@@ -374,6 +390,21 @@ private void handleIncomingConnection(final PeerConnection peerConnection) {
...
+            connectionsById.values().stream()
+                .filter(RlpxConnection::isActive)
+                .filter(RlpxConnection::initiatedRemotely)
+                .count());
+    if (remotelyInitiatedConnectionsCount == 0) {
+      return true;
+    }
+    final double fractionRemoteConnections =
+        (double) remotelyInitiatedConnectionsCount / (double) getConnectionCount();
+    return !(fractionRemoteConnections > fractionRemoteConnectionsAllowed);
```
shemnon:
```suggestion
    return fractionRemoteConnections > fractionRemoteConnectionsAllowed;
```

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/rlpx/RlpxAgent.java
abdelhamidbakhta: Done.

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -236,6 +237,21 @@ void setBootnodes(final List<String> values) {
           "Maximum P2P peer connections that can be established (default: ${DEFAULT-VALUE})")
   private final Integer maxPeers = DEFAULT_MAX_PEERS;
 
+  @Option(
```
shemnon: I think these names need a once over for UX.  I opened PIE-1766 to cover that.

PR #1667, opened 2019-07-10
[PAN-2852] privacy group id consistent

iikirilov:
# PR description
Handle base64 <-> BytesValue conversion uniformly throughout Pantheon

As a side effect also fixes the private transaction gas required to 23992 as specified in [PAN-2795]

## Fixed Issue(s)
[PAN-2852]

[PAN-2795]: https://pegasys1.atlassian.net/browse/PAN-2795
[PAN-2852]: https://pegasys1.atlassian.net/browse/PAN-2852


acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/web3j/privacy/EnclaveErrorAcceptanceTest.java
```diff
@@ -32,7 +32,7 @@
 import org.junit.Before;
 import org.junit.Test;
 
-public class PrivateTxEnclaveErrorAcceptanceTest extends PrivacyAcceptanceTestBase {
+public class EnclaveErrorAcceptanceTest extends PrivacyAcceptanceTestBase {
```
shemnon: Did this file get renamed?  It's unclear from the Github commit.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/precompiles/privacy/PrivacyPrecompiledContract.java
```diff
@@ -90,7 +88,7 @@ public void setPrivateTransactionProcessor(
 
   @Override
   public Gas gasRequirement(final BytesValue input) {
-    return Gas.of(40_000L); // Not sure
+    return Gas.of(0);
```
shemnon: Is this the EEA specified gas?

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/privacy/EeaDeletePrivacyGroup.java
```diff
@@ -46,8 +46,8 @@ public String getName() {
   public JsonRpcResponse response(final JsonRpcRequest request) {
     LOG.trace("Executing {}", RpcMethod.EEA_DELETE_PRIVACY_GROUP.getMethodName());
 
-    final String privacyGroupId = parameters.required(request.getParams(), 1, String.class);
```
shemnon: This seems like a non-change.  If there are any other alterations can this be reverted?

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/privacy/EeaGetTransactionCount.java
```diff
@@ -51,6 +51,7 @@ public JsonRpcResponse response(final JsonRpcRequest request) {
     final String privacyGroupId = parameters.required(request.getParams(), 1, String.class);
 
     final long nonce = privateTransactionHandler.getSenderNonce(address, privacyGroupId);
+
```
shemnon: Again, another non-change.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/web3j/privacy/EnclaveErrorAcceptanceTest.java
iikirilov: Yes, the file was renamed. It should be `PrivateTransactionEnclaveErrorAcceptanceTest` but that is a mouthful so I have gone with `EnclaveErrorAcceptanceTest` which describes the test well

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/precompiles/privacy/PrivacyPrecompiledContract.java
iikirilov: There is no requirement in the EEA spec. The reasoning here is that implementing EEA compliant private transaction support without a Precompile contract would cost 21000 gas like a normal transaction. This precompile should not consume gas.

PR #1668, opened 2019-07-10
PAN-2076  Update PantheonCommand to accept minTransactionGasPriceWei as an integer

matkt:
## PR description

Eliminates the need to put quotes in a TOML file for the min-gas-price field. this field is now an integer


matkt:
> I think a lower touch solution would be to add an if block for Wei in the TomlConfigFileDefaultProvider -
> 
> https://github.com/PegaSysEng/pantheon/blob/94f39d1d9c93f89c374ce1bc0e877aecb1545d44/pantheon/src/main/java/tech/pegasys/pantheon/cli/util/TomlConfigFileDefaultProvider.java#L59

Indeed changing this file is easier and faster. Good comment

PR #1671, opened 2019-07-11
[PAN-2829] - add hooks for validation 

shemnon:
## PR description

EIP-1602 specifies a new validation step.  Add hooks for those to occur.
Move contract size checking into the validation hooks.


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/ContractValidationRule.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package tech.pegasys.pantheon.ethereum.mainnet;
+
+import tech.pegasys.pantheon.ethereum.vm.MessageFrame;
+
+public interface ContractValidationRule {
```
abdelhamidbakhta: (Discussion) What about using a FunctionalInterface ?

shemnon: It may gain more methods in the future, but we can walk it back if that becomes the case.

PR #1672, opened 2019-07-11
[PAN-2897] Return hashrate of 0 when not mining.

rojotek:
Return Hashrate of 0 when not mining.


ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/EthHashrate.java
```diff
@@ -40,10 +40,9 @@ public String getName() {
   public JsonRpcResponse response(final JsonRpcRequest req) {
     try {
       final Optional<Long> hashesPerSecond = miningCoordinator.hashesPerSecond();
-      if (hashesPerSecond.isPresent()) {
-        return new JsonRpcSuccessResponse(req.getId(), Quantity.create(hashesPerSecond.get()));
-      }
-      return new JsonRpcErrorResponse(req.getId(), JsonRpcError.NO_HASHES_PER_SECOND);
+      return hashesPerSecond
+          .map(hashes -> new JsonRpcSuccessResponse(req.getId(), Quantity.create(hashes)))
+          .orElseGet(() -> new JsonRpcSuccessResponse(req.getId(), Quantity.create(0)));
```
ajsutton: `return new JsonRpcSuccessResponse(req.getId(), Quantity.create(hashesPerSecond.orElse(0))`?

rojotek: yep -- I made it better -- you've helped more -- thanks.

ajsutton:
You've also accidentally reverted the reference test upgrade.  Checkout master, run `git submodule update` then checkout this branch again and `git commit -a` I think to commit it back to the right version so the build passes.

PR #1675, opened 2019-07-11
[PAN-2829] Add accountVersion to MessageFrame

shemnon:
## PR description

Add the current contract's version to the message frame. For a contract
creation transaction the version is the fork default version.


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/vm/MessageFrame.java
```diff
@@ -912,6 +920,11 @@ public Builder contract(final Address contract) {
       return this;
     }
 
+    public Builder contractAccountVersion(final int contractAccountVersion) {
+      this.contractAccountVersion = contractAccountVersion;
```
mbaxter: Probably worth asserting the value is `>= 0`.

shemnon: done

PR #1676, opened 2019-07-11
[PAN-1683] Fix handling of remote connection limit

mbaxter:
## PR description
This PR fixes a few bugs in how remote connection limits are enforced:
* Fix an order of operations error in the calculation of the fraction of remotely connected peers
* Remove the branch that returns early if no peers are currently connected (its possible that 1 remote connection is too many)
* Change disconnect reason to `TOO_MANY_PEERS`
* Validate that `fractionRemoteWireConnectionsAllowed` is between 0.0 and 1.0 inclusive
  * A fraction of 0 is valid, meaning no untrusted remote connections should be allowed
* Add additional tests
* Ignore tests that are broken by this new functionality (for now).  These limits will be reworked in a follow-up PR and these test will be re-enabled.


ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/rlpx/RlpxAgent.java
```diff
@@ -397,11 +397,8 @@ private boolean remoteConnectionExceedsLimit() {
                 .filter(RlpxConnection::isActive)
                 .filter(RlpxConnection::initiatedRemotely)
                 .count());
-    if (remotelyInitiatedConnectionsCount == 0) {
-      return false;
-    }
     final double fractionRemoteConnections =
-        (double) remotelyInitiatedConnectionsCount + 1 / (double) maxPeers;
+        (double) (remotelyInitiatedConnectionsCount + 1) / (double) maxPeers;
```
shemnon: Will maxPeers=0 cause issues?

mbaxter: good eye!  however, turns out this actually works.  this will evaluate to `Infinity` in that case and the comparison will return `true`.

PR #1682, opened 2019-07-12
Use a percentage instead of fraction for the remote connections percentage CLI option.

abdelhamidbakhta:
## PR description
Use a percentage instead of fraction for the remote connections percentage CLI option


pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -238,19 +239,19 @@ void setBootnodes(final List<String> values) {
   private final Integer maxPeers = DEFAULT_MAX_PEERS;
 
   @Option(
-      names = {"--limit-remote-wire-connections-enabled"},
+      names = {"--remote-connections-limit-enabled"},
```
mbaxter: Question - why did we end up with 2 options here (enabled + percentage), instead of 1 (just percentage)?  If the percentage is at 1, the limit is effectively disabled.

pantheon/src/test/java/tech/pegasys/pantheon/cli/PantheonCommandTest.java
```diff
@@ -1239,47 +1242,48 @@ public void maxpeersOptionMustBeUsed() {
...
-        "not-a-fraction");
+        "--remote-connections-limit-enabled",
+        "--remote-connections-percentage",
+        "not-a-percentage");
     verifyZeroInteractions(mockRunnerBuilder);
     assertThat(commandOutput.toString()).isEmpty();
     assertThat(commandErrorOutput.toString())
         .contains(
-            "Invalid value for option '--fraction-remote-connections-allowed': cannot convert 'not-a-fraction' to Fraction");
+            "Invalid value for option '--remote-connections-percentage': cannot convert 'not-a-percentage' to Percentage");
```
mbaxter: Question - where does this phrase "cannot convert 'not-a-percentage'" come from?  Is that generated internally by PicoCLI?  If we add a custom error message to the exception that's thrown can we get a better error message?

util/src/main/java/tech/pegasys/pantheon/util/number/Fraction.java
```diff
@@ -25,6 +25,14 @@ private Fraction(final double value) {
     this.value = value;
   }
 
+  public static Fraction fromPercentage(final Percentage percentage) {
```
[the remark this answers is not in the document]
abdelhamidbakhta: Yes it makes sense

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
abdelhamidbakhta: We agreed that this is such an important option that the user should explicitly enable this protection. It has a strong impact on the behaviour of the node.

shemnon: It is easier from a UX perspective to configure with a feature flag and a value flag.  I also ran this by Jake and there are many places we do this so it is a well worn pattern now.

pantheon/src/test/java/tech/pegasys/pantheon/cli/PantheonCommandTest.java
abdelhamidbakhta: Ok, will do the same logic applied for RPC APIs conversion.

pantheon/src/main/java/tech/pegasys/pantheon/cli/converter/FractionConverter.java
```diff
@@ -0,0 +1,30 @@
...
+
+import tech.pegasys.pantheon.cli.converter.exception.FractionConversionException;
+import tech.pegasys.pantheon.util.number.Fraction;
+
+import picocli.CommandLine;
+
+public class FractionConverter implements CommandLine.ITypeConverter<Fraction> {
+
+  @Override
+  public Fraction convert(final String value) throws FractionConversionException {
```
mbaxter: Can this just convert to a `double`?

pantheon/src/main/java/tech/pegasys/pantheon/cli/converter/PercentageConverter.java
```diff
@@ -0,0 +1,30 @@
...
+
+import tech.pegasys.pantheon.cli.converter.exception.PercentageConversionException;
+import tech.pegasys.pantheon.util.number.Percentage;
+
+import picocli.CommandLine;
+
+public class PercentageConverter implements CommandLine.ITypeConverter<Percentage> {
+
+  @Override
+  public Percentage convert(final String value) throws PercentageConversionException {
```
mbaxter: Can this just convert to a `float`?

abdelhamidbakhta: ok

abdelhamidbakhta: a int actually

pantheon/src/main/java/tech/pegasys/pantheon/cli/converter/FractionConverter.java
abdelhamidbakhta: ok

pantheon/src/main/java/tech/pegasys/pantheon/cli/converter/PercentageConverter.java
abdelhamidbakhta: done.

pantheon/src/main/java/tech/pegasys/pantheon/cli/converter/FractionConverter.java
abdelhamidbakhta: done.

pantheon/src/test/java/tech/pegasys/pantheon/cli/PantheonCommandTest.java
abdelhamidbakhta: done.

util/src/main/java/tech/pegasys/pantheon/util/number/Fraction.java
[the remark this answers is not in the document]
abdelhamidbakhta: done

PR #1683, opened 2019-07-12
 [PAN-2829] Add opcode and precompiled support for versioning 

shemnon:
## PR description

Different opcode sets and precompiled contract sets for different account versions


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/vm/OperationRegistry.java
```diff
@@ -12,31 +12,39 @@
...
+  private final Operation[][] operations;
 
   public OperationRegistry() {
-    this.operations = new Operation[NUM_OPERATIONS];
+    this(1);
+  }
+
+  public OperationRegistry(final int numVersions) {
+    Preconditions.checkArgument(numVersions >= 1);
+    this.operations = new Operation[numVersions][NUM_OPERATIONS];
```
mbaxter: (Optional) You might look into guava's `RangeMap`, which could map a range of versions to a single operation, rather than creating distinct references for every op-version combo.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/Account.java
```diff
@@ -39,6 +39,7 @@
   long DEFAULT_NONCE = 0L;
   Wei DEFAULT_BALANCE = Wei.ZERO;
   int DEFAULT_VERSION = 0;
+  int ISTANBUL_VERSION = 1;
```
mbaxter: (Optional) Not sure about storing `ISTANBUL_VERSION` here - it feels very mainnet-evm-specific.  Looks like you could get away with just keeping this as a local variable in `MainnetProtocolSpecs`.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/MainnetPrecompiledContractRegistries.java
```diff
@@ -30,45 +31,70 @@
...
     final PrecompileContractRegistry registry = new PrecompileContractRegistry();
-    populateForByzantium(registry, precompiledContractConfiguration.getGasCalculator());
+    populateForByzantium(
+        registry, precompiledContractConfiguration.getGasCalculator(), Account.DEFAULT_VERSION);
     return registry;
   }
 
-  public static PrecompileContractRegistry appendPrivacy(
-      final PrecompileContractRegistry registry,
+  public static PrecompileContractRegistry istanbul(
```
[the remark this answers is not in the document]
shemnon: I missed adding it to the spec builder.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/vm/OperationRegistry.java
shemnon: Since this is a "hotspot" I opted for native arrays, on the presumption native arrays are faster than interfaces.  When we get better performance monitoring in we can see how much the impact is.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/Account.java
shemnon: Done

PR #1684, opened 2019-07-12
Enclave refactoring

abdelhamidbakhta:
## PR description
- use final modifiers when appropriate to comply with Pantheon global policy
- use `assertj` assertions in tests
- don't use local variables when not necessary
- check exception message for `upCheck` method


enclave/build.gradle
```diff
@@ -7,8 +7,10 @@ dependencies {
   testImplementation 'junit:junit'
   testImplementation project(':testutil')
 
-  // integration test dependacies.
+  // integration test dependencies.
   integrationTestImplementation 'junit:junit'
+  integrationTestImplementation 'org.assertj:assertj-core'
```
iikirilov: what is the reason for using assertj over junit?

abdelhamidbakhta: Readability of the code mainly, it looks more a human natural semantic. Easier to read for non tech people for example. Also we try to use it in the rest of the code when possible. So  also for harmonization.

PR #1687, opened 2019-07-13
[PAN-2763]  Add JSON-RPC API to report validator block production information

matkt:
## PR description

Add JSON-RPC API to report validator block production information (CLIQUE and IBFT2)

-  The number of blocks from each proposer in a given block range. 
-  All validators present in the last block of the range should be listed are included even if they didn’t propose a block
-  The block number of the last block proposed by each validator (if any within the given range).


matkt:
Input 
``` json
{"jsonrpc":"2.0","method":"clique_getReportValidatorBlockProduction","params":["1","1000"], "id":1}
``` 

Ouput 

``` json
 {
    "jsonrpc": "2.0",
    "id": 1,
    "result": {
        "proposers": [
            {
                "address": "0x7ffc57839b00206d1ad20c69a1981b489f772031",
                "nbBlockProposed": 315
            },
            {
                "address": "0x42eb768f2244c8811c63729a21a3569731535f06",
                "nbBlockProposed": 342
            },
            {
                "address": "0xb279182d99e65703f0076e4812653aab85fca0f0",
                "nbBlockProposed": 342
            }
        ],
        "validatorsPresentInLastBlock": [
            {
                "address": "0x7ffc57839b00206d1ad20c69a1981b489f772031",
                "lastBlockProposed": 997
            },
            {
                "address": "0x42eb768f2244c8811c63729a21a3569731535f06",
                "lastBlockProposed": 999
            },
            {
                "address": "0xb279182d99e65703f0076e4812653aab85fca0f0",
                "lastBlockProposed": 998
            }
        ]
    }
}
```

consensus/common/src/main/java/tech/pegasys/pantheon/consensus/common/jsonrpc/AbstractGetSignerMetricsMethod.java
```diff
@@ -0,0 +1,127 @@
...
+  public JsonRpcResponse response(final JsonRpcRequest request) {
+
+    final long startBlock = parameters.required(request.getParams(), 0, Long.class);
+    final long endBlock = parameters.required(request.getParams(), 1, Long.class);
+
+    if (isValidParameters(startBlock, endBlock)) {
+
+      final Map<Address, ProposerReportBlockProductionResult> proposersMap = new HashMap<>();
+      final Map<Address, ValidatorReportBlockProductionResult> validatorInLastBlockMap =
+          new HashMap<>();
```
ajsutton: Hopefully the clarifications in my overall comment help to make sense of this but we'd only need the one map here.  The basic approach would be:

1. Iterate through each block in the range and increment the count of number of proposed blocks for the proposer of that block.  Note you might still get some proposers who don't yet exist in the map if they've been voted out before the end of the range.
2. For each validator in the last block of the range not already in `proposerMap`, add it with 0 proposed blocks. That will ensure they're all present in the report even if none ever propose blocks.

consensus/common/src/main/java/tech/pegasys/pantheon/consensus/common/jsonrpc/AbstractGetSignerMetricsMethod.java
```diff
@@ -0,0 +1,127 @@
...
+
+      // go through each block (startBlock is inclusive and endBlock is exclusive)
+      LongStream.range(startBlock, endBlock)
+          .boxed()
+          .sorted(Collections.reverseOrder())
+          .forEach(
```
ajsutton: It's actually simpler and probably faster to iterate this from start to end block so you can just use `LongStream.range(startBlock, endBlock).forEach(...)`.

Going forwards through the range makes more sense because the list of validators is adjusted based on votes from  in the block so you have to process blocks in order to know who the current list of validators is.  Pantheon has some caching for those results so we don't always have to reprocess a lot of blocks, but if it's a big range it will be a lot faster to go forwards than backwards because it works better with that caching.

consensus/common/src/main/java/tech/pegasys/pantheon/consensus/common/jsonrpc/AbstractGetSignerMetricsMethod.java
```diff
@@ -0,0 +1,127 @@
...
+                          .forEach(
+                              proposerAddress -> {
+                                final ProposerReportBlockProductionResult proposer =
+                                    proposersMap.get(proposerAddress);
+                                if (proposer != null) {
+                                  proposer.incrementeNbBlock();
+                                } else {
+                                  proposersMap.put(
+                                      proposerAddress,
+                                      new ProposerReportBlockProductionResult(proposerAddress));
```
ajsutton: The `Map.computeIfAbsent` is really helpful here.  If `ProposerReportBlockProductionResult` defaulted to 0 blocks proposed instead of 1, then you could just use `proposersMap.computeIfAbsent(proposerAddress, ProposerReportBlockProductionResult::new).incrementeNbBlock();`

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/results/report/block/production/ProposerReportBlockProductionResult.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.Objects;
+
+import com.fasterxml.jackson.annotation.JsonGetter;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+
+@JsonPropertyOrder({"address", "nbBlockProposed"})
+public class ProposerReportBlockProductionResult {
+
+  public final String address;
+  public long nbBlockProposed;
```
ajsutton: We generally try to avoid abbreviations so something like `proposedBlockCount` would be a better name.

consensus/common/src/main/java/tech/pegasys/pantheon/consensus/common/jsonrpc/AbstractGetSignerMetricsMethod.java
```diff
@@ -0,0 +1,127 @@
...
+      final JsonRpcParameter parameter) {
+    this.blockchainQueries = blockchainQueries;
+    this.voteTallyCache = voteTallyCache;
+    this.parameters = parameter;
+  }
+
+  public JsonRpcResponse response(final JsonRpcRequest request) {
+
+    final long startBlock = parameters.required(request.getParams(), 0, Long.class);
+    final long endBlock = parameters.required(request.getParams(), 1, Long.class);
```
ajsutton: We should support the usual values for block parameters like `latest`.  `parameters().required(request.getParams(), 0, BlockParameter.class);` will handle the parsing of that.  `pending` and `latest` would mean the same thing (current chain head) in this context.

I'd probably make both arguments optional as well.  Start would default to 100 blocks before current chain head and end defaults to chain head.  That way you can call with no arguments and get a useful result (most of the time you want to know which of the current validators hasn't proposed recently), or specify a starting block to get from their to chain head easily, or both start and end for a custom range.

consensus/common/src/main/java/tech/pegasys/pantheon/consensus/common/jsonrpc/AbstractGetSignerMetricsMethod.java
```diff
@@ -0,0 +1,123 @@
...
+
+    final Optional<BlockParameter> startBlockParameter =
+        parameters.optional(request.getParams(), 0, BlockParameter.class);
+    final Optional<BlockParameter> endBlockParameter =
+        parameters.optional(request.getParams(), 1, BlockParameter.class);
+
+    final long fromBlockNumber = getFromBlockNumber(startBlockParameter);
+    final long toBlockNumber = getEndBlockNumber(endBlockParameter);
+
+    if (isValidParameters(fromBlockNumber, toBlockNumber)) {
```
ajsutton: We typically exit early in cases like this, so
```
if (!isValidParameters(fromBlockNumber, toBlockNubmer)) {
  return new JsonRpcErrorResponse(request.getId(), JsonRpcError.INVALID_PARAMS);
}
```

consensus/common/src/main/java/tech/pegasys/pantheon/consensus/common/jsonrpc/AbstractGetSignerMetricsMethod.java
```diff
@@ -0,0 +1,123 @@
...
+                    .orElse(blockchainQueries.headBlockNumber() - DEFAULT_RANGE_BLOCK))
+        .orElseGet(() -> blockchainQueries.headBlockNumber() - DEFAULT_RANGE_BLOCK);
+  }
+
+  private long getEndBlockNumber(final Optional<BlockParameter> endBlockParameter) {
+    return endBlockParameter
+        .map(
+            blockParameter ->
+                blockParameter.getNumber().orElse(blockchainQueries.headBlockNumber()))
+        .orElseGet(blockchainQueries::headBlockNumber);
```
ajsutton: `BlockParameter` is slightly more annoying than you might imagine.  It's `getNumber` method will return empty if any of the "alias" options were used (`earliest`, `latest` or `pending`).

You'll need something like:
```
private long resolveBlockNumber(final BlockParameter param) {
   if (param.getNumber().isPresent()) {
       return param.getNumber().getAsLong();
   if (param.isEarliest) {
        return BlockHeader.GENESIS_BLOCK_NUMBER;
   } else if (param.isLatest() || param.isPending()) {
        return blockchain.getChainHeadBlockNumber();
   } else {
        throw new IllegalStateException("Unknown block parameter type.");
    }
  }
```

ajsutton:
I realised we should be reporting the values in the JSON response as hex (to match all the other quantity values).  This is unfortunately a lot less readable but just how JSON-RPC works.

https://github.com/ajsutton/pantheon/commit/0bd0892e36dc08e70f93f00660799d2c8dc900de applies the couple of suggestions I made above, plus the change to use hex. It's on the branch https://github.com/ajsutton/pantheon/commits/validator-metrics if it's easier to just pull that in.

matkt:
> I realised we should be reporting the values in the JSON response as hex (to match all the other quantity values). This is unfortunately a lot less readable but just how JSON-RPC works.
> 
> [ajsutton@0bd0892](https://github.com/ajsutton/pantheon/commit/0bd0892e36dc08e70f93f00660799d2c8dc900de) applies the couple of suggestions I made above, plus the change to use hex. It's on the branch https://github.com/ajsutton/pantheon/commits/validator-metrics if it's easier to just pull that in.

Ok thanks you very much for your modification. I will merge your branch

consensus/clique/src/test/java/tech/pegasys/pantheon/consensus/clique/jsonrpc/methods/CliqueGetSignerMetricsTest.java
```diff
@@ -0,0 +1,257 @@
...
+
+  @Test
+  public void getSignerMetricsWhenNoParams() {
+
+    Long startBlock = 1L;
+    Long endBlock = 3L;
+
+    when(blockchainQueries.headBlockNumber()).thenReturn(endBlock);
+
+    List<SignerMetricResult> signerMetricResultList = new java.util.ArrayList<>();
```
ajsutton: nit: Shouldn't don't need the full package name here
```suggestion
    List<SignerMetricResult> signerMetricResultList = new ArrayList<>();
```

ajsutton: Also for the other `new java.util.ArrayList()` calls below.

consensus/ibft/src/test/java/tech/pegasys/pantheon/consensus/ibft/jsonrpc/methods/IbftGetSignerMetricsTest.java
```diff
@@ -0,0 +1,269 @@
...
+    expectedException.expectMessage("Invalid json rpc parameter at index 1");
+
+    method.response(request);
+  }
+
+  @Test
+  public void getSignerMetricsWhenNoParams() {
+
+    Long startBlock = 1L;
+    Long endBlock = 3L;
```
ajsutton: nit: These should just be primitives and we add the `final` keyword whenever possible in Pantheon (well try to - we're not as consistent as we should be).  So:
```suggestion
    final long endBlock = 3L;
```

consensus/ibft/src/test/java/tech/pegasys/pantheon/consensus/ibft/jsonrpc/methods/IbftGetSignerMetricsTest.java
```diff
@@ -0,0 +1,269 @@
...
+        .forEach(value -> signerMetricResultList.add(generateBlock(value)));
+
+    final JsonRpcRequest request = requestWithParams();
+
+    final JsonRpcSuccessResponse response = (JsonRpcSuccessResponse) method.response(request);
+
+    LongStream.range(startBlock, endBlock)
+        .forEach(value -> verify(blockchainQueries).getBlockHeaderByNumber(value));
+
+    verify(blockchainQueries, atLeast(2)).headBlockNumber();
```
ajsutton: I don't think there's any reason to assert these class to `blockchainQueries` - they're just implementation details that it needs to get the information.  We just check it comes up with the right answer.

consensus/ibft/src/test/java/tech/pegasys/pantheon/consensus/ibft/jsonrpc/methods/IbftGetSignerMetricsTest.java
```diff
@@ -0,0 +1,269 @@
...
+
+    signerMetricResultList.forEach(
+        signerMetricResult ->
+            assertThat(response.getResult().toString()).contains(signerMetricResult.toString()));
```
ajsutton: These assertions should actually be stronger - we expect that the result only contains the expected signers and the order isn't important (because the list of stats is built off of a `HashMap` the order isn't guaranteed). So:

```
    assertThat((Collection<SignerMetricResult>) response.getResult())
        .containsExactlyInAnyOrderElementsOf(signerMetricResultList);
```

consensus/ibft/src/test/java/tech/pegasys/pantheon/consensus/ibft/jsonrpc/methods/IbftGetSignerMetricsTest.java
```diff
@@ -0,0 +1,269 @@
...
+  }
+
+  private JsonRpcRequest requestWithParams(final Object... params) {
+    return new JsonRpcRequest(JSON_RPC_VERSION, IBFT_METHOD, params);
+  }
+
+  private SignerMetricResult generateBlock(final long number) {
+    final SECP256K1.KeyPair proposerKeyPair = SECP256K1.KeyPair.generate();
+    final Address proposerAddressBlock = Util.publicKeyToAddress(proposerKeyPair.getPublicKey());
```
ajsutton: Creating keypairs is fairly expensive and we really want to test that the number of blocks gets incremented so it's probably worth creating a few key pairs as static fields which are just iterated through here (e.g. `final SECP256K1.KeyPair proposerKeyPair = KEYPAIRS[number % KEYPAIRS.length]`)

consensus/clique/src/test/java/tech/pegasys/pantheon/consensus/clique/jsonrpc/methods/CliqueGetSignerMetricsTest.java
matkt: yes an error that I had not seen. I correct that

consensus/ibft/src/test/java/tech/pegasys/pantheon/consensus/ibft/jsonrpc/methods/IbftGetSignerMetricsTest.java
matkt: I wanted to try this proposal but I have an error during the pipeline "unchecked cast assertThat ((Collection <SignerMetricResult>) response.getResult ()) "that's why I used another way

matkt: I will add the @SuppressWarnings("unchecked") annotation

consensus/ibft/src/test/java/tech/pegasys/pantheon/consensus/ibft/jsonrpc/methods/IbftGetSignerMetricsTest.java
matkt: done

consensus/ibft/src/test/java/tech/pegasys/pantheon/consensus/ibft/jsonrpc/methods/IbftGetSignerMetricsTest.java
matkt: is removed

consensus/ibft/src/test/java/tech/pegasys/pantheon/consensus/ibft/jsonrpc/methods/IbftGetSignerMetricsTest.java
matkt: done

PR #1690, opened 2019-07-14
[PAN-2903] Implement EIP-1344 ChainID Opcode

shemnon:
## PR description

Add the ChainID Opcode


shemnon:
Force push was a revert, EIP1108 was accidentally pushed on this branch.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/MainnetProtocolSpecs.java
```diff
@@ -285,12 +286,13 @@ private MainnetProtocolSpecs() {}
       final OptionalInt configContractSizeLimit,
       final OptionalInt configStackSizeLimit,
       final boolean enableRevertReason) {
+    checkArgument(chainId.isPresent(), "Istanbul requires the use of chainId");
```
mbaxter: If you trace back to where the `chainId` is set, it looks like it's only ever missing in test cases.  Might be worth going through and making the `chainId` argument a plain unwrapped `BigInteger`, in which case we wouldn't need this check.

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/vm/operations/ChainIdOperationTest.java
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  public ChainIdOperationTest(final String chainIdString, final int expectedGas) {
+    chainId = Bytes32.fromHexString(chainIdString);
+    this.expectedGas = expectedGas;
+    operation = new ChainIdOperation(new ConstantinopleGasCalculator(), chainId);
+  }
+
+  @Test
+  public void shouldReturnChanId() {
```
mbaxter:
```suggestion
  public void shouldReturnChainId() {
```

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/vm/operations/ChainIdOperationTest.java
```diff
@@ -0,0 +1,66 @@
...
+    final ArgumentCaptor<Bytes32> arg = ArgumentCaptor.forClass(Bytes32.class);
+    operation.execute(messageFrame);
+    Mockito.verify(messageFrame).pushStackItem(arg.capture());
```
mbaxter: You should probably also verify that `pushStackItem` is only called once.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/MainnetProtocolSpecs.java
shemnon: FixedDifficultyProtocolSchedule uses it too.  It's at least a 4 file change, larger than this PR, so that level of change would deserve its own PR.

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/vm/operations/ChainIdOperationTest.java
shemnon: Done

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/vm/operations/ChainIdOperationTest.java
shemnon: Done

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/MainnetProtocolSpecs.java
ajsutton: There's no requirement that a chain has a `chainId` - it may be omitted from the genesis config.  This is a slightly interesting hole in the ChainID opcode spec. Doesn't apply to any public chain but there may be chains out there it does affect.  So maybe ChainID opcode should return 0 when there is no chain ID?

PR #1691, opened 2019-07-14
add README, re PAN-2906

lazaridiscom:
Add an index-file (README) to the `development` folder, see PAN-2906.

(Dev related documents should be moved herein, and indexed in the README)

PR #1701, opened 2019-07-15
PAN-2906: move dev related files to development folder

lazaridiscom:
* delete folder 'docs/community'

* correct links

## Rationale

* the files within "community" are "development" related files.
* both folders ("community" and "development") are not published on the docu site (so should be anyways removed from "docs").

## Next Steps

* move "development" folder out of the docs-site sources (e.g. into root, possibly rename to "dev")

More info: 

https://pegasys1.atlassian.net/browse/PAN-2906


ajsutton:
Thanks but as mentioned on your previous PR #1691 and PAN-2906 this is not a change we're going to accept.  These files are typically found in the root directory and our README already points to them so they're easy to find.

lazaridiscom:
>  These files are typically found in the root directory and our README already points to them so they're easy to find.

@ajsutton , did you even take the time to read the rationales? This is not about the root-files README, but about a redundant folder with the doc-site-sources ("community").

PR #1704, opened 2019-07-15
[PAN-2904] EIP-1108 - Reprice alt_bn128

shemnon:
## PR description

Add configurability to how the ECADD, ECMUL, and Pairing check precompiled
contracts calculate their gas price, and expose static methods for Byzantium
and Istanbul era prices.


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/precompiles/AltBN128PairingPrecompiledContract.java
```diff
@@ -41,14 +40,27 @@
       BytesValue.fromHexString(
           "0x0000000000000000000000000000000000000000000000000000000000000001");
 
-  public AltBN128PairingPrecompiledContract(final GasCalculator gasCalculator) {
-    super("AltBN128Pairing", gasCalculator);
+  private final Gas pairingGasCost;
+  private final Gas baseGasCost;
+
+  private AltBN128PairingPrecompiledContract(final Gas pairingGasCost, final Gas baseGasCost) {
+    super("AltBN128Pairing", null);
```
mbaxter: It looks like you're using a new model for modifying the gas values by hardfork.  Other precompiles pull versioned gas values from the `GasCalculator`.  Any reason why that existing pattern won't work here?

shemnon: It becomes a very high touch solution.  In this case I would have to introduce two more interface methods in a separate file.  This will make it easier to implement EIPs that are laser focused on one particular set of opcode pricing without a large amount of interaction.

It also makes it difficult to split our version based logic cleanly.  In the future gas prices may diverge based on Account Version of the contract, and they can vary from call frame to call frame.  By putting it in the contract it reduces the logic since we have a different contract registries per version and per fork.  While right now it would be a simple if check we may wind up having many of those in separate methods in a file separated from the objects representing the contract or operation.

This will also make it easier to stand up "Fork+EIPX" tests like Geth and parity have been talking about doing, when the operations and contracts become more independent of each other instead of needing to work with a shared gas calculator class.

mbaxter: I'm not a big fan of the `GasCalculator` design as it is for the reasons you mention - it spreads out op-related data and logic in this kind of messy way.  But, it seems that if we want to change that pattern, we should have a plan to refactor.  Rather than introducing 2 patterns at once.

mbaxter: Discussed offline - we'll keep this new pattern for now, and follow-up with a ticket to do a more comprehensive refactor.

shemnon: PAN-2911 is the ticket.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/precompiles/AltBN128MulPrecompiledContract.java
```diff
@@ -30,13 +30,24 @@
...
+    super("AltBN128Mul", gasCalculator);
+    this.gasCost = gasCost;
+  }
+
+  public static AltBN128MulPrecompiledContract byzantium(final GasCalculator gasCalculator) {
+    return new AltBN128MulPrecompiledContract(gasCalculator, Gas.of(40_000));
+  }
+
+  public static AltBN128MulPrecompiledContract istanbul(final GasCalculator gasCalculator) {
+    return new AltBN128MulPrecompiledContract(gasCalculator, Gas.of(8_000));
```
holiman: Shouldn't this be `6000` gas ?

shemnon: Yep.  I missed it on review because our schedule was split across 3 files. - https://github.com/PegaSysEng/pantheon/pull/1874

PR #1706, opened 2019-07-15
[PAN-2905] EIP-1706 - Disable SSTORE with gasleft lt call stipend 

shemnon:
## PR description

Add minimum remaining gas check to exceptional halt reasons.  Configure it
to zero pre-istanbul and 2300 for istanbul.

### DIFF
--- a/ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/vm/operations/SStoreOperation.java
+++ b/ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/vm/operations/SStoreOperation.java
@@ -27,8 +27,14 @@ import java.util.Optional;
 
 public class SStoreOperation extends AbstractOperation {
 
-  public SStoreOperation(final GasCalculator gasCalculator) {
+  public static final Gas FRONTIER_MINIMUM = Gas.ZERO;
+  public static final Gas EIP_1706_MINIMUM = Gas.of(2300);
+
+  private final Gas minumumGasRemaining;
+
+  public SStoreOperation(final GasCalculator gasCalculator, final Gas minumumGasRemaining) {
     super(0x55, "SSTORE", 2, 0, false, 1, gasCalculator);
+    this.minumumGasRemaining = minumumGasRemaining;
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/vm/operations/SStoreOperation.java
```diff
@@ -27,8 +27,14 @@
 
 public class SStoreOperation extends AbstractOperation {
 
-  public SStoreOperation(final GasCalculator gasCalculator) {
+  public static final Gas FRONTIER_MINIMUM = Gas.ZERO;
+  public static final Gas EIP_1706_MINIMUM = Gas.of(2300);
```
mbaxter: If we follow the existing pattern, I think these gas values belong on the `GasCalculator`?

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/vm/operations/SStoreOperation.java
```diff
@@ -27,8 +27,14 @@
...
+
+  private final Gas minumumGasRemaining;
+
+  public SStoreOperation(final GasCalculator gasCalculator, final Gas minumumGasRemaining) {
```
mbaxter: (discussion) Since this new minimum gas check is backwards-compatible, should we just always apply it?  Rather than having separate minima for frontier vs istanbul?

shemnon: It is not backwards compatible.  Over 100 reference tests have broken so far in my test run and the build is not done yet.

shemnon: 169 to be exact.  Across a broad set of tests.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/vm/operations/SStoreOperation.java
shemnon: see https://github.com/PegaSysEng/pantheon/pull/1704#discussion_r303614431

PR #1707, opened 2019-07-15
[PAN-2910] Eea get private transaction fix

iikirilov:
## PR description
The response of the ee_getPrivateTransaction endpoint should return numeric values encoded in hex and enclave public keys encoded in base64. It should also return the private transaction with the privateFrom/privateFor or privateFrom/privacyGroupId combination depending on how the transaction was sent.

This PR also extends to fix the way transactions are sent with support for the following combinations:
* privateFrom and privateFor
* privateFrom and privacyGroupId
* privacyGroupId without privateFrom - the privateFrom field is populated from the enclave key used in the command line arguments to start up Pantheon.

This PR also extends from #1667 to handle PMT payload consistently
* the privacy marker transaction payload is required in base64 for orion receive requests

## Fixed Issue(s)
[PAN-2910]

[PAN-2910]: https://pegasys1.atlassian.net/browse/PAN-2910


ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/privacy/PrivateTransactionTest.java
```diff
@@ -39,58 +39,53 @@
...
-          + "a0ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"
-          + "ffff801ba048b55bfa915ac795c431978d8a6a992b628d557da5ff759b307d"
-          + "495a36649353a01fffd310ac743f371de3b9f7f9cb56c0b28ad43601b4ab94"
-          + "9f53faa07bd2c804ac41316156744d784c4355486d425648586f5a7a7a4267"
-          + "5062572f776a3561784470573958386c393153476f3df85aac41316156744d"
-          + "784c4355486d425648586f5a7a7a42675062572f776a356178447057395838"
-          + "6c393153476f3dac4b6f32625671442b6e4e6c4e594c35454537793349644f"
-          + "6e766966746a69697a706a52742b4854754642733d8a726573747269637465"
-          + "64";
+      "0xf8ef800182520894095e7baea6a6c7c4c2dfeb977efac326af552d87a0f"
```
iikirilov: Test vectors have changed here because enclave keys encoded correctly as 32 byte hex rather than 44 byte UTF_8 representation of base64

PR #1710, opened 2019-07-16
[PIE-1784] Select the pivot block from a minimal peer set

mbaxter:
## PR description
Wait for a minimum number of height estimates to be received before choosing a pivot block.  This is intended to improve our pivot block selection, making a selection of a low pivot block number less likely.


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/fastsync/FastSyncActions.java
```diff
@@ -80,14 +80,8 @@ public FastSyncActions(
...
     }
 
-    final WaitForPeersTask waitForPeersTask =
-        WaitForPeersTask.create(
-            ethContext, syncConfig.getFastSyncMinimumPeerCount(), metricsSystem);
-
     LOG.debug("Waiting for at least {} peers.", syncConfig.getFastSyncMinimumPeerCount());
-    return ethContext
-        .getScheduler()
-        .scheduleServiceTask(waitForPeersTask)
```
mbaxter: @ajsutton - do you know if there was a reason we were scheduling the `waitForPeersTask` asynchronously?  I cut this part, which should be fine as far as I can tell.

ajsutton: I'd like to say there was some sequence of refactoring that led to this and a final step just got missed, but most likely I was just an idiot when I wrote it initially. :)

mbaxter: 🤣

PR #1723, opened 2019-07-19
[PAN-2927]  Add documentation for getSignerMetrics

matkt:
## PR description

Add documentation for the new clique_getSignerMetrics and ibft_getSignerMetrics JSON-RPC methods.


docs/Reference/Pantheon-API-Methods.md
```diff
@@ -3451,6 +3505,60 @@ Proposes [adding or removing a validator](../Consensus-Protocols/IBFT.md#adding-
...
+- The block number of the last block proposed by each validator (if any within the given range).
+- All validators present in the last block of the range even if they didn’t propose a block.
+
+**Parameters**
+
+`fromBlockNumber` - Integer representing a block number or the string tag `earliest` as described in [Block Parameter](../Pantheon-API/Using-JSON-RPC-API.md#block-parameter). 
+
+`toBlockNumber` - Integer representing a block number or one of the string tags `latest` or `pending`, as described in [Block Parameter](../Pantheon-API/Using-JSON-RPC-API.md#block-parameter). (optional and default to `latest`)
+
+If no parameters are specified, only the last 100 blocks will be processed.
```
MadelineMurray: When I tried using this method with no parameters: 
`{"jsonrpc":"2.0","method":"ibft_getSignerMetrics","params":[], "id":1}` 

I got the following error: 
```
2019-07-20 14:37:05.183+10:00 | vert.x-worker-thread-5 | ERROR | JsonRpcHttpService | Error processing JSON-RPC request
java.lang.IllegalArgumentException: Argument must be positive, got -95
	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:203) ~[guava-28.0-jre.jar:?]
	at tech.pegasys.pantheon.util.uint.UInt256Bytes.of(UInt256Bytes.java:365) ~[pantheon-util-1.2.1-SNAPSHOT.jar:1.2.1-dev-40a939d3]
	at tech.pegasys.pantheon.ethereum.storage.keyvalue.KeyValueStoragePrefixedKeyBlockchainStorage.getBlockHash(KeyValueStoragePrefixedKeyBlockchainStorage.java:92) ~[pantheon-core-1.2.1-SNAPSHOT.jar:1.2.1-dev-40a939d3]
	at tech.pegasys.pantheon.ethereum.chain.DefaultMutableBlockchain.getBlockHeader(DefaultMutableBlockchain.java:137) ~[pantheon-core-1.2.1-SNAPSHOT.jar:1.2.1-dev-40a939d3]
	at tech.pegasys.pantheon.ethereum.jsonrpc.internal.queries.BlockchainQueries.getBlockHeaderByNumber(BlockchainQueries.java:345) ~[pantheon-json-rpc-1.2.1-SNAPSHOT.jar:1.2.1-dev-40a939d3]
...
``` 

Does the method itself need fixing (ie, it should accept no parameters) or does the doc need updating (ie, the parameters are required)?  I could use the method with a single parameter but not sure what the intended behaviour is with one parameter. 

Thanks for taking the time to contribute doc again!

matkt: indeed looking at the trace. I see that the method is trying to recover the last 100 blocks. while your test network only has 5. I will fix it so that it takes 100 or less depending on the number of blocks available. with a single parameter the second parameter takes the default value which is latest. I will explain this use case

matkt: I resolved the issue https://github.com/PegaSysEng/pantheon/pull/1725 and I updated the documentation. Could you say me if it's clear enough for you?

MadelineMurray: Thanks @matkt.  When #1725 is merged I'll try it out again and re-review the doc :-)

matkt:
> With ibft_getSignerMetrics, I'm getting a proposed block from address 0x000000000000 in the results which I'm assuming shouldn't be there?
> 
> ```
> {
>     "jsonrpc": "2.0",
>     "id": 1,
>     "result": [
>         {
>             "address": "0x120f5d5df0fbe9f85b566c3cf7ed2749f496d49b",
>             "proposedBlockCount": "0x3",
>             "lastProposedBlockNumber": "0x3"
>         },
>         {
>             "address": "0x0000000000000000000000000000000000000000",
>             "proposedBlockCount": "0x1",
>             "lastProposedBlockNumber": "0x0"
>         }
>     ]
> }
> ```

for me it is the address  0x000000000000 which is the proposer of block 0. I do not see any particular error. I will inquire to check

ajsutton:
This is an interesting little corner case - technically the genesis block doesn't have a proposer, it just always existed since the dawn of the chain.  That gets very confusing though so we tend to treat it as 0x0... is the proposer which is why it's showing up here.  I'm guessing your chain is less than 200 blocks long so the genesis block is being included in the report.

I think this result is pretty reasonable - the only other thing we could do is exclude the genesis block from this report entirely.  That would also be reasonable but have the odd corner case of reporting nothing if you ask for only the genesis block or if you haven't yet sync'd any blocks.

Maybe just add a note to explain that the genesis block's proposer is reported as 0x0...?

matkt:
> This is an interesting little corner case - technically the genesis block doesn't have a proposer, it just always existed since the dawn of the chain. That gets very confusing though so we tend to treat it as 0x0... is the proposer which is why it's showing up here. I'm guessing your chain is less than 200 blocks long so the genesis block is being included in the report.
> 
> I think this result is pretty reasonable - the only other thing we could do is exclude the genesis block from this report entirely. That would also be reasonable but have the odd corner case of reporting nothing if you ask for only the genesis block or if you haven't yet sync'd any blocks.
> 
> Maybe just add a note to explain that the genesis block's proposer is reported as 0x0...?
I think it is a good way to fix the problem. @MadelineMurray

matkt:
> With ibft_getSignerMetrics, I'm getting a proposed block from address 0x000000000000 in the results which I'm assuming shouldn't be there?
> 
> ```
> {
>     "jsonrpc": "2.0",
>     "id": 1,
>     "result": [
>         {
>             "address": "0x120f5d5df0fbe9f85b566c3cf7ed2749f496d49b",
>             "proposedBlockCount": "0x3",
>             "lastProposedBlockNumber": "0x3"
>         },
>         {
>             "address": "0x0000000000000000000000000000000000000000",
>             "proposedBlockCount": "0x1",
>             "lastProposedBlockNumber": "0x0"
>         }
>     ]
> }
> ```

I just added a note for the genesis block

PR #1734, opened 2019-07-22
Added CPU requirements

MadelineMurray:
Added CPU requirements


docs/Installation/System-Requirements.md
```diff
@@ -0,0 +1,31 @@
...
+
+## RAM 
+
+When connected to a public Ethereum network, 8GB of RAM is required.
+
+When connected to a private network, 4GB of RAM is required. 
+
+## Disk Space 
+
+Syncing to the Ethereum mainnet requires 1.5 TB to 2 TB for a full sync.
```
ajsutton: A full sync is now at least 2.5TB.  I'd say 3TB because the required size keeps growing with every added block.

docs/Installation/System-Requirements.md
```diff
@@ -0,0 +1,31 @@
...
+The system requirements for Pantheon depend on a number of factors: 
+
+* Network - are you connecting to a public Ethereum network (eg, mainnet) or a private network 
+* Network size if connecting to a private network 
+* Number and complexity of queries being performed by node 
+
+## CPU Requirements for Syncing to Ethereum Mainnet
+
+Syncing to the Ethereum mainnet using a t3.xlarge takes about 15 hours.  The sync to mainnet can be completed with 
+a t3.large but takes longer.
```
ajsutton: I wouldn't include an expected fast sync time - it's hugely variable depending on which nodes you find and if the pivot block has to be updated.

The required instance type is far more complex than this. An `i3.xlarge` can sync MainNet reliably but has very little extra capacity to service JSON-RPC queries. An `i3.2xlarge` is faster to sync and has more headroom.  A `t3.2xlarge` can also sync main net but is slower to sync than either of the `i3` instances.  We haven't got any meaningful info on other `t3` category instance sizes.

docs/Installation/System-Requirements.md
```diff
@@ -0,0 +1,31 @@
...
+* Network size if connecting to a private network 
+* Number and complexity of queries being performed by node 
+
+## CPU Requirements for Syncing to Ethereum Mainnet
+
+Syncing to the Ethereum mainnet using a t3.xlarge takes about 15 hours.  The sync to mainnet can be completed with 
+a t3.large but takes longer.  
+
+!!! tip
+    Once the initial sync has been completed, the instance size can be reduced.
```
ajsutton: I wouldn't include this.  It depends on the type of JSON-RPC queries you need to perform and the category of instance you're running.  For example an `i3.large` can fast sync MainNet reliably but is often unable to keep up with the number of transactions flowing around the network (ie you need to increase the config after sync completes).

docs/Installation/System-Requirements.md
```diff
@@ -0,0 +1,31 @@
...
+
+Syncing to the Ethereum mainnet using a t3.xlarge takes about 15 hours.  The sync to mainnet can be completed with 
+a t3.large but takes longer.  
+
+!!! tip
+    Once the initial sync has been completed, the instance size can be reduced. 
+
+## RAM 
+
+When connected to a public Ethereum network, 8GB of RAM is required.
```
ajsutton: There are big public networks and small ones.  MainNet needs 8GB of RAM, Görli would be fine on 4GB or maybe even less.

docs/Installation/System-Requirements.md
MadelineMurray: Fixed

docs/Installation/System-Requirements.md
MadelineMurray: Removed with the reworking

docs/Installation/System-Requirements.md
MadelineMurray: Updated to be specific to mainnet

docs/Installation/System-Requirements.md
```diff
@@ -0,0 +1,29 @@
...
+
+The system requirements for Pantheon depend on a number of factors: 
+
+* Network - are you connecting to a public Ethereum network (eg, mainnet) or a private network 
+* Network size if connecting to a private network 
+* Number and complexity of queries being performed by node 
+
+## Determining System Requirements  
+
+To determine system requirements, monitor CPU and disk space requirements using [Prometheus](https://docs.pantheon.pegasys.tech/en/stable/Monitoring/Monitoring-Performance/#monitor-node-performance-using-prometheus).
```
joshuafernandes: :+1:

docs/Installation/System-Requirements.md
```diff
@@ -0,0 +1,29 @@
...
+A sample [Grafana dashboard](https://grafana.com/grafana/dashboards/10273) is provided for Pantheon. 
+
+!!! tip
+    CPU requirements are highest when syncing to the network and reduce once the node is synchronized to the chain head. 
+
+## RAM 
+
+When connected to the Ethereum mainnet, 8GB of RAM is required.
+
+When connected to a private network, 4GB of RAM is required.
```
ajsutton: I'd phrase this more along the lines of "Pantheon requires 4GB of RAM.  For Ethereum MainNet, a minimum of 8GB of RAM is required".  It's not whether the network is public or private, it's how big its world state is.

docs/Installation/System-Requirements.md
```diff
@@ -0,0 +1,29 @@
...
+* Network size if connecting to a private network 
+* Number and complexity of queries being performed by node 
+
+## Determining System Requirements  
+
+To determine system requirements, monitor CPU and disk space requirements using [Prometheus](https://docs.pantheon.pegasys.tech/en/stable/Monitoring/Monitoring-Performance/#monitor-node-performance-using-prometheus). 
+A sample [Grafana dashboard](https://grafana.com/grafana/dashboards/10273) is provided for Pantheon. 
+
+!!! tip
+    CPU requirements are highest when syncing to the network and reduce once the node is synchronized to the chain head.
```
ajsutton: It's not guaranteed that less CPU will be used when synchronised.

```suggestion
    CPU requirements are highest when syncing to the network and will typically reduce once the node is synchronized to the chain head. 
```

docs/Installation/System-Requirements.md
```diff
@@ -0,0 +1,29 @@
...
+
+## RAM 
+
+When connected to the Ethereum mainnet, 8GB of RAM is required.
+
+When connected to a private network, 4GB of RAM is required. 
+
+## Disk Space 
+
+Syncing to the Ethereum mainnet requires 3TB for a full sync.
```
joshuafernandes: Optionally add that Fast Sync takes about 1.5TB?

docs/Installation/System-Requirements.md
```diff
@@ -0,0 +1,29 @@
+title: System Requirements
+description: System requirements to sync and run Pantheon 
+<!--- END of page meta data -->
+
+# System Requirements 
+
+The system requirements for Pantheon depend on a number of factors: 
+
+* Network - are you connecting to a public Ethereum network (eg, mainnet) or a private network 
+* Network size if connecting to a private network
```
ajsutton: The network you connect to doesn't really matter, it's how big the world state is for that network, how many transactions are submitted to the network and how high the block gas limit is.

docs/Installation/System-Requirements.md
```diff
@@ -0,0 +1,29 @@
...
+description: System requirements to sync and run Pantheon 
+<!--- END of page meta data -->
+
+# System Requirements 
+
+The system requirements for Pantheon depend on a number of factors: 
+
+* Network - are you connecting to a public Ethereum network (eg, mainnet) or a private network 
+* Network size if connecting to a private network 
+* Number and complexity of queries being performed by node
```
ajsutton: `performed by node` doesn't make a lot of sense to me. 
```suggestion
* Number and complexity of JSON-RPC, PubSub or GraphQL queries being handled by the node
```

docs/Installation/System-Requirements.md
ajsutton: Fast sync takes about 200Gb when it first completes and then grows from there as blocks are added.  We partly allocate larger disk on our test nodes because we then get a reasonable amount of IOps.

docs/Installation/System-Requirements.md
MadelineMurray: Fixed

docs/Installation/System-Requirements.md
MadelineMurray: Fixed

docs/Installation/System-Requirements.md
MadelineMurray: Fixed

docs/Installation/System-Requirements.md
MadelineMurray: Fixed

PR #1738, opened 2019-07-23
[PAN-2952] Add preimage cache to DebuggableWorldState

shemnon:
## PR description
  
As part of getting the storage range we will need to be able to reverse the
hashes. Realistically this means storing the preimages as we calculate them.


shemnon:
#1780 and  #1772 did it better.

PR #1739, opened 2019-07-23
[PAN-2756] EIP-2028 - Reduce intrinsic gas cost

shemnon:
## PR description

Reduce the non-zero byte of intrinsic gas cose from 68 to 16.
Keep all other values the same.


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/IstanbulGasCalculator.java
```diff
@@ -0,0 +1,48 @@
...
+
+  @Override
+  public Gas transactionIntrinsicGasCost(final Transaction transaction) {
+    final BytesValue payload = transaction.getPayload();
+    int zeros = 0;
+    for (int i = 0; i < payload.size(); i++) {
+      if (payload.get(i) == 0) {
+        ++zeros;
+      }
+    }
```
mbaxter: (optional) would be nice to dedupe this zero counting logic

ethereum/core/src/test-support/java/tech/pegasys/pantheon/ethereum/core/ExecutionContextTestFixture.java
```diff
@@ -110,7 +110,7 @@ public ExecutionContextTestFixture build() {
       if (protocolSchedule == null) {
         protocolSchedule =
             new ProtocolScheduleBuilder<>(
-                    new StubGenesisConfigOptions().istanbulBlock(0),
+                    new StubGenesisConfigOptions().constantinopleFixBlock(0),
```
mbaxter: what's this change for?

shemnon: The prices changed and the old unit tests were based on Petersburg prices.  I had speculatively changed this test to Istanbul when I added Istanbul support now I'm moving it back.  Once Istanbul is on mainnet we should revisit these tests and assertions.

PR #1745, opened 2019-07-24
PAN-2723: Added static nodes acceptance test

lucassaldanha:


acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/StaticNodesUtils.java
```diff
@@ -0,0 +1,45 @@
...
+public class StaticNodesUtils {
+
+  public static Path createStaticNodesFile(final Path directory, final List<String> staticNodes) {
+    try {
+      final Path tempFile = Files.createTempFile(directory, "", "");
+      tempFile.toFile().deleteOnExit();
+
+      final Path staticNodesFile = tempFile.getParent().resolve("static-nodes.json");
+      Files.move(tempFile, staticNodesFile);
+      staticNodesFile.toFile().deleteOnExit();
```
usmansaleem: can we override the file name "static-nodes.json" later on at Pantheon CLI? if we can, then we can avoid the `Files.move`.

usmansaleem: The other option is to use Files.createTempDirectory(directory, ...) and then create static-nodes.json in it ...

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/PantheonNode.java
```diff
@@ -159,6 +160,7 @@ public PantheonNode(
           }
         });
     this.extraCLIOptions = extraCLIOptions;
+    this.staticNodes = staticNodes;
```
usmansaleem: should we check for null staticNodes and assign a new one?

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/ThreadPantheonNodeRunner.java
```diff
@@ -154,6 +154,10 @@ public void startNode(final PantheonNode node) {
             .metricsConfiguration(node.metricsConfiguration())
             .p2pEnabled(node.isP2pEnabled())
             .graphQLConfiguration(GraphQLConfiguration.createDefault())
+            .staticNodes(
+                node.getStaticNodes().stream()
```
usmansaleem: this has a potential of NPE unless we make sure getStaticNodes will never return null.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/configuration/PantheonFactoryConfigurationBuilder.java
```diff
@@ -193,6 +194,11 @@ public PantheonFactoryConfigurationBuilder revertReasonEnabled() {
     return this;
   }
 
+  public PantheonFactoryConfigurationBuilder staticNodes(final List<String> staticNodes) {
+    this.staticNodes = staticNodes;
```
usmansaleem: null check?

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/StaticNodesUtils.java
lucassaldanha: The name is hardcoded to `static-nodes.json`. The reason I'm not creating a temp directory is 'cause we already have the node home directory to use (that's where the static-nodes.json) file should be.

I'll take a look and see if I can optimize this in my following PR.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/ThreadPantheonNodeRunner.java
lucassaldanha: We always initialize it in `PantheonFactoryConfigurationBuilder` as an empty list. There is a lot of indirection to get there but we should be safe for now.

PR #1746, opened 2019-07-24
RevertReason changed to BytesValue

rain-on:
The revert reason returned from the EVM has been changed to a bytes value to
better represent the specification (of it being an ABI encoded data block).

This has meant that the TransactionReceipt returned from
ethGetTransactionReceipt now provides a hex-encoded string (rather than UTF-8).


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/TransactionReceipt.java
```diff
@@ -182,15 +181,15 @@ public static TransactionReceipt readFrom(
...
+      final Optional<BytesValue> revertReason;
       if (input.isEndOfCurrentList()) {
         revertReason = Optional.empty();
       } else {
         if (!revertReasonAllowed) {
           throw new RLPException("Unexpected value at end of TransactionReceipt");
         }
         final byte[] bytes = input.readBytesValue().getArrayUnsafe();
-        revertReason = Optional.of(new String(bytes, StandardCharsets.UTF_8));
+        revertReason = Optional.of(BytesValue.wrap(bytes));
```
ajsutton: We don't need to re-wrap the byte array here, just use the `BytesValue` returned from `input`.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/vm/DebugOperationTracer.java
```diff
@@ -68,7 +69,12 @@ public void traceExecution(
               stack,
               memory,
               storage,
-              frame.getRevertReason().orElse(null)));
+              frame
+                  .getRevertReason()
+                  .map(
+                      revertReason ->
+                          new String(revertReason.extractArray(), StandardCharsets.UTF_8))
+                  .orElse(null)));
```
ajsutton: This is still wrong, the revert reason isn't a string - we need to output the result as a hex array in the trace output.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/TransactionReceipt.java
rain-on: Done.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/results/StructLog.java
```diff
@@ -36,7 +37,7 @@
   private final int pc;
   private final String[] stack;
   private final Object storage;
-  private final String reason;
+  private final BytesValue reason;
```
ajsutton: I think this should be a String here, but calculated from the `BytesValue` with `BytesValue.toUnprefixedString`.  That would then match how the stack and return values are all formatted.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/results/TransactionReceiptResult.java
```diff
@@ -81,7 +81,11 @@ public TransactionReceiptResult(final TransactionReceiptWithMetadata receiptWith
     this.to = receiptWithMetadata.getTransaction().getTo().map(BytesValue::toString).orElse(null);
     this.transactionHash = receiptWithMetadata.getTransaction().hash().toString();
     this.transactionIndex = Quantity.create(receiptWithMetadata.getTransactionIndex());
-    this.revertReason = receipt.getRevertReason().orElse(null);
+    this.revertReason =
+        receipt
+            .getRevertReason()
+            .map(reason -> reason.isEmpty() ? "" : reason.toString())
```
ajsutton: I think this should just be:
```suggestion
            .map(BytesValue::toUnprefixedString)
```
as we want to output it as a hex value in the same way as `StructLog` does above.

rain-on: that means you get "revertReason": "0x" ... which ... seemed odd to me - but if thats the preferred solution ...

PR #1747, opened 2019-07-24
[PAN-2950] Use java.time.Clock instead of System.currentTimeMillis()

shemnon:
## PR description

To allow us to reset the timestamp in the blockchain we need to pass a Clock to
affected APIs and use that instead of the static method
System.currentTimeMillis().  The most consistent way to do this that will ensure
that the API does not sneak back in is to ban the method via ErrorProne.

TestClock was enhanced with a `fixedNow()` method that returns a fixed time that
is close to the current real time, as opposed to `fixed()` that returns a time
sometime in 1970.  All places where this was relevant in tests received the
`fixedNow()` clock


errorprone-checks/src/main/java/tech/pegasys/errorpronechecks/BannedMethod.java
```diff
@@ -34,12 +34,15 @@
...
     summary = "Some methods should not be used, make sure that doesn't happen.",
-    severity = WARNING)
+    severity = WARNING,
+    linkType = LinkType.NONE)
 public class BannedMethod extends BugChecker implements MethodInvocationTreeMatcher {
 
   private static final ImmutableMap<Matcher<ExpressionTree>, String> BANNED_METHOD_LIST =
       ImmutableMap.of(
-          allOf(staticMethod().onClass("com.google.common.base.Objects").withAnyName()),
+          staticMethod().onClass("java.lang.System").named("currentTimeMillis"),
```
shemnon: Drop the ban hammer here.

Ban Hammer

errorprone-checks/src/main/java/tech/pegasys/errorpronechecks/BannedMethod.java
```diff
@@ -34,12 +34,15 @@
...
+          "Do not use System.currentTimeMillis(), use a java.time.Clock passed into a constructor or as a static method parameter.",
+          staticMethod().onClass("com.google.common.base.Objects").withAnyName(),
```
mbaxter: Suggest adding a test for this new rule to `BannedMethodTest`

ethereum/p2p/src/main/java/tech/pegasys/pantheon/ethereum/p2p/rlpx/connections/netty/NettyConnectionInitializer.java
```diff
@@ -165,7 +169,7 @@ public void subscribeIncomingConnect(final ConnectCallback callback) {
     final CompletableFuture<PeerConnection> connectionFuture = new CompletableFuture<>();
 
     if (peer instanceof DiscoveryPeer) {
-      ((DiscoveryPeer) peer).setLastAttemptedConnection(System.currentTimeMillis());
+      ((DiscoveryPeer) peer).setLastAttemptedConnection(clock.millis());
```
mbaxter: This line duplicates logic in `RlpxAgent`, so we can cut it here and cut the `Clock` dependency as well.

pantheon/src/main/java/tech/pegasys/pantheon/cli/error/PantheonExceptionHandler.java
```diff
@@ -32,7 +32,7 @@ public PantheonExceptionHandler(final Supplier<Level> levelSupplier) {
   public List<Object> handleParseException(
       final CommandLine.ParameterException ex, final String[] args) {
     final Level logLevel = levelSupplier.get();
-    if (logLevel != null && Level.DEBUG.isMoreSpecificThan(logLevel)) {
+    if (ex.getMessage() == null || (logLevel != null && Level.DEBUG.isMoreSpecificThan(logLevel))) {
```
mbaxter: Just checking you meant to include this change?

errorprone-checks/src/main/java/tech/pegasys/errorpronechecks/BannedMethod.java
shemnon: It's already present.  [here](https://github.com/PegaSysEng/pantheon/pull/1747/files#diff-2614b8be9cbaff7fa1e2cb660f4bad9aR28) and [here](https://github.com/PegaSysEng/pantheon/pull/1747/files#diff-47d8ca838f0b31e45606b8d97a895f51R31)

pantheon/src/main/java/tech/pegasys/pantheon/cli/error/PantheonExceptionHandler.java
shemnon: it was a debugging aid that I forgot to clean up, so I'll cut it.  May be worthy of its own PR.

PR #1751, opened 2019-07-24
Reduce number of builds retained for master

ajsutton:
## PR description
We were keeping 90 days of builds for master which is using a lot of disk space and Jenkins keeps running out of disk space.  Reduce it to 30 days and reduce artefact retention to 7 days.


joshuafernandes:
LGTM

PR #1755, opened 2019-07-25
[PAN-2957] Add Java 11+ as a prerequisite for installing Pantheon using Homebrew.

bgravenorst:
# PR description
Add a prerequisite to the documentation informing the user that Java 11+ is required to install Pantheon using Homebrew.


docs/Installation/Install-Binaries.md
```diff
@@ -8,46 +8,50 @@ description: Install Pantheon from binary distribution
 ### Prerequisites
 
 * [Homebrew](https://brew.sh/)
+* [Java JDK](http://www.oracle.com/technetwork/java/javase/downloads/index.html)
+
+!!!attention
+    Pantheon requires Java 11+ to compile; earlier versions are not supported.
```
ajsutton: I think this is misleading, it implies you need to manually install Java, but users are probably better off installing it via homebrew rather than being linked to the Oracle site.

ajsutton: The brew error message provides the easiest way to do this:

`brew cask install adoptopenjdk`

Not quite sure what the wording around this should be, but that's the command I'd recommend for the average user.  Not sure if it's worth pointing out there are other ways to install specific versions if the user wants. https://github.com/AdoptOpenJDK/homebrew-openjdk has instructions to allow users to select from a very confusing list of possible JDKs. :)

bgravenorst: Would it make more sense to remove the Oracle link, and update the admonition to, "Pantheon requires Java 11+ to compile; earlier versions are not supported. If the appropriate Java version is not installed, then Homebrew will provide you with the appropriate command to install Java. Alternatively, you can manually install the Java JDK. "?

ajsutton: Yeah I think something like that, but I would include the `brew cask install adoptopenjdk` command to make it easy for users instead of them having to fail to install pantheon to find the command they need.

bgravenorst: Thanks :). I updated the description to include the install command. After much deliberation I decided not to include the link to the homebrew-openjdk page. But I'm happy to add it if you have strong feelings about it.

ajsutton: I suspect that's a good choice.

PR #1772, opened 2019-07-26
[PAN-2952] Add storage key preimage tracking

mbaxter:
## PR description
This PR enables `debug_storageRangeAt` to display storage trie key preimages by adding a new storage class: `WorldStatePreimageStorage`.  This class is backed by a new type of `KeyValueStorage` (`LimitedInMemoryKeyValueStorage`) which caps the amount of data kept in-memory.  This new storage object is provided to `WorldStateArchive` which passes it to any `WorldState` instances created from the archive.  Accounts returned from `WorldState` instances then use this preimage storage to return `AccountStorageEntry`'s with optional `key` preimages from `storageEntriesFrom()`.


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/storage/keyvalue/RocksDbStorageProvider.java
```diff
@@ -31,38 +32,58 @@
 import org.apache.logging.log4j.Logger;
 
 public class RocksDbStorageProvider {
+  public static long DEFAULT_WORLD_STATE_PREIMAGE_CACHE_SIZE = 5_000L;
```
shemnon: Do we want to expose a `--X` option for this value?  Either as the default or the value that is wired into the world state storage?

mbaxter: Maybe?  I think we should be adding those options on an as-needed basis for the time being, so I'll leave it for a potential follow-up PR.

PR #1777, opened 2019-07-29
[PAN-2971] Remove enclave public key from parameter

iikirilov:
# PR description

Removes enclave public key from parameter list as it is used in the implementation in Orion to not send the request to itself.

## Fixed Issue(s)
[PAN-2971]

[PAN-2971]: https://pegasys1.atlassian.net/browse/PAN-2971


lucassaldanha:
> I'm not sure this is the change we want (ie, we want to confirm the request is coming from the Orion node that's part of the group being deleted and it should otherwise fail).

From Ivaylo:
The Orion node won't ever in any case be part of a group it cannot delete. The ID is only used not to send the request to itself. i.e. the request to delete is sent to the list of participants except for the deletor - because they are making the request.

PR #1778, opened 2019-07-29
[PAN-2982] Modifies PrivGetPrivateTransaction to take public tx hash

josh-richardson:
## PR description
eea_sendTransaction returns the public tx hash. 

priv_getPrivateTx takes the input field from the object returned by eth_getTransactionByHash  as a parameter.   So before calling priv_getPrivateTx, the user must call eth_getTransactionByHash specifying the public tx hash as input. 

Updating priv_getPrivateTransaction to take public tx hash as parameter means instead of the user performing 2 requests to get the private tx: eea_sendTransaction -> returns public tx hash -> eth_getTransactionByHash -> input value -> priv_getPrivateTx 

It would be 1: eea_sendTransaction -> returns public tx hash -> priv_getPrivateTx 

Acceptance criteria: 

priv_getPrivateTx takes the public tx hash as a parameter and returns the private tx


josh-richardson:
Seems to be stalling while pulling my branch?

PR #1780, opened 2019-07-29
[PAN-2953] Track world state account key preimages

mbaxter:
## PR description

When new accounts are created, save the mapping from addressHash to address in `WorldStatePreimageStorage`.  This allows us to iterate over account values in the world state trie, filling in the account address from the preimage storage.


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/WorldState.java
```diff
@@ -34,8 +36,10 @@
   /**
    * A stream of all the accounts in this world state.
    *
+   * @param startKeyHash The trie key at which to start iterating
+   * @param limit The maximum number of results to return
    * @return a stream of all the accounts (in no particular order) contained in the world state
    *     represented by the root hash of this object at the time of the call.
    */
-  Stream<Account> streamAccounts();
+  Stream<Account> streamAccounts(Bytes32 startKeyHash, int limit);
```
shemnon: Awesome, I had to fake this out on my end.

PR #1784, opened 2019-07-30
[PAN-2954] - Allow fixedDifficulty=1 to mean "all solutions"

shemnon:
## PR description

If you try and do a difficulty of 1 the UInt256 class used for comparison throws
an error because 2^256 is 33 bytes.  Instead use 2^256-1 which is 32 bytes.

### DIFF
--- a/ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/headervalidationrules/ProofOfWorkValidationRule.java
+++ b/ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/headervalidationrules/ProofOfWorkValidationRule.java
@@ -31,9 +31,9 @@ public final class ProofOfWorkValidationRule implements DetachedBlockHeaderValid
 
   private static final Logger LOG = LogManager.getLogger();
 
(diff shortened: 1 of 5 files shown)


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/headervalidationrules/ProofOfWorkValidationRule.java
```diff
@@ -47,8 +47,10 @@ public boolean validate(final BlockHeader header, final BlockHeader parent) {
     }
     final BigInteger difficulty =
         BytesValues.asUnsignedBigInteger(header.getDifficulty().getBytes());
-    final UInt256 target = UInt256.of(ETHHASH_TARGET_UPPER_BOUND.divide(difficulty));
-
+    final UInt256 target =
```
mbaxter: What do you think about creating a new `ethash` package somewhere (ethereum/core/.../mainnet/ethash??) to consolidate these types of utilities?

shemnon: Sounds like a good idea, probably as part of ProgPow work.  Opened as PAN-2998

PR #1789, opened 2019-07-30
[PAN-2996] Remove enclave public key from parameter

iikirilov:
## PR description

Enforce that the creator of a privacy group is the node. This simplifies the privacy group creation process and inhibits users to make mistakes when attempting to create a privacy group.

## Fixed Issue(s)
[PAN-2996]

[PAN-2996]: https://pegasys1.atlassian.net/browse/PAN-2996


iikirilov:
They never were. This PR makes that happen. 

N.B. The order of the parameters has changed such that the mandatory parameters come first.

PR #1795, opened 2019-07-30
[PAN-2989] EIP-1884 - Repricing for trie-size-dependent opcodes

shemnon:
## PR description

Add the new gas costs and new operation to the EVM for Istanbul.


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/vm/operations/SelfBalanceOperation.java
```diff
@@ -0,0 +1,40 @@
...
+
+  @Override
+  public Gas cost(final MessageFrame frame) {
+    return gasCalculator().getLowTierGasCost();
+  }
+
+  @Override
+  public void execute(final MessageFrame frame) {
+    final Address accountAddress = frame.getContractAddress();
+    final Account account = frame.getWorldState().get(accountAddress);
```
mbaxter: Do we want to cache the contract balance on `MessageFrame`?  Otherwise, the cheaper pricing isn't really justified.

shemnon: Yes, we want to.

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/vm/EnvironmentInformation.java
```diff
@@ -69,6 +71,7 @@
   @JsonCreator
   public EnvironmentInformation(
       @JsonProperty("address") final String account,
+      @JsonProperty("balance") final String balance,
```
mbaxter: Are there new reference tests that use this?  Or is this speculative?

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/vm/EnvironmentInformation.java
```diff
@@ -81,6 +84,7 @@ public EnvironmentInformation(
         code,
         0,
         account == null ? null : Address.fromHexString(account),
+        balance == null ? null : Wei.fromHexString(balance),
```
mbaxter: Should this default to `Wei.ZERO`?

shemnon:
Just one little NPE, what's the worst that can happen? ;)

fixed.

holiman:
I may be totally missing something here, since I'm not all too familiar with how pantheon works under the hood. But it looks to me like you are setting the balance prior to entering the callframe, and then keep it as a context variable for the duration of the call? 

I'm just wondering if this implementation works with e.g. 
`SELFBALANCE; CALL(0, 0,1,0,0,0,0); SELFBALANCE` (where the second invocation of `SELFBALANCE` should be one wei less than before ?

shemnon:
That optimization wasn't as easy as expected.  We will roll it back for now - https://github.com/PegaSysEng/pantheon/pull/1875

PR #1798, opened 2019-07-30
[PIE-1632] newHeads subscription emits events only for canonical blocks

lucassaldanha:
#PR description
* Updated `NewBlockHeadersSubscriptionService#onBlockAdded` logic to check if the block is in the canonical chain before sending a notification
* Added unit test to cover scenario
* Refactored some of the other unit tests to improve readability


ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/websocket/subscription/blockheaders/NewBlockHeadersSubscriptionService.java
```diff
@@ -36,21 +36,23 @@ public NewBlockHeadersSubscriptionService(
 
   @Override
   public void onBlockAdded(final BlockAddedEvent event, final Blockchain blockchain) {
-    subscriptionManager.notifySubscribersOnWorkerThread(
-        SubscriptionType.NEW_BLOCK_HEADERS,
-        NewBlockHeadersSubscription.class,
-        subscribers -> {
-          final Hash newBlockHash = event.getBlock().getHash();
+    if (event.isNewCanonicalHead()) {
```
usmansaleem: nitpick ... `if (!event.isNewCanonicalHead()) {
   return;
  }`

would cause less code change :)

PR #1806, opened 2019-07-31
exclude files that are not expected to be visible in the doc

NicolasMassart:
## PR description
exclude files that are not expected to be visible in the doc directly

it prevents some pages like developer/ and community/ pages that are not linked
 or not in the nav but we did not decided yet what to do with to appear in the doc
search.

Files like includes for test accounts also don't have any reason to be accessed
directly.

See https://github.com/apenwarr/mkdocs-exclude for this new feature doc.
## Fixed Issue(s)

no ticket for this doc quick fix


docs/requirements.txt
```diff
@@ -5,4 +5,5 @@ Markdown>=3.1
 markdown-fenced-code-tabs>=1.0
 markdown-include>=0.5
 MarkupSafe>=1.1
-mkdocs-markdownextradata-plugin==0.0.5
\ No newline at end of file
+mkdocs-markdownextradata-plugin==0.0.5
```
sgregglives:
```suggestion
mkdocs-markdownextradata-plugin>=0.0.5
```

PR #1809, opened 2019-07-31
[PIE-1809] Clean up genesis parsing

mbaxter:
## PR description
Remove `vertx` dependency from `tech.pegasys.pantheon.config` package.  Instead of using `vertx`'s json parsing utilities, just use the underlying `jackson` parsing library.


config/src/main/java/tech/pegasys/pantheon/config/JsonUtil.java
```diff
@@ -0,0 +1,149 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+public class JsonUtil {
+  public static OptionalLong getOptionalLong(final ObjectNode json, final String key) {
+    return getValue(json, key, JsonNode::asLong).map(OptionalLong::of).orElse(OptionalLong.empty());
+  }
+
+  public static <T> Optional<T> getValue(
```
shemnon: Have you considered Int, Long, String, and a boolean specific versions of getValue?  Having to pass the JsonNode::asFoo for each call feels a bit clunky.  We could then call the static method directly which should save a few cycles instead of passing an argument around and doing an indirect call.

mbaxter: Yeah, that's probably worth doing.  I'll try reworking this method.

mbaxter: Updated

pantheon/src/test/java/tech/pegasys/pantheon/cli/operator/OperatorSubCommandTest.java
```diff
@@ -152,15 +153,36 @@ public void privateKeyFileNameShouldBeEqualToOption() throws IOException {
...
-  @Test(expected = CommandLine.ExecutionException.class)
-  public void shouldFailIfDuplicateFiles() throws IOException {
-    runCmdAndCheckOutput(
-        cmd("--private-key-file-name", "dup.test", "--public-key-file-name", "dup.test"),
-        "/operator/config_generate_keys.json",
-        tmpOutputDirectoryPath,
-        "genesis.json",
-        true,
-        asList("key.pub", "priv.test"));
+  @Test
```
shemnon: Did this sneak in from another clangelist?

mbaxter: No, I made the validation stricter [here](https://github.com/PegaSysEng/pantheon/pull/1809/files#diff-0463d13bbf95f450907e60b2d512ddf8R187), so added a test for that, and reworked this pre-existing test to use `assertThatThrownBy` rather than the `@Test(expected = ...)` annotation.

This subcommand was using `vertx` libraries because it was interacting with `JsonGenesisConfigOptions`.

PR #1811, opened 2019-08-01
[PAN-3009] Fixes invalid JsonGetter, comment

josh-richardson:
## PR description
Fixes the following: 
The json getter on line 49 is mislabeled 
It should be changed from "privateFor" to "privacyGroupId"
Also, the comment above the class is not accurate


ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/results/privacy/PrivateTransactionGroupResult.java
```diff
@@ -36,7 +36,7 @@
 })
 /*
  The original deserialised private transaction sent via eea_sendRawTransaction
- This class is used if the original request was sent with privateFrom and privateFor
+ This class is used if the original request was sent with privacyGroupId
```
rain-on: It'd be good to update the comment in PrivateTransactionLegacyResult as well (same reason).

josh-richardson: This has now been addressed. Sorry for the delay.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/results/privacy/PrivateTransactionLegacyResult.java
```diff
@@ -39,7 +39,7 @@
 })
 /*
  The original deserialised private transaction sent via eea_sendRawTransaction
- This class is used if the original request was sent with privateFrom and privateFor
+ This class is used if the original request was sent with privacyGroupId
```
iikirilov: the comment here was correct

rain-on: that is true - what was I thinking?

PR #1813, opened 2019-08-01
[PAN-2954] Support for NoReward and NoProof seal engines

shemnon:
## PR description

Retesteth requires support for mining without block rewards and without PoW.
No PoW simply requires exposing the nonce generator and using a special
EthHashSolver.

No Reward in the tests is interpreted to mean Zero Reward, which presents
problems pre EIP158 with regard to zero balance accounts as the tests point the
coinbase to a non-existent account.  We need to add a sentinel NO_REWARD reward
value (in addition to existing support for a zero value which is a short
circuit) that will proceed with the block reward logic at zero value.
UInt256.MAX_VALUE was chosen as that sentinel as it is a value that would never
make sense otherwise, causing instant integer overflow in any real use case.

Retesteth will use Wei.ZERO or Wei.NO_REWARD based on the configured fork when
mining in no-reward mode.


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/Wei.java
```diff
@@ -23,6 +23,7 @@
 public final class Wei extends BaseUInt256Value<Wei> {
 
   public static final Wei ZERO = of(0);
+  public static final Wei NO_REWARD = of(UInt256.MAX_VALUE);
```
mbaxter: This feels like the wrong place for this constant - maybe `BlockProcessor` instead?

ethereum/blockcreation/src/main/java/tech/pegasys/pantheon/ethereum/blockcreation/AbstractBlockCreator.java
```diff
@@ -245,24 +245,25 @@ public boolean isCancelled() {
...
-      final Wei blockReward) {
+      final Wei blockRewardSpec) {
 
     // TODO(tmm): Added to make this work, should come from blockProcessor.
     final int MAX_GENERATION = 6;
-    if (blockReward.isZero()) {
+    if (blockRewardSpec.isZero()) {
       return true;
     }
+    final Wei blockReward = blockRewardSpec.equals(Wei.NO_REWARD) ? Wei.ZERO : blockRewardSpec;
```
mbaxter: Instead of having this special `Wei.NO_REWARD` value - what about just adding a flag like `forceUpdateBeneficiaryAccount(boolean shouldForceUpdate)` or something like that?  I think that would make the intentions behind this logic clearer.  Then you can return early on line 263 iff `blockRewards.isZero() && !forceUpdate`.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/MainnetBlockProcessor.java
```diff
@@ -158,6 +158,8 @@ private boolean rewardCoinbase(
     if (blockReward.isZero()) {
```
mbaxter: would be nice if we could dedupe this method :D, but probably out of scope for this change

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/MainnetBlockProcessor.java
```diff
@@ -158,6 +158,8 @@ private boolean rewardCoinbase(
...
       return true;
     }
+    // This reference equality check is deliberate.
+    final Wei blockReward = this.blockReward.equals(Wei.NO_REWARD) ? Wei.ZERO : this.blockReward;
```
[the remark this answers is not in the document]
shemnon: Yes, stray comment.  Will Fix.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/Wei.java
shemnon: I can move it there.  It also needs to be visible to AbstractBlockCreator and MainnetBlockProcessor, but it looks like the needed dependencies are in place.

ethereum/blockcreation/src/main/java/tech/pegasys/pantheon/ethereum/blockcreation/AbstractBlockCreator.java
shemnon: We would need to add this flag to both the block creation and the block validation.  That would involve passing a flag into the ProtocolScheduleBuilder, and the question would be "is EIP 158 active?"

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/MainnetBlockProcessor.java
shemnon: yes, out of scope I'de say.  Possibly could become static methods.

ethereum/blockcreation/src/main/java/tech/pegasys/pantheon/ethereum/blockcreation/AbstractBlockCreator.java
shemnon: I just pused an alternate approach where we expose EIP158 status in the ProtocolSpec, seems to get to both locations with a minimum of grief.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/ProtocolSpec.java
```diff
@@ -237,6 +241,15 @@ public Wei getBlockReward() {
     return blockReward;
   }
 
+  /**
+   * EIP 158 prohibits adding new "empty" accounts to the account trie.
+   *
+   * @return if EIP158 is in effect for this spec.
+   */
+  public boolean isEIP158() {
```
mbaxter: Naming quibbles - maybe this could be `isEIP158Enabled`.  Not sure if it would be a good idea name it something more "meaningful" like `prohibitEmptyAccounts`?

shemnon:
I ran the reference tests but not the acceptance tests... fix is underway.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/ProtocolSpec.java
shemnon: I'll go with `IsEIP158Enabled` when I fix the acceptance tests.  My long term vision is that we have an `isEIPEnabled(int eip)` option and we track all the EIPs.  That way we can speculatively include EIPs in the code base and use the EIP registry as feature flags.  But since we care about only one at this point I'll put the number in the name,

shemnon: Turns out our clique tests dependency, so labeling it as EIP158 treatment is inaccurave, so I went with skipZeroBlockRewards.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/MainnetBlockProcessor.java
```diff
@@ -154,10 +158,12 @@ public Result processBlock(
...
       final MutableWorldState worldState,
       final ProcessableBlockHeader header,
-      final List<BlockHeader> ommers) {
-    if (blockReward.isZero()) {
+      final List<BlockHeader> ommers,
+      final boolean skipZeroBlockRewards) {
+    if (skipZeroBlockRewards && blockReward.isZero()) {
       return true;
     }
+    // This reference equality check is deliberate.
```
mbaxter: stray comment

ethereum/blockcreation/src/test/java/tech/pegasys/pantheon/ethereum/blockcreation/EthHashBlockCreatorTest.java
```diff
@@ -136,4 +143,62 @@ public void createMainnetBlock1_fixedDifficulty1() throws IOException {
     // If we weren't setting difficulty to 2^256-1 a difficulty of 1 would have caused a
     // IllegalArgumentException at the previous line, as 2^256 is 33 bytes.
   }
+
+  @Test
+  public void createMainnetBlock1_noReward() {
```
mbaxter: Suggest a name that's more specific to what you're testing:
```suggestion
  public void rewardBeneficiary_zeroReward_skipZeroRewardsFalse() {
```

ethereum/blockcreation/src/test/java/tech/pegasys/pantheon/ethereum/blockcreation/EthHashBlockCreatorTest.java
```diff
@@ -136,4 +143,62 @@ public void createMainnetBlock1_fixedDifficulty1() throws IOException {
...
+            .parentHash(Hash.ZERO)
+            .coinbase(BLOCK_1_COINBASE)
+            .difficulty(UInt256.ONE)
+            .number(1)
+            .gasLimit(1)
+            .timestamp(1)
+            .buildProcessableBlockHeader();
+
+    blockCreator.rewardBeneficiary(
+        mutableWorldState, header, Collections.emptyList(), Wei.ZERO, false);
```
mbaxter: Probably worth creating another version of this test where `skipZeroBlockRewards` is `true`

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/MainnetBlockProcessor.java
shemnon: removed.

ethereum/blockcreation/src/test/java/tech/pegasys/pantheon/ethereum/blockcreation/EthHashBlockCreatorTest.java
shemnon: Done

ethereum/blockcreation/src/test/java/tech/pegasys/pantheon/ethereum/blockcreation/EthHashBlockCreatorTest.java
shemnon: done

PR #1815, opened 2019-08-01
[PAN-2946] - changes in core JSON-RPC method to support ReTestEth

shemnon:
## PR description

Some of the methods need to have a changeable reference to stuff like
blockchainqueries.  For the impacted methods the solution is to wrap them
in a Supplier<> interface.  This includes new constructors for re-used methods.

Also include the debug_accountRangeAt method as it'ss namespaced as debug.

Also, rename the getter methods to align with standards.


ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/DebugAccountRangeAt.java
```diff
@@ -0,0 +1,98 @@
...
+
+    if (state.isEmpty()) {
+      return new DebugAccountRangeAtResult(Map.of(), Bytes32.ZERO.toUnprefixedString());
+    } else {
+      final List<Account> accounts =
+          state
+              .get()
+              .streamAccounts(Bytes32.fromHexStringLenient(addressHash), maxResults + 1)
+              .collect(Collectors.toList());
+      Bytes32 nextKey = Bytes32.ZERO;
```
mbaxter: For symmetry with `debug_storageRangeAt`, I think `nextKey` should be `null` (not returned) in this case.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/results/BlockResult.java
```diff
@@ -89,6 +102,7 @@
     this.timestamp = Quantity.create(header.getTimestamp());
     this.ommers = ommers;
     this.transactions = transactions;
+    this.coinbase = includeCoinbase ? header.getCoinbase().toString() : null;
```
mbaxter: I think it's cleaner to just always include `coinbase` instead of adding a flag to switch on this.  Shouldn't hurt anything if there's extra data returned that users aren't expecting.  Kind of odd that it's not a standard field anyway.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/results/BlockResult.java
```diff
@@ -185,4 +199,10 @@ public String getTimestamp() {
   public List<TransactionResult> getTransactions() {
     return transactions;
   }
+
+  @JsonGetter(value = "author")
```
mbaxter: This is an odd name for this field 🤔

ethereum/jsonrpc/src/test/java/tech/pegasys/pantheon/ethereum/jsonrpc/EthJsonRpcHttpBySpecTest.java
```diff
@@ -216,28 +216,8 @@ Formatter will be turned on to make this easier to read (one spec per line)
 
     specs.put(EthNewPendingTransactionFilter.class, "eth_newPendingTransactionFilter");
 
-    specs.put(EthUninstallFilter.class, "eth_uninstallFilter_NonexistentFilter");
-    specs.put(EthUninstallFilter.class, "eth_uninstallFilter_FilterIdTooLong");
-    specs.put(EthUninstallFilter.class, "eth_uninstallFilter_FilterIdNegative");
```
mbaxter: What happened to these - bad merge??

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/DebugAccountRangeAt.java
```diff
@@ -0,0 +1,98 @@
...
+
+  public DebugAccountRangeAt(
+      final JsonRpcParameter parameters, final Supplier<BlockchainQueries> blockchainQueries) {
+    super(blockchainQueries, parameters);
+    this.parameters = parameters;
+  }
+
+  @Override
+  protected BlockParameter blockParameter(final JsonRpcRequest request) {
+    return parameters.required(request.getParams(), 0, BlockParameter.class);
```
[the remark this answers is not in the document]
shemnon: The docs say hash, but we never see a hash in practice.  If it becomes a retesteth requirement I will probably go the `BlockParameterOrHash` route, but as of now the docs require a form that is not used.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/DebugAccountRangeAt.java
shemnon: retesteh requires the zeros return.  Not the most fault tolerant code.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/results/BlockResult.java
shemnon: Not only is it not in the JSON-RPC Spec, what reteseth wants is an `author` field which has the same meaning test wise as `coinbase`.  If we expose it for all I would still want a flag and we expose `coinbase` not `author`.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/results/BlockResult.java
shemnon: very odd.

ethereum/jsonrpc/src/test/java/tech/pegasys/pantheon/ethereum/jsonrpc/EthJsonRpcHttpBySpecTest.java
shemnon: bad merge.  Will fix.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/DebugAccountRangeAt.java
```diff
@@ -0,0 +1,98 @@
...
+
+  @Override
+  public String getName() {
+    return "debug_accountRangeAt";
+  }
+
+  @Override
+  protected Object resultByBlockNumber(final JsonRpcRequest request, final long blockNumber) {
+    final Object[] params = request.getParams();
+    final String addressHash = parameters.required(params, 2, String.class);
```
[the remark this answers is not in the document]
shemnon: Retesteth likes to pass in odd length addresses, particularly zero addresses.  Which is why I take it in as a string and at line 81 parse it in a lenient fashion.  Going straight to hash throws `Caused by: java.lang.IllegalArgumentException: Invalid odd-length hex binary representation 0`

ethereum/jsonrpc/src/test/resources/tech/pegasys/pantheon/ethereum/jsonrpc/debug_accountRangeAt_complete.json
```diff
@@ -0,0 +1,27 @@
+{
+  "request": {
+    "id": 414,
+    "jsonrpc": "2.0",
+    "method": "debug_accountRangeAt",
```
mbaxter: This method needs a rename for the tests too

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/DebugAccountRangeAt.java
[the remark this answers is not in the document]
shemnon: LooksLike DebugStorageRangeAt will need the `BlockParameterOrHash` treatment.

ethereum/jsonrpc/src/test/resources/tech/pegasys/pantheon/ethereum/jsonrpc/debug_accountRangeAt_complete.json
shemnon: fixed, temporarily.  Retesteth is changing the name but is unstable where they changed it.

shemnon:
`BlockParameterOrHash` change is in, to both Debug methods.

PR #1820, opened 2019-08-05
[PAN-2995] Make account.getAddress() return an Optional<Address>

abdelhamidbakhta:
## PR description
We should make `Account.getAddress()` return an `Optional` value.  This will allow us to display a null address value in the response to `debug_accountRangeAt` when the address preimage is not available.


ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/DebugAccountRange.java
```diff
@@ -102,7 +102,13 @@ public JsonRpcResponse response(final JsonRpcRequest request) {
                   .collect(
                       Collectors.toMap(
                           account -> account.getAddressHash().toString(),
-                          account -> account.getAddress().toString())),
+                          account -> {
```
mbaxter: (optional) You can make this more concise with something like:
```suggestion
                          account -> account.getAddress().map(Address::toString()).orElse(null)
```

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/DebugAccountRange.java
```diff
@@ -102,7 +102,13 @@ public JsonRpcResponse response(final JsonRpcRequest request) {
...
+                            if (account.getAddress().isPresent()) {
+                              return account.getAddress().get().toString();
+                            } else {
+                              return "null";
```
mbaxter: I think you actually want `null` here rather than the string `"null"`.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/worldstate/DefaultMutableWorldState.java
```diff
@@ -413,7 +413,10 @@ public void commit() {
         }
 
         // Save address preimage
-        wrapped.newAccountKeyPreimages.put(updated.getAddressHash(), updated.getAddress());
+        updated
+            .getAddress()
```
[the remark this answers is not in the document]
abdelhamidbakhta: With your previous suggested changes the number of Optional.get calls has clearly decreased. Why not in a follow up PR.

PR #1821, opened 2019-08-06
PAN-2985: eth_coinbase now returns the local node address for Clique/…

bgravenorst:
eth_coinbase now returns the local node address for Clique/IBFT 2.0 networks. The documentation now explicitly states this.


bgravenorst:
Ah ok, that makes sense. I'll close this PR in that case.

PR #1826, opened 2019-08-06
[PAN-2946] refactor normalizeKeys method

shemnon:
## PR description

Move the `normalizeKeys` method into JsonUtil as Retesteth will re-use it.


config/src/main/java/tech/pegasys/pantheon/config/JsonUtil.java
```diff
@@ -28,6 +29,31 @@
 
 public class JsonUtil {
 
+  /**
+   * Converts all to lowercase for easier lookup. This is useful incases such as the 'genesis.json'
```
mbaxter:
```suggestion
   * Converts all to lowercase for easier lookup. This is useful in cases such as the 'genesis.json'
```

PR #1830, opened 2019-08-07
[PAN-2818] Database versioning and enable multi-column database

RatanRSur:
## PR description
Turns on multi-column database by default.

Introduce database metadata json file to database directory, which only contains version info as of now.

* If there is no database, create the latest version.
* Otherwise use the database loader for the version found
  * If there is no database metadata found, assume version 0 loader
  * Otherwise use loader for version specified in database metadata file.

## Fixed Issue(s)
https://pegasys1.atlassian.net/secure/RapidBoard.jspa?rapidView=28&modal=detail&selectedIssue=PAN-2818


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/storage/keyvalue/DatabaseMetadata.java
```diff
@@ -0,0 +1,53 @@
...
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.nio.file.Path;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public class DatabaseMetadata {
+  private static final String METADATA_FILENAME = "DATABASE_METADATA.json";
```
shemnon: As far as I know our configuration files have all been `.toml` to date (auth, permissioning, CLI options).  How hard would it be to adapt this to a TOML format?

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/storage/keyvalue/RocksDbStorageProvider.java
```diff
@@ -27,13 +27,23 @@
...
 public class RocksDbStorageProvider {
   public static long DEFAULT_WORLD_STATE_PREIMAGE_CACHE_SIZE = 5_000L;
   private static final Logger LOG = LogManager.getLogger();
+  public static final int CURRENT_VERSION = 1;
+  /**
+   * This list contains the methods used to load the version at its index. Index 0 of the list loads
+   * a version 0 database and so on...
+   */
+  private static final List<StorageProviderFunction> PROVIDERS_BY_VERSION =
+      Arrays.asList(RocksDbStorageProvider::ofUnsegmented, RocksDbStorageProvider::ofSegmented);
```
shemnon: Would a Map<Integer,StorageProviderFunction> work better in the event that version numbers are non-consecutive?  For example if we drop support for a specific experimental version?

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/storage/keyvalue/DatabaseMetadata.java
```diff
@@ -0,0 +1,53 @@
...
+    this.version = version;
+  }
+
+  public int getVersion() {
+    return version;
+  }
+
+  static DatabaseMetadata fromDirectory(final Path databaseDir) throws IOException {
+    try {
+      return MAPPER.readValue(getDefaultMetadataFile(databaseDir), DatabaseMetadata.class);
```
shemnon: What about the case where the file exists but is corrupted?  I would prefer a total service failure instead of proceeding at version 0, which may be incorrect.

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/storage/keyvalue/RocksDbStorageProviderTest.java
```diff
@@ -0,0 +1,69 @@
...
+  @Test
+  public void shouldDetectCorrectVersionIfMetadataFileExists() throws Exception {
+    final Path tempDatabaseDir = temporaryFolder.newFolder().toPath().resolve("db");
+    Files.createDirectories(tempDatabaseDir);
+    tempDatabaseDir.resolve("IDENTITY").toFile().createNewFile();
+    new DatabaseMetadata(1).writeToDirectory(tempDatabaseDir);
+    when(rocksDbConfiguration.getDatabaseDir()).thenReturn(tempDatabaseDir);
+    RocksDbStorageProvider.create(rocksDbConfiguration, metricsSystem);
+    assertEquals(1, DatabaseMetadata.fromDirectory(tempDatabaseDir).getVersion());
+  }
```
shemnon: per my previous comment, we need a case where the file exists and is invalid.  (corrupt, has unsupported version number, etc.)

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/storage/keyvalue/RocksDbStorageProvider.java
```diff
@@ -46,34 +56,48 @@ public static StorageProvider create(
...
-      return createUnsegmentedProvider(
-          rocksDbConfiguration, metricsSystem, worldStatePreimageCacheSize);
+      databaseVersion = CURRENT_VERSION;
+      LOG.info("No existing database detected. Using version {}", databaseVersion);
+      Files.createDirectories(databaseDir);
+      new DatabaseMetadata(databaseVersion).writeToDirectory(databaseDir);
+    }
+
+    try {
+      return PROVIDERS_BY_VERSION
```
shemnon: If we open a DB of the wrong version what happens?  Does it fail sanity checks or does it fail catastrophically and destroy data?

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/storage/keyvalue/RocksDbStorageProvider.java
RatanRSur: Good idea!

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/storage/keyvalue/DatabaseMetadata.java
mbaxter: Genesis files are JSON-based, and this is an internal metadata file rather than user-facing config so I don't think consistency is a big concern ...  Which is to say, I don't think its a big deal if we use JSON here.  I also think that `jackson` is a pretty good library to lean on and it doesn't support TOML.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/storage/keyvalue/RocksDbStorageProvider.java
RatanRSur: What does rocks db do when you try to load a database with options/schema it wasn't created with?

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/storage/keyvalue/DatabaseMetadata.java
RatanRSur: When the file is corrupted, currently it's a failure with a jackson `UnrecognizedPropertyException` since I'm only catching `FileNotFoundException`. Do you want me to wrap it our own exception or is the `UnrecognizedPropertyException` ok?
```
com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "🦄" (class tech.pegasys.pantheon.ethereum.storage.keyvalue.DatabaseMetadata), not marked as ignorable (one known property: "version"])
 at [Source: (File); line: 1, column: 11] (through reference chain: tech.pegasys.pantheon.ethereum.storage.keyvalue.DatabaseMetadata["🦄"])
```

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/storage/keyvalue/RocksDbStorageProvider.java
shemnon: I'm not sure, we should try and see if any data changes.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/storage/keyvalue/DatabaseMetadata.java
shemnon: We use the Cava TOML parser in places like the whitelist persister - https://github.com/PegaSysEng/pantheon/blob/3fffd605dd065a2efbb8d793429713a3d9076d1d/ethereum/permissioning/src/main/java/tech/pegasys/pantheon/ethereum/permissioning/WhitelistPersistor.java#L89

The RocksDB config files in that directory, OPTIONS-#####, are also TOML based.  genesis.json seems to be the only outlier when it comes to files living on the file system.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/storage/keyvalue/DatabaseMetadata.java
shemnon: My main concern is that we don't continue at version 0 unless the file is truly not there.

RatanRSur: Yeah, if the file doesn't exist, then we continue at version 0. Otherwise we try to parse it. I added the test cases you mentioned in the other comments and will go ahead and wrap the exception.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/storage/keyvalue/RocksDbStorageProvider.java
ajsutton: Depends on the changes.  If you try to open a database for writing without specifying all the columns used in that database it will throw an exception and refuse to open.  So opening a columnar db with our legacy non-columnar format will just refuse to open, but I doubt that's true of all potentially incompatible changes we'd make.

RatanRSur: Response to my question on RocksDB dev group:

> I cannot think of an option that could not be changed back. Your right that some options have persistent impact on DB but it should not prevent the DB to be successfully loaded after the option change. In many cases, if the option has a persistent impact it is remembered in the properties of the SST file, so the SST file could still be read properly even if the option is not set anymore. The newly generated files, however, will follow the newest values of options. For "optimize filters for hits" option in particular, you might need to do a manual compaction on all db if you want to see the optimization impact of the option change immediately.

We're not at risk of accidentally breaking the database by loading it with the wrong options. If we do make a developer error and the database is being loaded with the wrong version, it will either fail like @ajsutton mentioned for columns, which is a deep structural property, or load with performance characteristics we're not expecting. At least this case is something that doesn't destroy the existing data while we push out a fix.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/storage/keyvalue/DatabaseMetadata.java
RatanRSur: Unless I'm missing something, it looks like cava is archived? https://github.com/ConsenSys/cava

It's much defintely much cleaner in jackson than cava but I see your point about consistency :P

It seems our options are:
* bring in a dependency on a maintained toml library just for this config
  * then in another pr, migrate other areas over to that library, removing cava dependency
* use jackson

RatanRSur: Marking this resolved for now as cava doesn't have utilities to write toml. We would have to pull in another toml dependency that has writing or roll our own.

With just the version flag as of now, it would be relatively simple to write the string on our own but in case we end up using more of the toml grammar in the future, it could be unwieldy to maintain.

PR #1832, opened 2019-08-07
[PIE-1809] Add import chain json utility

mbaxter:
## PR description
Add a CLI utility for importing blocks from a json file containing minimally specified block and transaction data.

Example command:
```
pantheon --network=DEV blocks import --format=JSON --from="/path/to/file.json"
```

Example import file:
```
{
  "blocks": [
    {
      "number": 1,
      "transactions": [
        {
          "fromPrivateKey": "8f2a55949038a9610f50fb23b5883af3b4ecb3c3bb792cbcefbd1542c692be63",
          "gasLimit": "0xFFFFF1",
          "gasPrice": "0xFF",
          "value": "0x01",
          "to": "627306090abaB3A6e1400e9345bC60c78a8BEf57"
        },
        {
          "fromPrivateKey": "8f2a55949038a9610f50fb23b5883af3b4ecb3c3bb792cbcefbd1542c692be63",
          "gasLimit": "0xFFFFF2",
          "gasPrice": "0xEF",
          "to": "f17f52151EbEF6C7334FAD080c5704D77216b732"
        }
      ]
    }
  ]
}
```

Example output:
```
2019-08-07 19:07:58.453-04:00 | main | INFO  | BlocksSubCommand | Import JSON block data from /path/to/file.json 
2019-08-07 19:07:59.334-04:00 | main | INFO  | KeyPairUtil | Loaded key 0x5285b91de23f6e519836e0d0fe8413eef43f453a2e0af2d154b6bbae92175bdbfca37d0eec7095bf4b8ddfbf813aa669b26640dbeaf218ed3846744cd869bc01 from /data/key 
2019-08-07 19:07:59.594-04:00 | main | INFO  | ProtocolScheduleBuilder | Protocol schedule created with milestones: [ConstantinopleFix: 0] 
2019-08-07 19:07:59.775-04:00 | main | WARN  | ChainImporter | Importing to a non-empty database with chain height 2.  This may cause imported blocks to be considered non-canonical. 
2019-08-07 19:07:59.828-04:00 | main | INFO  | ChainImporter | Preparing to import block at height 1 (parent: 0xa08d1edb37ba1c62db764ef7c2566cbe368b850f5b3762c6c24114a3fd97b87f) 
2019-08-07 19:08:02.961-04:00 | main | INFO  | ChainImporter | Successfully created and imported block at height 1 (0xb062d54fbcdbe150f2ee38d10f2393c796f66048d5fc36ef8cee02c0bf807e3e) 
2019-08-07 19:08:02.967-04:00 | main | WARN  | ChainImporter | 1 / 1 imported blocks are not on the canonical chain: #1 (0xb062d54fbcdbe150f2ee38d10f2393c796f66048d5fc36ef8cee02c0bf807e3e) 
```


pantheon/src/main/java/tech/pegasys/pantheon/cli/subcommands/blocks/BlocksSubCommand.java
```diff
@@ -256,4 +322,9 @@ private void outputBlock(final List<Block> blocks) {
     }
     return metricsService;
   }
+
+  @FunctionalInterface
+  public interface ChainImporterFactory {
```
RatanRSur: Why does this need to be an explicit functional interface? Could we not use a `Supplier`?

pantheon/src/main/java/tech/pegasys/pantheon/chainimport/ChainImporter.java
```diff
@@ -0,0 +1,251 @@
...
+    Optional<Block> maybeBlock =
+        miner.createBlock(parentHeader, transactions, Collections.emptyList());
+
+    if (maybeBlock.isEmpty()) {
+      // Some MiningCoordinator's (specific to consensus type) do not support block-level imports
+      throw new IllegalArgumentException(
+          "Unable to create block using current consensus engine: "
+              + genesisConfigOptions.getConsensusEngine());
+    }
+    return maybeBlock.get();
```
RatanRSur: (optional) use `orElseThrow`

pantheon/src/main/java/tech/pegasys/pantheon/chainimport/ChainImporter.java
```diff
@@ -0,0 +1,251 @@
...
+    this.controller = controller;
+    mapper = new ObjectMapper();
+    // Jdk8Module allows us to easily parse {@code Optional} values from json
+    mapper.registerModule(new Jdk8Module());
+    // Ignore casing of properties
+    mapper.configure(MapperFeature.ACCEPT_CASE_INSENSITIVE_PROPERTIES, true);
+  }
+
+  public void importChain(final String chainJson) throws IOException {
+    checkDatabaseIsEmpty();
```
RatanRSur: (optional) this sounds a bit like it would fail if untrue, how about `warnIfDatabaseNotEmpty`?

pantheon/src/main/java/tech/pegasys/pantheon/chainimport/ChainData.java
```diff
@@ -0,0 +1,33 @@
+/*
```
RatanRSur: Is there a way to parse the blocks directly into a sequence without the need of a separate class?

pantheon/src/main/java/tech/pegasys/pantheon/chainimport/ChainImporter.java
```diff
@@ -0,0 +1,251 @@
...
+
+    ChainData chainData = mapper.readValue(chainJson, ChainData.class);
```
RatanRSur: This reads like there's probably more to ChainData than just a sequence of blocks

RatanRSur: can this be `final`?

pantheon/src/main/java/tech/pegasys/pantheon/chainimport/ChainImporter.java
```diff
@@ -0,0 +1,251 @@
...
+
+    final List<Block> importedBlocks = new ArrayList<>();
+    for (final BlockData blockData : chainData.getBlocks()) {
+      final BlockHeader parentHeader = getParentHeader(blockData, importedBlocks);
+      final Block importedBlock = processBlockData(blockData, parentHeader);
+      importedBlocks.add(importedBlock);
+    }
+
+    // Check that imported blocks form the canonical chain, if not print a warning
+    this.checkImportedBlocksAreOnCanonicalChain(importedBlocks);
```
RatanRSur: (optional) The comment could probably be removed (and thus not at risk of getting outdated) if the method was something like `warnIfImportedBlocksNotOnCanonicalChain`

pantheon/src/main/java/tech/pegasys/pantheon/chainimport/ChainImporter.java
```diff
@@ -0,0 +1,251 @@
...
+      final List<Transaction> transactions) {
+    final MiningCoordinator miner = controller.getMiningCoordinator();
+    final GenesisConfigOptions genesisConfigOptions = controller.getGenesisConfigOptions();
+    setOptionalFields(miner, blockData, genesisConfigOptions);
+
+    Optional<Block> maybeBlock =
+        miner.createBlock(parentHeader, transactions, Collections.emptyList());
+
+    if (maybeBlock.isEmpty()) {
+      // Some MiningCoordinator's (specific to consensus type) do not support block-level imports
```
RatanRSur:
```suggestion
      // Some MiningCoordinators (specific to consensus type) do not support block-level imports
```

pantheon/src/test/java/tech/pegasys/pantheon/cli/subcommands/blocks/BlockSubCommandTest.java
```diff
@@ -64,11 +69,13 @@
           + System.lineSeparator();
 
   private static final String EXPECTED_BLOCK_IMPORT_USAGE =
-      "Usage: pantheon blocks import [-hV] --from=<FILE>"
+      "Usage: pantheon blocks import [-hV] [--format=<format>] --from=<FILE>"
           + System.lineSeparator()
           + "This command imports blocks from a file into the database."
           + System.lineSeparator()
-          + "      --from=<FILE>   File containing blocks to import"
+          + "      --format=<format>   The type of data to be imported (default: RLP)."
```
RatanRSur: Is there somewhere the user can see that JSON is a valid option to this flag?

pantheon/src/main/java/tech/pegasys/pantheon/cli/subcommands/blocks/BlocksSubCommand.java
mbaxter: We could use `Function<PantheonController<T>, ChainImporter<T>>` instead.  But, I prefer making explicit interfaces for reasons of readability and discoverability.  It's a pattern we use pretty extensively in the codebase.

pantheon/src/main/java/tech/pegasys/pantheon/chainimport/ChainData.java
mbaxter: Maybe?  But I think the structure is pretty clear as is and it provides a place to add additional data if we need it in the future.

pantheon/src/main/java/tech/pegasys/pantheon/chainimport/ChainImporter.java
mbaxter: yep - I always miss the finals 😄

pantheon/src/main/java/tech/pegasys/pantheon/chainimport/ChainImporter.java
mbaxter: Since `MiningCoordinator` is a class not a word, I usually pluralize with `'s` 🤷‍♀

pantheon/src/test/java/tech/pegasys/pantheon/cli/subcommands/blocks/BlockSubCommandTest.java
mbaxter: Good call - will update!

pantheon/src/main/java/tech/pegasys/pantheon/cli/subcommands/blocks/BlocksSubCommand.java
RatanRSur: Interesting, I find the explicit interface less readable because the parameter looks like it could be a full blown object with more methods. Not a blocking issue though

pantheon/src/main/java/tech/pegasys/pantheon/cli/subcommands/blocks/BlocksSubCommand.java
```diff
@@ -121,7 +121,9 @@ public void run() {
 
     @Option(
         names = "--format",
-        description = "The type of data to be imported (default: ${DEFAULT-VALUE}).",
+        hidden = true,
+        description =
+            "The type of data to be imported, possible values are: ${COMPLETION-CANDIDATES} (default: ${DEFAULT-VALUE}).",
```
shemnon: I didn't know `${COMPLETION-CANDIDATES}` existed.  Looks like I should open another PR to add this to other options as well.

PR #1833, opened 2019-08-08
[PAN-2900] Fix FlatFileTaskCollection tests

shemnon:
## PR description

Windows has some slightly different Java NIO semantics, regarding size
after writing and whether files are deleted on close (they are not immediately deleted).

The first issue is we shouldn't be using Integer.SIZE when we mean
Integer.BYTES.  Integer.SIZE is bit count, BYTES is byte count.

The second issue is we cannot count on these work files showing up or
being deleted from the file system consistently across platforms.  The
ordering is consistent within platforms but not across.  The test was
re-written to check the read and write file numbers instead.

### DIFF
--- a/services/tasks/src/test/java/tech/pegasys/pantheon/services/tasks/FlatFileTaskCollectionTest.java
+++ b/services/tasks/src/test/java/tech/pegasys/pantheon/services/tasks/FlatFileTaskCollectionTest.java
@@ -51,17 +51,15 @@ public class FlatFileTaskCollectionTest
       final List<BytesValue> tasks = new ArrayList<>();
 
       addItem(queue, tasks, 0);
-      final File[] currentFiles = getCurrentFiles(dataDir);
-      assertThat(currentFiles).hasSize(1);
-      final File firstFile = currentFiles[0];
+      assertThat(queue.getWriteFileNumber()).isEqualTo(0);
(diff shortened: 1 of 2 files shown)


services/tasks/src/test/java/tech/pegasys/pantheon/services/tasks/FlatFileTaskCollectionTest.java
```diff
@@ -72,8 +69,7 @@ public void shouldRollFilesWhenSizeExceeded() throws Exception {
         removedTasks.add(queue.remove().getData());
       }
 
-      // Fully read files should have been removed.
-      assertThat(getCurrentFiles(dataDir)).doesNotContain(firstFile);
+      assertThat(queue.getReadFileNumber()).isEqualTo(0);
```
mbaxter: If the first file has been read and closed, shouldn't the readFile number increment to 1??
```suggestion
      assertThat(queue.getReadFileNumber()).isEqualTo(1);
```

shemnon: The file number doesn't get updated until it actually reads the file.  This change will break the test but I'm adding a tweak to make sure it advances.

PR #1836, opened 2019-08-09
Support streaming for Pending transactions events

abdelhamidbakhta:
## PR description
Extend the functionality implemented in `PIE-1544: Events Data Stream for newHeads` to also support Pending Transaction events, as defined in Pantheon’s documentation. 

- update plugin api
- use rich data object for new blocks propagated
- add adapters to build plugin rich data objects from ethereum core objects
- add tests
- add new transaction added listener
- add transaction pool reference to `PantheonEventsImpl`


gradle/versions.gradle
```diff
@@ -89,6 +89,6 @@ dependencyManagement {
 
     dependency 'org.xerial.snappy:snappy-java:1.1.7.3'
 
-    dependency "tech.pegasys.pantheon:plugin-api:1.1.3"
+    dependency "tech.pegasys.pantheon:plugin-api:1.2.0-SNAPSHOT"
```
shemnon: We can't depend on snapshot versions.  This will need to wait until we get the plugin-api version updated.

ethereum/core/build.gradle
```diff
@@ -38,6 +38,8 @@ dependencies {
   implementation 'com.google.guava:guava'
   implementation 'io.vertx:vertx-core'
   implementation 'org.apache.logging.log4j:log4j-api'
+  implementation 'tech.pegasys.pantheon:plugin-api'
+
```
shemnon:
```suggestion
```

PR #1837, opened 2019-08-09
[PIE-1846] Add an experimental flag for disabling timers

mbaxter:
## PR description
Add an experimental flag for disabling metrics timers: `--Xmetrics-timers-enabled=<boolean>`

### DIFF
--- a/pantheon/src/main/java/tech/pegasys/pantheon/cli/options/MetricsCLIOptions.java
+++ b/pantheon/src/main/java/tech/pegasys/pantheon/cli/options/MetricsCLIOptions.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2019 ConsenSys AG.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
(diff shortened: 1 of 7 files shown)


pantheon/src/main/java/tech/pegasys/pantheon/cli/options/MetricsCLIOptions.java
```diff
@@ -0,0 +1,53 @@
...
+package tech.pegasys.pantheon.cli.options;
+
+import tech.pegasys.pantheon.metrics.prometheus.MetricsConfiguration;
+
+import java.util.Arrays;
+import java.util.List;
+
+import picocli.CommandLine;
+
+public class MetricsCLIOptions implements CLIOptions<MetricsConfiguration.Builder> {
```
shemnon: Not for this PR, but long term do you see us migrating the metrics HTTP Server CLI options here too?

mbaxter: Yeah, I think it would be great if we could start grouping all of the related options regardless of whether they're experimental or not.

PR #1840, opened 2019-08-12
Use git show-ref to retrieve hash reference

mbaxter:
## PR description
Use `git show-ref` to retrieve the hash associate with a git reference.  This handles the case where the current branch reference is stored in `.git/packed-refs` but not `.git/refs/heads`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -579,8 +579,21 @@ def getCheckedOutGitCommitHash() {
 
   if (isCommit) return head[0].trim().take(takeFromHash) // e5a7c79edabb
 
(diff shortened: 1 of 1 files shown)


shemnon:
This create a hard requirement on git being present to do a build.  Many of our build containers do not have git present.

Can the old method be kept as a primary and the show-ref version be attempted when it fails?

ajsutton:
For the record, I believe we've added git to all our build containers now. The latest version of the release plugin required it.  But we're planning to remove the release plugin entirely so maybe would be useful to make it work with or without git if it's not too hard.

mbaxter:
Cool, that makes sense.  Will update 👍

PR #1844, opened 2019-08-13
PrivacyMarkerTransaction to be signed with a randomly generated key

rain-on:
This udpates the PrivateTransactionHandler to sign privacy marker transactions with a randomly generated keypair (rather than the node key).

The code has been architectured such that moving forwards the user can either specify a keyfile on the command line (and pass into PrivacyParameters.Builder) - or the system will otherwise perform random-key signing.


pantheon/src/main/java/tech/pegasys/pantheon/controller/PantheonControllerBuilder.java
```diff
@@ -182,7 +182,6 @@
     checkArgument(
         storageProvider == null || rocksDbConfiguration == null,
         "Must supply either storage provider or RocksDB confguration, but not both");
-    privacyParameters.setSigningKeyPair(nodeKeys);
```
iikirilov: ATs are failing because the PMT signing key should be set somewhere. From the changes in `PrivacyParameters` they should be set in the builder somewhere. The reason for setting the signing key here is that the `nodeKeys` object was not accessible any earlier when initially implemented.

In JsonRpcMethodsFactory.createPrivateMarkerTransactionFactory() you will always be returning a `RandomSigningPrivateMarkerTransactionFactory` which will sign the PMT with a key that represents an account that has no funds in `dev.json` genesis file used in the AT

pantheon/src/main/java/tech/pegasys/pantheon/cli/PantheonCommand.java
```diff
@@ -1185,7 +1185,8 @@ private void synchronize(
       final WebSocketConfiguration webSocketConfiguration,
       final MetricsConfiguration metricsConfiguration,
       final Optional<PermissioningConfiguration> permissioningConfiguration,
-      final Collection<EnodeURL> staticNodes) {
+      final Collection<EnodeURL> staticNodes)
+      throws IOException {
```
ajsutton: I can't immediately see why this would now throw IOException.  I assume there's a chain of calls it now flows up that's not obvious in GitHub's UI but could you double check it is required with the final code?

rain-on: Removed - was hacked in, but resolved more suitably.

pantheon/src/main/java/tech/pegasys/pantheon/controller/PantheonControllerBuilder.java
rain-on: Thanks for the help, that would have been hard to track down.

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/acceptance/dsl/node/ProcessPantheonNodeRunner.java
```diff
@@ -92,6 +92,8 @@ public void startNode(final PantheonNode node) {
       params.add(node.getPrivacyParameters().getEnclavePublicKeyFile().getAbsolutePath());
       params.add("--privacy-precompiled-address");
       params.add(String.valueOf(node.getPrivacyParameters().getPrivacyAddress()));
+      params.add("--privacy-marker-transaction-signing-key-file");
```
iikirilov: I prefer a flag `--random-marker-signing` (or similar) which is a boolean that defaults to `false`. If it is `false` then the marker transaction signing key defaults to `--node-private-key-file` i.e. uses the fixed transaction signing strategy.

Otherwise, use the random transaction signing strategy.

iikirilov: That being said - this is also a viable approach. The downside is that in most cases a user needs to set `--node-private-key-file` and `--privacy-marker-transaction-signing-key-file` to the same value.

rain-on: The problem I see is when we go to mainnet, the account with funds to sign transactions may not be the same key as that used by the node.
Having said that - I suspect this needs to change to be "--signer-url", and Pantheon will need to submit the marker transaction to EthSigner to do the key-management stuff... Not sure how that plays with Acceptance Tests :/

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/JsonRpcMethodsFactory.java
```diff
@@ -360,4 +369,21 @@ public static void addMethods(
       methods.put(rpcMethod.getName(), rpcMethod);
     }
   }
+
+  private PrivateMarkerTransactionFactory createPrivateMarkerTransactionFactory(
+      final PrivacyParameters privacyParameters,
+      final BlockchainQueries blockchain,
```
jframe: Variable name is bit misleading, blockchainQueries?

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/privacy/markertransaction/FixedKeySigningPrivateMarkerTransactionFactory.java
```diff
@@ -0,0 +1,45 @@
...
+
+public class FixedKeySigningPrivateMarkerTransactionFactory
+    extends PrivateMarkerTransactionFactory {
+
+  private final NonceProvider nonceProvider;
+  private final KeyPair signingKey;
+  private final Address sender;
+
+  public FixedKeySigningPrivateMarkerTransactionFactory(
+      final Address privacyPrecompileAddress,
```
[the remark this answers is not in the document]
rain-on: Not sure I understand?

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/LatestNonceProvider.java
```diff
@@ -0,0 +1,39 @@
...
+import tech.pegasys.pantheon.ethereum.core.Address;
+import tech.pegasys.pantheon.ethereum.eth.transactions.PendingTransactions;
+import tech.pegasys.pantheon.ethereum.jsonrpc.internal.queries.BlockchainQueries;
+import tech.pegasys.pantheon.ethereum.util.NonceProvider;
+
+import java.util.OptionalLong;
+
+public class LatestNonceProvider implements NonceProvider {
+
+  final BlockchainQueries blockchain;
```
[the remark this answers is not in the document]
rain-on: done.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/privacy/markertransaction/PrivateMarkerTransactionFactory.java
```diff
@@ -0,0 +1,50 @@
...
+
+public abstract class PrivateMarkerTransactionFactory {
+
+  private final Address privacyPrecompileAddress;
+
+  public PrivateMarkerTransactionFactory(final Address privacyPrecompileAddress) {
+    this.privacyPrecompileAddress = privacyPrecompileAddress;
+  }
+
+  protected Address getPrivacyPrecompileAddress() {
```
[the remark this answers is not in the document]
rain-on: done.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/privacy/markertransaction/RandomSigningPrivateMarkerTransactionFactory.java
```diff
@@ -0,0 +1,32 @@
...
+
+  public RandomSigningPrivateMarkerTransactionFactory(final Address privacyPrecompileAddress) {
+    super(privacyPrecompileAddress);
+  }
+
+  @Override
+  public Transaction create(
+      final String transactionEnclaveKey, final PrivateTransaction privateTransaction) {
+    final KeyPair signingKey = KeyPair.generate();
+    return create(transactionEnclaveKey, privateTransaction, 0, signingKey);
```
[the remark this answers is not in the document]
rain-on: Went over this with AJS - we _could_ - but Ethereum is based on the premise that we won't have key collisions.
If we wanted to go down that route, we could extract the nonce from the worldstate then if not 0 - try another generation ... and then? Repeat until successful?

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/privacy/markertransaction/FixedKeySigningPrivateMarkerTransactionFactoryTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+  @Before
+  public void setup() {
+    when(privTransaction.getGasPrice()).thenReturn(gasPrice);
+    when(privTransaction.getGasLimit()).thenReturn(gasLimit);
+    when(privTransaction.getValue()).thenReturn(value);
+  }
+
+  @Test
+  public void nonceInsertedToTransactionIsThatProvidedFromNonceProvider() {
```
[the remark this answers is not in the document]
rain-on: yup - better. Done.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/JsonRpcMethodsFactory.java
rain-on: done.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/privacy/markertransaction/FixedKeySigningPrivateMarkerTransactionFactory.java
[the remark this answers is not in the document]
jframe: whoops meant to say the parameters could be final

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/privacy/markertransaction/RandomSigningPrivateMarkerTransactionFactory.java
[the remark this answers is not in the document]
jframe: True it all starts to fall apart if we are wondering about key collisions. Don't think we need to do anything then.

PR #1845, opened 2019-08-13
[PIE-1810] Add read-only blockchain factory method

mbaxter:
## PR description
This PR creates separate factory methods for mutable / immutable versions of `DefaultBlockchain`.  For the read-only, immutable blockchain, we don't require a genesis block to be supplied as long as the blockchain storage is non-empty.


ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/chain/DefaultBlockchainTest.java
```diff
@@ -734,11 +793,19 @@ private void assertTotalDifficultiesAreConsistent(final Blockchain blockchain, f
     assertEquals(td, blockchain.getChainHead().getTotalDifficulty());
   }
 
-  private DefaultMutableBlockchain createBlockchain(
+  private BlockchainStorage createStorage(final KeyValueStorage kvStore) {
+    return new KeyValueStoragePrefixedKeyBlockchainStorage(
+        kvStore, new MainnetBlockHeaderFunctions());
+  }
+
+  private DefaultBlockchain createMutableBlockchain(
```
shemnon: Does this need to be `DefaultBlockchain` or can it simply be `Blockchain`? i.e. do the tests need methods on `DefaultBlockchain` or is the type needed for other calls?

mbaxter: Some of the tests access protected methods, so just cast it here for simplicity.

PR #1850, opened 2019-08-14
Privacy multi-orion handler

jframe:
## PR description
Provides a abstraction hiding the concept of multiple enclaves so that existing code will be able to use one or many enclaves without any significant changes (will need to use this handler though).

* This is first piece that hides that multiple Orions are present
* Introduces the concept of an EnclaveUser. There is a static defaultUser that can be used to define the single user that is currently used
* This doesn't replace uses of Enclave yet, that will be done in another ticket.


ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/privacy/EnclaveHandlerTest.java
```diff
@@ -0,0 +1,224 @@
...
+    assertThatThrownBy(() -> enclaveHandler.sendData(unknownEnclaveUser, sendRequest))
+        .isInstanceOf(EnclaveException.class);
+    verifyZeroInteractions(enclave1, enclave2, enclave3);
+  }
+
+  @Test
+  public void retrievesDataForEnclaveUser() throws Exception {
+    final ReceiveRequest receiveRequest = new ReceiveRequest("abc123", "user2PublicKey");
+
+    when(enclave1.receive(any(ReceiveRequest.class))).thenThrow(EnclaveException.class);
```
rain-on: Is specifying "ReceiveRequest.class" as a parameter to "any', required, or just a nicety? I don't do it . ... (should I?)

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/privacy/EnclaveHandlerTest.java
```diff
@@ -0,0 +1,224 @@
...
+    when(enclave2.receive(any(ReceiveRequest.class))).thenThrow(EnclaveException.class);
+    when(enclave3.receive(receiveRequest)).thenReturn(new ReceiveResponse(null, "privacyGroup"));
+
+    assertThat(enclaveHandler.retrieveData(receiveRequest).getPrivacyGroupId())
+        .isEqualTo("privacyGroup");
+    verify(enclave1).receive(any());
```
rain-on: should it be receive(any()) or should it be receive(receiveRequest)?
If so - this applies to other "verify" checks as well.

util/src/main/java/tech/pegasys/pantheon/util/FutureUtils.java
```diff
@@ -131,4 +135,35 @@ public static void propagateCancellation(
           return null;
         });
   }
+
+  /**
+   * Creates a {@link CompletableFuture} that is completed when any of the provided
+   * CompletableFutures completes and matches the supplied predicate.
```
rain-on:
```suggestion
   * CompletableFutures complete and matches the supplied predicate.
```

util/src/test/java/tech/pegasys/pantheon/util/FutureUtilsTest.java
```diff
@@ -188,6 +192,50 @@ public void shouldNotCallErrorHandlerWhenFutureCompletesSuccessfully() {
     assertThat(result).isCompletedWithValue("Done");
   }
 
+  @Test
+  public void shouldReturnMatchingSuccessfulFutureWhenUsingAnyOfMatching() {
+    final CompletableFuture<String> future1 = new CompletableFuture<>();
+    final CompletableFuture<String> future2 = new CompletableFuture<>();
+    final CompletableFuture<String> future3 = new CompletableFuture<>();
+    future1.complete("a");
```
rain-on: it it worth showing that the future can complete *after* the creation of the "collation" future? (rather than completing them all prior to construction)?
Or maybe show that it is not complete if a sub-item completes with an non-matching value? I.e. do a multi-step test?

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/privacy/EnclaveHandlerTest.java
```diff
@@ -0,0 +1,224 @@
...
+  public void setup() {
+    final Map<EnclaveUser, Enclave> enclaves =
+        ImmutableMap.of(enclaveUser1, enclave1, enclaveUser2, enclave2, enclaveUser3, enclave3);
+    enclaveHandler = new EnclaveHandler(enclaves);
+  }
+
+  @Test
+  public void sendsDataToUsersOrion() throws Exception {
+    final SendRequest sendRequest = new SendRequestPantheon("payload", "from", "privacyGroup");
+    when(enclave1.send(sendRequest)).thenReturn(new SendResponse("abc123"));
```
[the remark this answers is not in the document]
jframe: done

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/privacy/EnclaveHandlerTest.java
```diff
@@ -0,0 +1,224 @@
...
+
+    assertThat(enclaveHandler.sendData(enclaveUser1, sendRequest))
+        .isEqualToComparingFieldByField(new SendResponse("abc123"));
+    verify(enclave1).send(sendRequest);
+    verifyZeroInteractions(enclave2);
```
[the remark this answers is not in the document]
jframe: yes it should. done

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/privacy/EnclaveHandlerTest.java
jframe: I was going to say nicety, but I don't think it adds any value here as the type for the parameter is strict. making it simpler using just any()

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/privacy/EnclaveHandlerTest.java
jframe: should be receive(receiveRequest) for the verify. fixed everywhere hopefully.

util/src/main/java/tech/pegasys/pantheon/util/FutureUtils.java
jframe: done

util/src/test/java/tech/pegasys/pantheon/util/FutureUtilsTest.java
jframe: I think it is, will try add something

PR #1852, opened 2019-08-14
[PIE-1810] Update export subcommand to export blocks in rlp format

mbaxter:
## PR description
Update blocks export subcommand to export blocks in rlp format.  Rearrange import / export utilities so that these classes are grouped together with consistent naming.


pantheon/src/main/java/tech/pegasys/pantheon/chainexport/BlockExporter.java
```diff
@@ -0,0 +1,94 @@
...
+      final Optional<Long> maybeStartBlock,
+      final Optional<Long> maybeEndBlock)
+      throws IOException {
+
+    // Get range to export
+    final long startBlock = maybeStartBlock.orElse(BlockHeader.GENESIS_BLOCK_NUMBER);
+    final long endBlock = maybeEndBlock.orElse(blockchain.getChainHeadBlockNumber() + 1L);
+    checkArgument(startBlock >= 0 && endBlock >= 0, "Start and end blocks must be greater than 0.");
+    checkArgument(startBlock < endBlock, "Start block must be less than end block");
+
```
shemnon: [optional] maybe a check arg that startblock <= chainHead?

pantheon/src/main/java/tech/pegasys/pantheon/chainexport/RlpBlockExporter.java
```diff
@@ -0,0 +1,35 @@
...
+
+  public RlpBlockExporter(final ProtocolContext<?> context) {
+    super(context);
+  }
+
+  @Override
+  protected void exportBlock(final FileOutputStream outputStream, final Block block)
+      throws IOException {
+    final BytesValue rlp = RLP.encode(block::writeTo);
+    outputStream.write(rlp.getArrayUnsafe());
```
abdelhamidbakhta: Question: Are you writing binary data in the file ? Is it what we want ?

pantheon/src/main/java/tech/pegasys/pantheon/cli/subcommands/blocks/BlocksSubCommand.java
```diff
@@ -241,71 +241,127 @@ private MiningParameters getMiningParameters() {
...
+        names = "--format",
+        hidden = true,
+        description =
+            "The format to export, possible values are: ${COMPLETION-CANDIDATES} (default: ${DEFAULT-VALUE}).",
+        arity = "1..1")
+    private final BlockExportFormat format = BlockExportFormat.RLP;
+
     @Option(
         names = "--to",
+        required = true,
```
shemnon: So we don't currently support export to standard output correct?  An absent `--to` option is how I wold expect that to be configured.  It may be interesting for json export but it sounds wrong for RLP export.

pantheon/src/main/java/tech/pegasys/pantheon/cli/subcommands/blocks/BlocksSubCommand.java
```diff
@@ -241,71 +241,127 @@ private MiningParameters getMiningParameters() {
...
-        } else if (!exportResult.allBlocksAreFound) {
-          throw new ExecutionException(
-              new CommandLine(this),
-              "Partial export due to inability to recover all requested blocks");
+        switch (format) {
+          case RLP:
+            exportRlpFormat(controller);
+            break;
+          default:
+            throw new ParameterException(
```
shemnon: Nice, this future proofs us against random formats being added to the enum and by effect being exposed to the CLI.

pantheon/src/test/java/tech/pegasys/pantheon/chainexport/RlpBlockExporterTest.java
```diff
@@ -70,42 +64,46 @@ public static void initPantheonController() throws IOException {
...
+  //
+  //    assertThat(exportResult.blocks).contains(blockFromBlockchain);
+  //  }
+  //
+  //  @Test
+  //  public void callingBlockExporterWithStartBlockAndBlockShouldReturnSeveralBlocks()
+  //      throws Exception {
+  //
+  //    final long startBlock = 0L;
+  //    final long endBlock = 1L;
```
abdelhamidbakhta: Why are those tests commented ?

pantheon/src/main/java/tech/pegasys/pantheon/cli/subcommands/blocks/BlocksSubCommand.java
```diff
@@ -241,71 +241,127 @@ private MiningParameters getMiningParameters() {
...
+      if (maybeStartBlock.isPresent() && maybeEndBlock.isPresent()) {
+        if (endBlock <= startBlock) {
+          throw new CommandLine.ParameterException(
+              spec.commandLine(),
+              "Parameter --end-block ("
+                  + endBlock
+                  + ") must be greater start block ("
+                  + startBlock
+                  + ").");
+        }
```
shemnon: We don't know the state of the blockchain here correct?  So we can't assert startblock is in the chain?  I think this could be a user error we can detect (somehow) and we should provide the user feedback on.  Perhaps we could count the number of blocks exported and return it as part of the `exportBlocks` method, and log a warning or output the console if no blocks were exported.

pantheon/src/main/java/tech/pegasys/pantheon/chainexport/RlpBlockExporter.java
mbaxter: Yep, this matches the format the `blocks import` accepts

pantheon/src/test/java/tech/pegasys/pantheon/chainexport/RlpBlockExporterTest.java
mbaxter: old commit

pantheon/src/main/java/tech/pegasys/pantheon/cli/subcommands/blocks/BlocksSubCommand.java
mbaxter: Right, no one wants to read binary data from std out :D But yeah, we might need to rethink it for json export.  Although given you can potentially export massive amounts of data, doesn't seem like the worst thing to require a file to export to.

pantheon/src/main/java/tech/pegasys/pantheon/cli/subcommands/blocks/BlocksSubCommand.java
```diff
@@ -241,71 +241,127 @@ private MiningParameters getMiningParameters() {
...
+
+      // Error if data directory is empty
+      Path databasePath =
+          Paths.get(
+              parentCommand.parentCommand.dataDir().toAbsolutePath().toString(),
+              PantheonController.DATABASE_PATH);
+      File databaseDirectory = new File(databasePath.toString());
+      if (!databaseDirectory.exists()
+          || !databaseDirectory.isDirectory()
+          || databaseDirectory.list().length == 0) {
```
RatanRSur: `isDirectory` implies `exists` so `exists` can be removed.

pantheon/src/main/java/tech/pegasys/pantheon/cli/subcommands/blocks/BlocksSubCommand.java
mbaxter: We'll currently log a warning if any blocks in the defined range were missing.  Might be a good idea to error out if no blocks are exported though ... I may follow-up on that in another PR.

pantheon/src/main/java/tech/pegasys/pantheon/cli/subcommands/blocks/BlocksSubCommand.java
RatanRSur: I don't think this is more readable but just for fun, the whole thing can be determined from `list()` because it returns `null` if it's not a directory or doesn't exist :P 
`Optional.ofNullable(databaseDirectory.list()).filter(files -> files.length == 0).isPresent()`

PR #1854, opened 2019-08-14
[PAN-3064] sweep state roots before child nodes

RatanRSur:
## PR description
* Removes state roots before child nodes because state root existence are used to determine if world state is available

* Adds `long removeAccountStateTrieNode(key)` method to `WorldStateStorage.Updater`

## Fixed Issue(s)

https://pegasys1.atlassian.net/secure/RapidBoard.jspa?rapidView=28&modal=detail&selectedIssue=PAN-3064&assignee=5ce472e018de8f0dc3552e9c


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/storage/keyvalue/WorldStateKeyValueStorage.java
```diff
@@ -83,10 +83,15 @@ public Updater updater() {
   }
 
   @Override
-  public long prune(final Predicate<BytesValue> inUseCheck) {
+  public long removeUnless(final Predicate<BytesValue> inUseCheck) {
     return keyValueStorage.removeUnless(inUseCheck);
   }
 
+  @Override
+  public long remove(final BytesValue key) {
```
mbaxter: Should this method be on the `Updater` instead?

services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/KeyValueStorage.java
```diff
@@ -35,6 +35,14 @@ default boolean containsKey(final BytesValue key) throws StorageException {
    */
   Optional<BytesValue> get(BytesValue key) throws StorageException;
 
+  /**
+   * Removes a key/value pair from storage.
+   *
+   * @param key Index into storage
+   * @return 1 if a key was removed, 0 otherwise
+   */
+  long remove(BytesValue key) throws StorageException;
```
mbaxter: If you're modifying this interface, you can quickly add tests across the board by updating `AbstractKeyValueStorageTest`.  But not sure if we actually need this new method, since the `Transaction` already supports removals ...

services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/LimitedInMemoryKeyValueStorage.java
```diff
@@ -76,10 +76,28 @@ public boolean containsKey(final BytesValue key) throws StorageException {
...
+    } finally {
+      lock.unlock();
+    }
+
+    return entriesRemoved;
+  }
+
   @Override
   public long removeUnless(final Predicate<BytesValue> inUseCheck) {
+    long initialSize = storage.size();
```
mbaxter: Side note - should this writeLock be used here?

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/worldstate/MarkSweepPruner.java
```diff
@@ -104,11 +108,26 @@ public void mark(final Hash rootHash) {
...
+    for (long blockNumber = markedBlockNumber - 1; blockNumber >= 0; blockNumber--) {
+      final Hash candidateStateRootHash =
+          blockchain.getBlockHeader(blockNumber).get().getStateRoot();
+
+      if (!worldStateStorage.contains(candidateStateRootHash)) {
+        break;
+      }
+
+      prunedNodeCount += worldStateStorage.remove(candidateStateRootHash);
+      markStorage.remove(candidateStateRootHash);
```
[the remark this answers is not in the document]
RatanRSur: Good catch! I'll add a test case that repros this.

services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/KeyValueStorage.java
RatanRSur: Hmm, ok, I'm leaning towards just using the `Transaction` from `KeyValueStorage` but it looks like `WorldStateKeyValueStorage` doesn't implement `KeyValueStorage`, was this an intentional decision?

RatanRSur: Hmm, ok, I'm leaning towards just using the `Transaction` from `KeyValueStorage` but it looks like `WorldStateKeyValueStorage` doesn't implement `KeyValueStorage`, was this an intentional decision?

mbaxter: `KeyValueStorage` is a layer below `WorldStateKeyValueStorage` (this class is implemented using a `KeyValueStorage`).  I think you'd need to add a method like `removeAccountStateTrieNode` to the `Updater`.

services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/LimitedInMemoryKeyValueStorage.java
mbaxter: (Also worth checking that other implementations of this are thread-safe)

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/worldstate/MarkSweepPrunerTest.java
```diff
@@ -45,26 +58,109 @@ public void shouldMarkAllNodesInCurrentWorldState() {
...
+    when(preSweepHeader.getStateRoot()).thenReturn(preSweepStateRoot);
+
+    pruner.mark(stateRoot);
+    pruner.mark(preSweepStateRoot);
+    pruner.flushPendingMarks();
+    final Set<BytesValue> markedNodes = new HashSet<>(markStorage.keySet());
+
+    // All those new nodes should be removed when we sweep
+    pruner.sweepBefore(1);
+    assertThat(stateStorage.keySet()).containsExactlyInAnyOrderElementsOf(markedNodes);
```
mbaxter: This is a little hard to follow - can you add explicit assertions on the state roots you don't expect to be removed: `assertThat(stateStorage.containsKey(preSweepStateRoot));`

services/kvstore/src/main/java/tech/pegasys/pantheon/services/kvstore/InMemoryKeyValueStorage.java
```diff
@@ -68,8 +75,15 @@ public boolean containsKey(final BytesValue key) throws StorageException {
 
   @Override
   public long removeUnless(final Predicate<BytesValue> inUseCheck) {
```
mbaxter: Can we add tests for this method to `AbstractKeyValueStorageTest`?

ethereum/core/src/test/java/tech/pegasys/pantheon/ethereum/worldstate/MarkSweepPrunerTest.java
RatanRSur: Good point, as it stands this test's scope overlaps with another test's scope right now anyway so it can be narrowed.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/worldstate/MarkSweepPruner.java
```diff
@@ -104,11 +108,30 @@ public void mark(final Hash rootHash) {
...
+      final Hash candidateStateRootHash =
+          blockchain.getBlockHeader(blockNumber).get().getStateRoot();
+
+      if (!worldStateStorage.isWorldStateAvailable(candidateStateRootHash)) {
+        break;
+      }
+
+      if (!markStorage.containsKey(candidateStateRootHash)) {
+        updater.removeAccountStateTrieNode(candidateStateRootHash);
+        prunedNodeCount++;
```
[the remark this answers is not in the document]
RatanRSur: How many transactions per commit is typical? I see that Adrian uses 1000 for marking, could I rename and reuse that value? Should this be exposed to configuration?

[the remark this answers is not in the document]
mbaxter: Yeah, I'd just rename the existing constant and use that.  No reason to expose that externally at this point.  We should just wire things up as we need them (YAGNI).

PR #1862, opened 2019-08-16
[PAN-2842] Blake 2b tweaks

shemnon:
## PR description

* Adjust endianess on byte payload
* f flag can have only 0 or 1
* Fewer successful vector tests
* Tests for failure and edge cases


crypto/src/test/java/tech/pegasys/pantheon/crypto/Blake2bfMessageDigestTest.java
```diff
@@ -27,21 +27,21 @@
   private Blake2bfMessageDigest messageDigest;
 
   // output when input is all 0
-  private byte[] blake2bfAllZero =
+  private static final byte[] BLAKE2F_ALL_ZERO =
```
mbaxter: Where are these test cases coming from?  Is there a spec we can link to?  Would the tests be clearer if kept these values as hexary strings?

crypto/src/main/java/tech/pegasys/pantheon/crypto/Blake2bfMessageDigest.java
```diff
@@ -224,7 +224,7 @@ private int bytesToInt(final byte[] bytes) {
     }
 
     private long bytesToLong(final byte[] bytes) {
-      return Pack.bigEndianToLong(bytes, 0);
+      return Pack.littleEndianToLong(bytes, 0);
```
mbaxter: Pretty bizarre that the payload is a mix of big endian and little endian values :\

crypto/src/test/java/tech/pegasys/pantheon/crypto/Blake2bfMessageDigestTest.java
shemnon: These tests are not as useful as the precompiled test vectors.  Those have more coverage.  As far as I can tell they were calculated from the blake RFC.

crypto/src/main/java/tech/pegasys/pantheon/crypto/Blake2bfMessageDigest.java
shemnon: Agreed.  But that's what some web3/solidity folks asked for in the EIP and got.

PR #1869, opened 2019-08-20
Pruning Command Line Flags

RatanRSur:
## PR description
Adds pruning and pruning configuration dependent flags

## Fixed Issue(s)

https://pegasys1.atlassian.net/secure/RapidBoard.jspa?rapidView=28&modal=detail&selectedIssue=PAN-2878&assignee=5ce472e018de8f0dc3552e9c


pantheon/src/main/java/tech/pegasys/pantheon/Runner.java
```diff
@@ -76,9 +76,7 @@
   public void start() {
     try {
       LOG.info("Starting Ethereum main loop ... ");
-      if (natManager.isPresent()) {
-        natManager.get().start();
-      }
+      natManager.ifPresent(UpnpNatManager::start);
```
abdelhamidbakhta: nice catch

pantheon/src/main/java/tech/pegasys/pantheon/Runner.java
```diff
@@ -76,9 +76,7 @@
   public void start() {
     try {
       LOG.info("Starting Ethereum main loop ... ");
-      if (natManager.isPresent()) {
```
shemnon: Can this go in a different PR?  It's unrelated.

pantheon/src/main/java/tech/pegasys/pantheon/cli/DefaultCommandValues.java
```diff
@@ -43,6 +43,8 @@
   String MANDATORY_NETWORK_FORMAT_HELP = "<NETWORK>";
   String MANDATORY_NODE_ID_FORMAT_HELP = "<NODEID>";
   Wei DEFAULT_MIN_TRANSACTION_GAS_PRICE = Wei.of(1000);
+  long DEFAULT_RETENTION_PERIOD = 1000;
```
shemnon: Should we up this to be at least 1024?  Maybe more?  I would like the default to be higher to ensure sweep doesn't cut past the finality gadget horizon, which starts at a 1024 block minimum.  In a live system it will be functionally more but in a static replay we may get in trouble.

RatanRSur: How about we up it to 1024 for now and then see what the state size hovers around when we do our tests to see if we want to increase it further?

PR #1871, opened 2019-08-21
[PAN-3073] push ordered transaction commitment history to enclave for privacy group modification

iikirilov:
* TODO: acceptance test

## PR description

add push history to enclave endpoint and use in privacy precompile

## Fixed Issue(s)
[PAN-3073]

[PAN-3073]: https://pegasys1.atlassian.net/browse/PAN-3073


enclave/src/integration-test/java/tech/pegasys/pantheon/enclave/EnclaveTest.java
```diff
@@ -164,6 +165,33 @@ public void testCreateFindDeleteFindPrivacyGroup() throws Exception {
...
+  @Test
+  public void testPushToHistory() throws Exception {
+    final List<String> publicKeys = testHarness.getPublicKeys();
+    final String name = "testName";
+    final String description = "testDesc";
+    final CreatePrivacyGroupRequest privacyGroupRequest =
+        new CreatePrivacyGroupRequest(
+            publicKeys.toArray(new String[0]), publicKeys.get(0), name, description);
+
+    final PrivacyGroup privacyGroupResponse = enclave.createPrivacyGroup(privacyGroupRequest);
```
pinges: rename to privacyGroup?

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/MainnetTransactionProcessor.java
```diff
@@ -231,6 +231,7 @@ public Result processTransaction(
               .blockchain(blockchain)
               .worldState(worldUpdater.updater())
               .initialGas(gasAvailable)
+              .transactionHash(transaction.hash())
```
pinges: Why do we need the transaction hash for main-net?

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/MainnetTransactionProcessor.java
```diff
@@ -262,6 +263,7 @@ public Result processTransaction(
...
```
pinges: s.a.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/vm/MessageFrame.java
```diff
@@ -1026,6 +1045,7 @@ private void validate() {
       checkState(messageFrameStack != null, "Missing message frame message frame stack");
       checkState(worldState != null, "Missing message frame world state");
       checkState(initialGas != null, "Missing message frame initial getGasRemaining");
+      checkState(transactionHash != null, "Missing message frame transactionHash");
```
pinges: There are a few cases where the MessageFrame.builder() is used that do not set the transactionHash, e.g. in AbstractCreateOperation.java line 105). These builders should fail because we are checking for null here. Does that mean that the code where these builders are used is dead code, or does it mean that we do not have tests that executes these builders, but they could be used and would fail?

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/precompiles/privacy/PrivacyPrecompiledContract.java
```diff
@@ -144,6 +145,22 @@ public BytesValue compute(final BytesValue input, final MessageFrame messageFram
...
           disposablePrivateState.rootHash(),
           privacyGroupId);
+
+      try {
+        enclave.pushToHistory(
+            new PushToHistoryRequest(
+                BytesValues.asBase64String(privacyGroupId),
+                messageFrame.getTransactionHash().toString(),
+                key));
+      } catch (Exception e) {
```
pinges: e can be final

enclave/src/main/java/tech/pegasys/pantheon/enclave/Enclave.java
```diff
@@ -68,6 +69,10 @@ public ReceiveResponse receive(final ReceiveRequest content) throws Exception {
     return executePost(buildPostRequest(ORION, content, "/receive"), ReceiveResponse.class);
   }
 
+  public boolean pushToHistory(final PushToHistoryRequest content) throws Exception {
+    return executePost(buildPostRequest(JSON, content, "/pushToHistory"), Boolean.class);
```
pinges: This is not available yet in Orion ...

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/MainnetTransactionProcessor.java
iikirilov: we don't

enclave/src/main/java/tech/pegasys/pantheon/enclave/Enclave.java
iikirilov: https://github.com/PegaSysEng/orion/pull/283

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/vm/MessageFrame.java
iikirilov: true - The main issue is in `VMReferenceTest` - because the transaction hash is not required for running the spec tests I have passed `Hash.EMPTY`.

The other option right now is to make the transaction hash optional - i.e. don't check if it is `null` - I don't like this personally as it is not consistent but there is no immediate negative impact.

enclave/src/main/java/tech/pegasys/pantheon/enclave/Enclave.java
iikirilov: this is also why CI is currently failing - which is a good thing so we cannot prematurely merge this :)

PR #1864, opened 2019-08-19
[PAN-2941] Refactor privacy acceptance tests

iikirilov:
## PR description

* Restructures the Privacy acceptance tests to follow the dsl and structure of Mainnet acceptance tests.

* Updates to web3j 4.5.0

* Renames some configuration classes in the acceptance to more suitable name - mainly `PantheonFactoryConfiguration` -> `PantheonNodeConfiguration` and similar. Considering this is not a configuration for a factory, but rather for the node produced by the factory that is being used in tests. This removes some confusion.

* Updates EthSigner to 0.3.0

* Fixes Ibft privacy tests - still in flight

**N.B.** There is 1 known failing test

## Fixed Issue(s)
[PAN-2941][PAN-3026][PAN-2643]

[PAN-2941]: https://pegasys1.atlassian.net/browse/PAN-2941
[PAN-3026]: https://pegasys1.atlassian.net/browse/PAN-3026
[PAN-2643]: https://pegasys1.atlassian.net/browse/PAN-2643

PR #1872, opened 2019-08-21
[PAN-2928] return null private transaction receipt instead of error

iikirilov:
## PR description

return null instead of error if private transaction does not exists

## Fixed Issue(s)
[PAN-2928]

[PAN-2928]: https://pegasys1.atlassian.net/browse/PAN-2928


ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/privacy/eea/EeaGetTransactionReceipt.java
```diff
@@ -102,8 +102,7 @@ public JsonRpcResponse response(final JsonRpcRequest request) {
       privacyGroupId = receiveResponse.getPrivacyGroupId();
     } catch (Exception e) {
       LOG.error("Failed to fetch transaction from Enclave with error " + e.getMessage());
-      return new JsonRpcSuccessResponse(
-          request.getId(), JsonRpcError.PRIVATE_TRANSACTION_RECEIPT_ERROR);
+      return new JsonRpcSuccessResponse(request.getId(), null);
```
rain-on: Should the log error be removed? Are there any exceptions which actually are errors?

iikirilov: There are no errors - an exception here means you do not have access to the private transaction you are trying to get the receipt for.

iikirilov:
This PR is blocked by acceptance test problems that have been fixed in #1864 

As this PR does not block anything itself it will not be patched

ethereum/jsonrpc/src/test/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/privacy/methods/eea/EeaGetTransactionReceiptTest.java
```diff
@@ -108,80 +106,103 @@
...
-
-    doReturn(privateTransactionStorage).when(privacyParameters).getPrivateTransactionStorage();
-    when(privateTransactionStorage.getEvents(any(Bytes32.class)))
-        .thenReturn(Optional.of(mockLogList));
-    when(privateTransactionStorage.getOutput(any(Bytes32.class)))
-        .thenReturn(Optional.of(mockBytesValue));
+    when(privacyParameters.getPrivateTransactionStorage()).thenReturn(privateTransactionStorage);
+    when(privateTransactionStorage.getEvents(any(Bytes32.class))).thenReturn(Optional.empty());
+    when(privateTransactionStorage.getOutput(any(Bytes32.class))).thenReturn(Optional.empty());
+  }
```
antonydenyer: LGTM

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/privacy/methods/eea/EeaGetTransactionReceipt.java
```diff
@@ -101,9 +100,7 @@ public JsonRpcResponse response(final JsonRpcRequest request) {
       privateTransaction = PrivateTransaction.readFrom(bytesValueRLPInput);
       privacyGroupId = receiveResponse.getPrivacyGroupId();
     } catch (Exception e) {
-      LOG.error("Failed to fetch transaction from Enclave with error " + e.getMessage());
-      return new JsonRpcSuccessResponse(
-          request.getId(), JsonRpcError.PRIVATE_TRANSACTION_RECEIPT_ERROR);
+      return new JsonRpcSuccessResponse(request.getId(), null);
```
rain-on: _Could_ any reasons for the Exception be represented as an error? I.e. if the enclave is offline, is that a successful response? (I agree, the user probably can't do much about it ...)

ethereum/jsonrpc/src/test/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/privacy/methods/eea/EeaGetTransactionReceiptTest.java
```diff
@@ -108,80 +106,103 @@
...
+    final JsonRpcSuccessResponse response =
+        (JsonRpcSuccessResponse) eeaGetTransactionReceipt.response(request);
+    final PrivateTransactionReceiptResult result =
+        (PrivateTransactionReceiptResult) response.getResult();
+
+    assertThat(result).isEqualToComparingFieldByField(expectedResult);
+  }
+
+  @Test
+  public void returnNullIfPMTDoesNotExist() {
```
rain-on: Maybe worth renaming - eg. enclaveThrowingResultsInSuccessButNullResponse?

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/privacy/methods/eea/EeaGetTransactionReceipt.java
rain-on: btw - "throws Exception" appears unrequired on getReceiveResponseFromEnclave

ethereum/jsonrpc/src/test/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/privacy/methods/eea/EeaGetTransactionReceiptTest.java
iikirilov: actually i have mixed up the method names here

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/privacy/methods/eea/EeaGetTransactionReceipt.java
iikirilov: It _could_ be represented as an error. The reason for this PR is to make the execution the same as `eth_getTransactionReceipt`, which essentially gets information from the public storage. From looking at a `RocksDbKeyValueStorage.get(...)` it throws a `StorageException` (a `RuntimeException`) which is analogous to "enclave is offline". This would return an `INTERNAL_ERROR` response.

I will make the change.

enclave/src/main/java/tech/pegasys/pantheon/enclave/Enclave.java
```diff
@@ -104,7 +104,7 @@ private Request buildPostRequest(
       response = client.newCall(request).execute();
       responseBody = response.body().string();
     } catch (final IOException e) {
-      throw new EnclaveException("Failed to contact Enclave", e);
+      throw new RuntimeException("Failed to contact Enclave", e);
```
iikirilov: Not sure about this change. `EnclaveException` extends `IllegalArgumentException` which is a little misleading as this is a connection error.

If this is not ok let me know.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/privacy/methods/eea/EeaGetTransactionReceipt.java
```diff
@@ -92,18 +92,19 @@ public JsonRpcResponse response(final JsonRpcRequest request) {
...
           new BytesValueRLPInput(BytesValues.fromBase64(receiveResponse.getPayload()), false);
 
       privateTransaction = PrivateTransaction.readFrom(bytesValueRLPInput);
       privacyGroupId = receiveResponse.getPrivacyGroupId();
-    } catch (Exception e) {
-      LOG.error("Failed to fetch transaction from Enclave with error " + e.getMessage());
-      return new JsonRpcSuccessResponse(
-          request.getId(), JsonRpcError.PRIVATE_TRANSACTION_RECEIPT_ERROR);
+    } catch (final EnclaveException e) {
+      if (e.getMessage() != null && e.getMessage().equals("EnclavePayloadNotFound")) {
```
rain-on: Using the message in an exception as the indicator of a given problem can be dangerous - the message is really meant for human consumption - if you need to do something like this, its probably better if the exception contained an error code explicitly defining the failure.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/privacy/methods/eea/EeaGetTransactionReceipt.java
```diff
@@ -92,18 +92,19 @@ public JsonRpcResponse response(final JsonRpcRequest request) {
...
+        return new JsonRpcSuccessResponse(request.getId(), null);
+      }
+      throw e;
```
rain-on: Should this function throw - or should it return a suitable JsonRpcError - i.e. handlers above this won't be able to perform RPC-specific handling of the error.

iikirilov: Yes - it will be handled the same way as if `eth_getTransactionReceipt` failed to read from rocks db.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/privacy/methods/eea/EeaGetTransactionReceipt.java
iikirilov: yes - I have used the correct converter here

enclave/src/main/java/tech/pegasys/pantheon/enclave/Enclave.java
rain-on: 👍 Worse things have/will happen.

PR #1878, opened 2019-08-22
[PIE-1804] Tweak JSON import format

mbaxter:
## PR description
This PR makes a few small changes to the JSON import format:
* Allow "comment" fields for documentation
* Use "secretKey" as the field name for private keys to match the reference test json format

### DIFF
--- a/pantheon/src/main/java/tech/pegasys/pantheon/chainimport/internal/BlockData.java
+++ b/pantheon/src/main/java/tech/pegasys/pantheon/chainimport/internal/BlockData.java
@@ -28,8 +28,10 @@ import java.util.Optional;
 import java.util.stream.Stream;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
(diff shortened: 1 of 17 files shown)


pantheon/src/main/java/tech/pegasys/pantheon/chainimport/internal/BlockData.java
```diff
@@ -28,8 +28,10 @@
 import java.util.stream.Stream;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
+@JsonIgnoreProperties("comment")
```
shemnon: Would `@JsonIgnoreProperties(ignoreUnknown=true)` work here?  Or are you only explicitly wanting to allow the `comment` field?

mbaxter: That would work, but I think it's better to explicitly ignore.  That way, it will be obvious if some optional field is misspelled or something.

PR #1879, opened 2019-08-23
[WIP][PAN-2030] Add formating and color to logging in CLI

pscott:
## PR description
Add formating and color to logging in CLI

## Fixed Issue(s)
fixes #2030

### DIFF
--- a/pantheon/src/main/resources/log4j2.xml
+++ b/pantheon/src/main/resources/log4j2.xml
@@ -6,7 +6,11 @@
 
   <Appenders>
     <Console name="Console" target="SYSTEM_OUT">
-      <PatternLayout pattern="%d{yyyy-MM-dd HH:mm:ss.SSSZZZ} | %-5level | %t | %c{1} | %msg%n" />    </Console>
+      <PatternLayout pattern="%style{%d{yyyy-MM-dd HH:mm:ss.SSSZZZ}}{dim}
+      %style{|}{dim} %highlight{%-5level}{INFO=cyan, ERROR=bright red, WARN=bright yellow}
+      %style{|}{dim} %style{%t}{cyan}
+      %style{|}{dim} %style{%c{1}}{cyan}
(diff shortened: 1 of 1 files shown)


NicolasMassart:
Thanks @pscott, I like your way to start contributing and welcome to this community.
I assigned myself to your PR as I will follow it and merge it once you will finish the job (when you remove the WIP and tell me I can merge). I also assigned two reviewers for this PR, including @jakehaugen who wrote the Jira ticket.

ajsutton:
We'll need to think carefully about this - currently Pantheon only logs to stdout and then when run in a server environment that is piped to a file.  We need to ensure the file doesn't get filled with junk characters making the logs unreadable.  I suspect that will require detecting if output is actually going to a console and only using color in that case (possibly along with a CLI option to explicitly enable or disable color).

NicolasMassart:
> I suspect that will require detecting if output is actually going to a console and only using color in that case (possibly along with a CLI option to explicitly enable or disable color).

This is indeed the goal of https://pegasys1.atlassian.net/browse/PAN-2021 which is linked in https://pegasys1.atlassian.net/browse/PAN-2030

ajsutton:
PAN-2021 doesn't include the auto-detect requirement and we need to land both at the same time - we can't break logging by filling it full of color codes.  We could default to no colours with an option to enable before adding auto detection, but we ended o either auto detect or default to off.

@pscott It's great that you've picked this up, sorry that all the details aren't entirely sorted but we'd love to work through them with you.

NicolasMassart:
@pscott we are currently discussing changing https://pegasys1.atlassian.net/browse/PAN-2021 to include a default to no color. Then this PR may have to include this change to make sure we remain compatible with the way terminal output works today. I'll keep you informed when we confirm the decision. Thanks for being so patient.

NicolasMassart:
@pscott I updated [PAN-2021](https://pegasys1.atlassian.net/browse/PAN-2021) to reflect the off by default requirement. Would you like to work on it too as it's very closely related to this one? You could even make the changes for both in this PR IMO.

pscott:
Yes I believe everything should fit in this PR. WIP!

PR #1880, opened 2019-08-23
[PIE-1805] Simplify json rpc spec test setup

mbaxter:
## PR description
Simplify JSON-RPC spec test setup by walking resource directories to search for spec files rather than explicitly including every single spec file.


ethereum/jsonrpc/src/test/java/tech/pegasys/pantheon/ethereum/jsonrpc/AbstractJsonRpcHttpBySpecTest.java
```diff
@@ -29,22 +40,64 @@
...
-        Resources.toString(
-            AbstractJsonRpcHttpBySpecTest.class.getResource(testSpecFile), Charsets.UTF_8);
+  /**
+   * Searches the provided subdirectories for spec files containing test cases, and returns
+   * parameters required to run these tests.
+   *
+   * @param subDirectoryPaths The subdirectories to search.
+   * @return Parameters for this test which will contain the name of the test and the url of the
+   *     spec file to run.
+   */
```
shemnon: getParams seems to be a confusing name.  I understand that it is going to a JUnit parameterized tests, but on it's own it's unclear.  Perhaps `findSpecTests`?

PR #1881, opened 2019-08-23
[PIE-1805] Add test cases for trace_replayBlockTransactions

mbaxter:
## PR description
Update chain data and add test cases for `trace_replayBlockTransactions` tests.  Also:
* Update `BlockchainSetupUtil` with a constructor that accepts arbitrary genesis and block data files
* Update tests to be agnostic as to JSON-RPC response key ordering


ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/methods/TraceReplayBlockTransactions.java
```diff
@@ -90,4 +96,9 @@ private JsonNode formatTrace(
     ObjectMapper mapper = new ObjectMapper();
     return mapper.createObjectNode();
   }
+
+  private Object emptyResult() {
+    ObjectMapper mapper = new ObjectMapper();
```
abdelhamidbakhta: Suggestion: final

PR #1882, opened 2019-08-26
[PAN-2824] Blake validation errors are hard errors

shemnon:
## PR description

Change the blake validation errors from soft errors (empty data) to hard errors
(precompile failed).


ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/precompiles/BLAKE2BFPrecompileContract.java
```diff
@@ -50,10 +50,10 @@ public Gas gasRequirement(final BytesValue input) {
   @Override
   public BytesValue compute(final BytesValue input, final MessageFrame messageFrame) {
     if (input.size() != MESSAGE_LENGTH_BYTES) {
-      return BytesValue.EMPTY;
+      return null;
```
abdelhamidbakhta: (Discussion) What about creating an explicit error and throw a Runtime exception ? What would be the resulting error when you return null ? Is it something meaningful ?

abdelhamidbakhta: Ok sounds good.

PR #1886, opened 2019-08-27
Implement trace replay block transactions trace option

abdelhamidbakhta:
## PR description

- iterate through `TraceFrame` and transform to parity style
- create `FlatTrace`s from aggregated `TraceFrame`s included in `TransactionTrace` object
- detect `CALL` opcode to retrieve contract address required in parity format
- detect `RETURN` opcode to close a subtrace
- use https://github.com/mbaxter/json-rpc-test-cases to generate test cases


abdelhamidbakhta:
> The trace assembly code looks fine, but unless we are in Kotlin w/ data objects or using something like Google's autovalues I feel the overhead of writing POJOs + builders will slow down development time. We can put the data directly into JSON objects and return that just as well since these APIs are used just to create JSON-RPC replys.

I would stay with POJO because i will have to reuse fields, for example to compute the cumulative gas cost of each trace. Also i think some of the objects i created might be usable for others trace options(`vmTrace` and `stateDiff`). If you don't mind i will stay like this to focus on the business logic. When the feature will be working completely i would be happy to do some refactoring if needed. Does it makes sense ?

abdelhamidbakhta:
> ## PR description
> * start business logic implementation
> * iterate through `TraceFrame` and transform to parity style
> * detect `CALL` opcode to retrieve contract address required in parity format
> 
> ## Fixed Issue(s)

- compute gas used for each sub trace
- detect `RETURN` opcode
    - to get `ouput` value from `memory`
    - detect end of subtrace

abdelhamidbakhta:
Known bugs:
- gas computation (`gas` field in `action` node)
- gas used computation (`gasUsed` field in `result` node)
- `input` values are wrong with several levels deep
- `traceAddress` does not work with several levels deep

ethereum/jsonrpc/src/test/resources/tech/pegasys/pantheon/ethereum/jsonrpc/trace/trace_replayBlockTransactions_pending.json
```diff
@@ -5,13 +5,13 @@
...
...pe":"call"},{"action":{"callType":"call","from":"0x0040000000000000000000000000000000000000","gas":"0xffaa71","input":"0x0000000000000000000000000040000000000000000000000000000000000000","to":"0x0040000000000000000000000000000000000000","value":"0x0"},"result":{"gasUsed":"0x6","output":"0xf000000000000000000000000000000000000000000000000000000000000002"},"subtraces":1,"traceAddress":[0],"type":"call"},{"action":{"callType":"call","from":"0x0040000000000000000000000000000000000000","gas":"0xfba8e4","input":"0x0000000000000000000000000030000000000000000000000000000000000000","to":"0x0040000000000000000000000000000000000000","value":"0x0"},"result":{"gasUsed":"0x6","output":"0xf000000000000000000000000000000000000000000000000000000000000002"},"subtraces":1,"traceAddress":[0,1],"type":"call"},{"action":{"callType":"call","from":"0x0040000000000000000000000000000000000000","gas":"0xf7b763","input":"0xf000000000000000000000000000000000000000000000000000000000000001","to":"0x0030000000000000000000000000000000000000","value":"0x0"},"result":{"gasUsed":"0x1b","output":"0xf000000000000000000000000000000000000000000000000000000000000002"},"subtraces":0,"traceAddress":[0,1,2],"type":"call"}]}]
```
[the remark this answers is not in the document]
abdelhamidbakhta: sure.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/results/tracing/Trace.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+ * an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ */
+package tech.pegasys.pantheon.ethereum.jsonrpc.internal.results.tracing;
+
+public interface Trace {
+  @FunctionalInterface
+  interface ResultWriter {
```
shemnon: Style: why an interface in an interface?  I would expect just one interface `TraceResultWriter`

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/results/tracing/FlatTraceGenerator.java
```diff
@@ -0,0 +1,296 @@
...
+              flatTraceBuilder.type("create");
+            },
+            // set output otherwise
+            () ->
+                resultBuilder.output(
+                    traceFrame.getMemory().isPresent() && traceFrame.getMemory().get().length > 0
+                        ? traceFrame.getMemory().get()[0].toString()
+                        : "0x"));
+        ctx.markAsReturned();
+        continueToPollContexts = false;
```
shemnon: I wouldn't use a sentinel boolean I would just use a `break` statement

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/results/tracing/FlatTraceGenerator.java
```diff
@@ -0,0 +1,296 @@
...
+      final TraceFrame traceFrame,
+      final Optional<String> smartContractAddress,
+      final Deque<FlatTrace.Context> tracesContexts) {
+    final Deque<FlatTrace.Context> polledContexts = new ArrayDeque<>();
+    FlatTrace.Context ctx;
+    boolean continueToPollContexts = true;
+    // find last non returned transactionTrace
+    while (continueToPollContexts && (ctx = tracesContexts.pollLast()) != null) {
+      polledContexts.addFirst(ctx);
+      if (!ctx.isReturned()) {
```
shemnon: I think a guard block instead of a nested if would read better since the if covers the rest of the while block.

```suggestion
      if (ctx.isReturned()) {
        continue;
      }
```

Then remove the matching brace and un indent the block.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/results/tracing/FlatTraceGenerator.java
```diff
@@ -0,0 +1,296 @@
...
+        Action.createCallAction(
+            transactionTrace.getTransaction(),
+            lastContractAddress,
+            contractCallAddress,
+            traceFrame,
+            traceFrameAfterCall.getGasRemaining());
+
+    final long gasCost = cumulativeGasCost.longValue();
+    // retrieve the previous transactionTrace context
+    Optional.ofNullable(tracesContexts.peekLast())
```
shemnon: This use of optional feels a bit clunky and creates an extra optional object.  Would this work better?
```java
    if (!traceContexts.isEmpty()) {
        FlatTrace.Context previousContext = traceContexts.peekLast();
        // increment sub traces counter of previous transactionTrace
        previousContext.getBuilder().incSubTraces();
        // set gas cost of previous transactionTrace
        previousContext
            .getBuilder()
            .getResultBuilder()
            .orElse(Result.builder())
            .gasUsed(Gas.of(gasCost).toHexString());
      }

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/results/tracing/Trace.java
abdelhamidbakhta: `Trace` interface is the common type for `FlatTrace` and the other types of traces we will add for `vmTrace` and `stateDiff` options. So i'll keep `Trace` interface but as you suggest i'll move `ResultWriter`to a separate file.

abdelhamidbakhta: Does that make sense ?

abdelhamidbakhta:
> There are several classes that have both setter methods and a builder. I wouldn't expect both. Either the object is mutable and setters are used or the object is immutable and a builder is used.

You right, i have removed setters and use only builders.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/results/tracing/FlatTraceGenerator.java
abdelhamidbakhta: Done.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/results/tracing/Trace.java
abdelhamidbakhta: Done.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/results/tracing/FlatTraceGenerator.java
abdelhamidbakhta: Nice suggestion. Done.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/results/tracing/FlatTraceGenerator.java
abdelhamidbakhta: Done.

ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/results/tracing/Trace.java
abdelhamidbakhta: Totally agree. i'll do that

abdelhamidbakhta: Done.

PR #1898, opened 2019-09-02
[PIE-1859] Extracted Metrics interfaces to plugins-api.

mark-terry:
## PR description
The Metrics interfaces have been extracted to the plugin-api.

Most of the changes are updates to import locations.  

The more noteworthy code changes are:

- A new interface `PantheonMetricsSystem` that  extends `MetricsSystem` has been added to include the streamObservations methods.
- `plugin-api` has been added as a module dependency in a few locations
- `UnsupportedOperationException`s have been added in `PantheonEventsImpl` where new interface methods had not yet been implemented.

## Fixed Issue(s)

Addresses PIE-1859


metrics/core/src/main/java/tech/pegasys/pantheon/metrics/PantheonMetricsSystem.java
```diff
@@ -12,11 +12,14 @@
...
 
-import java.util.Optional;
+import tech.pegasys.pantheon.plugin.services.MetricsSystem;
+import tech.pegasys.pantheon.plugin.services.metrics.MetricCategory;
 
-public interface MetricCategory {
+import java.util.stream.Stream;
 
-  String getName();
+public interface PantheonMetricsSystem extends MetricsSystem {
```
ajsutton: Can we call this something other than `Pantheon...`.  Maybe `ObservableMetricsSystem` or even `MetricsSource`.

PR #1901, opened 2019-09-03
[WIP] 3116 valid number rpc params

pscott:
## PR description
WIP:

- [x] Add assertLength method to JsonRpcRequest
- [ ] Use assertLength in every RPC method to check whether the number of parameters passed in is correct.
- [ ] Add tests for every method to ensure that behaviour is as expected when too many parameters are passed in.

## Fixed Issue(s)
Fixes #3116

### DIFF
--- a/ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/JsonRpcRequest.java
+++ b/ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/JsonRpcRequest.java
@@ -12,6 +12,7 @@
  */
 package tech.pegasys.pantheon.ethereum.jsonrpc.internal;
 
+import tech.pegasys.pantheon.ethereum.jsonrpc.internal.exception.InvalidJsonRpcParameters;
 import tech.pegasys.pantheon.ethereum.jsonrpc.internal.exception.InvalidJsonRpcRequestException;
 
 import java.util.Arrays;
@@ -67,6 +68,15 @@ public class JsonRpcRequest {
   @JsonGetter("params")
   public Object[] getParams() {
     return params;
+  }
+
+  @JsonIgnore
+  public void assertLength(final int expectedLength) {
+    final int length = getParamLength();
+    if (length > expectedLength) {
+      throw new InvalidJsonRpcParameters(
+          "Too many parameters, expected " + expectedLength + ", got " + length + " .");
+    }
   }
 
   @JsonIgnore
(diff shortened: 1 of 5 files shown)


ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/internal/JsonRpcRequest.java
```diff
@@ -69,6 +70,15 @@ public String getVersion() {
     return params;
   }
 
+  @JsonIgnore
+  public void assertMaxLength(final int expectedLength) {
```
mbaxter: I'd suggest reworking the structure of this code a bit by:
- Adding a method `int getMaxAllowedParams()` to the `JsonRpcMethod` interface
- Adding a `default` method `assertValidRequest(JsonRpcRequest request)` to `JsonRpcMethod` that throws an `InvalidJsonRcpParameters` exception as you have here if there are too many params
- Validating the request by calling `jsonRpcMethod.assertValidRequest(request)` before executing [`jsonRpcMethod.response(request)` in `JsonRpcHttpService`](https://github.com/PegaSysEng/pantheon/blob/fdfa35443bfe26a01919c358b74f1184d41c9b66/ethereum/jsonrpc/src/main/java/tech/pegasys/pantheon/ethereum/jsonrpc/JsonRpcHttpService.java#L497)

This will make it straightforward to consistently enforce this validation without expecting every method to run these checks manually.  And it allows methods to override the default validation if necessary.

pscott: Thank you so much! Definitely a better code structure, I will do my best to implement it.
Quick question : should I still manually add a tooManyParams test for every method, or should I just add one test for one method and "assume" that it will be correct for other methods ?

mbaxter: Seems kind of painful to handcraft tests for every method.  But if there was some way to "automatically" test each method, that would be cool.  One idea:
- Write a parameterized test ([example](https://github.com/PegaSysEng/pantheon/blob/fdfa35443bfe26a01919c358b74f1184d41c9b66/consensus/clique/src/test/java/tech/pegasys/pantheon/consensus/clique/headervalidationrules/VoteValidationRuleTest.java#L30-L42)) that runs for every method and constructs a request with 1 too many parameters, then verifies that the expected exception is thrown

Otherwise, I think we can probably get away with just adding a single test to `JsonRpcHttpServiceTest`.

PR #1915, opened 2019-09-09
[PAN-3126] add an override facility for genesis configs

shemnon:
## PR description

Add a generalized genesis config override facility for command line use.  Only useful for string or integer config values, not object values.

Sample use - enable Istanbul fork on block 99 million
`--override-genesis-config=istanbulFork=99000000`

Overriding a value with an empty string unsets the value

Sample use - disable Istanbul fork by un-setting `istanbulFork`: 
`--override-genesis-config=istanbulFork=`

### DIFF
--- a/config/src/main/java/tech/pegasys/pantheon/config/JsonGenesisConfigOptions.java
+++ b/config/src/main/java/tech/pegasys/pantheon/config/JsonGenesisConfigOptions.java
@@ -231,7 +231,7 @@ public class JsonGenesisConfigOptions implements GenesisConfigOptions {
       final String value = configOverrides.get(key);
       return value == null || value.isEmpty()
           ? OptionalLong.empty()
-          : OptionalLong.of(Long.valueOf(configOverrides.get(key)));
+          : OptionalLong.of(Long.valueOf(configOverrides.get(key), 10));
     } else {
       return JsonUtil.getLong(configRoot, key);
     }
@@ -242,7 +242,7 @@ public class JsonGenesisConfigOptions implements GenesisConfigOptions {
       final String value = configOverrides.get(key);
       return value == null || value.isEmpty()
           ? OptionalInt.empty()
-          : OptionalInt.of(Integer.valueOf(configOverrides.get(key)));
+          : OptionalInt.of(Integer.valueOf(configOverrides.get(key), 10));
     } else {
       return JsonUtil.getInt(configRoot, key);
     }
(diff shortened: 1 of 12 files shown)


config/src/main/java/tech/pegasys/pantheon/config/JsonGenesisConfigOptions.java
```diff
@@ -207,7 +217,45 @@ public OptionalInt getEvmStackSize() {
...
+    }
+  }
+
   private OptionalLong getOptionalLong(final String key) {
-    return JsonUtil.getLong(configRoot, key);
+    if (configOverrides.containsKey(key)) {
+      final String value = configOverrides.get(key);
+      return value == null || value.isEmpty()
+          ? OptionalLong.empty()
+          : OptionalLong.of(Long.valueOf(configOverrides.get(key)));
```
mbaxter: Should we explicitly specify the `radix` here?

config/src/main/java/tech/pegasys/pantheon/config/JsonGenesisConfigOptions.java
```diff
@@ -207,7 +217,45 @@ public OptionalInt getEvmStackSize() {
...
+      return JsonUtil.getLong(configRoot, key);
+    }
+  }
+
+  private OptionalInt getOptionalInt(final String key) {
+    if (configOverrides.containsKey(key)) {
+      final String value = configOverrides.get(key);
+      return value == null || value.isEmpty()
+          ? OptionalInt.empty()
+          : OptionalInt.of(Integer.valueOf(configOverrides.get(key)));
```
mbaxter: Should we explicitly specify the `radix` here?

config/src/test/java/tech/pegasys/pantheon/config/GenesisConfigFileTest.java
```diff
@@ -214,6 +215,59 @@ public void acceptComments() {
     // Unfortunately there is no good (non-flakey) way to assert logs.
   }
 
+  @Test
+  public void testOverridePresent() {
+    final GenesisConfigFile config = GenesisConfigFile.development();
+    final int bigBlock = 999_999_999;
+    final String bigBlockString = Integer.toString(bigBlock);
+    final Map<String, String> override = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
```
mbaxter: Can we add some tests that use a case-sensitive `Map` with various casing strategies for the keys to verify that values are replaced in a case-insensitive way?

shemnon: Done

config/src/main/java/tech/pegasys/pantheon/config/JsonGenesisConfigOptions.java
shemnon: Done

config/src/main/java/tech/pegasys/pantheon/config/JsonGenesisConfigOptions.java
shemnon: Done

PR #1916, opened 2019-09-09
[PAN-3122] Allow private transactions to access public state

josh-richardson:
## PR description
This allows private transactions to access public state. Still a draft, this description needs to be updated.


acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/web3j/privacy/PrivateContractPublicStateAcceptanceTest.java
```diff
@@ -0,0 +1,164 @@
...
+  public void mustAllowAccessToPublicStateFromPrivateTx() throws Exception {
+    final EventEmitter publicEventEmitter =
+        minerNode
+            .getPantheon()
+            .execute((contractTransactions.createSmartContract(EventEmitter.class)));
+
+    final TransactionReceipt receipt = publicEventEmitter.store(BigInteger.valueOf(12)).send();
+    assertNotNull(receipt);
+
+    CrossContractReader reader =
```
iikirilov: make local variable final

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/ReadOnlyMutableAccount.java
```diff
@@ -0,0 +1,115 @@
...
+    return mutableAccount.getStorageValue(key);
+  }
+
+  @Override
+  public UInt256 getOriginalStorageValue(final UInt256 key) {
+    return mutableAccount.getOriginalStorageValue(key);
+  }
+
+  @Override
+  public NavigableMap<Bytes32, AccountStorageEntry> storageEntriesFrom(
```
iikirilov: We need to double-check this

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/MainnetContractCreationProcessor.java
```diff
@@ -113,6 +114,9 @@ public void start(final MessageFrame frame) {
           "Contract creation error: account as already been created for address {}",
           frame.getContractAddress());
       frame.setState(MessageFrame.State.EXCEPTIONAL_HALT);
+    } else if (contract instanceof ReadOnlyMutableAccount) {
```
iikirilov: I don't think we need this. Public contracts created from private contract calls would go through a child message frame

acceptance-tests/src/test/java/tech/pegasys/pantheon/tests/web3j/privacy/PrivateContractPublicStateAcceptanceTest.java
```diff
@@ -0,0 +1,164 @@
...
+
+    var transactionReceipt =
+        (PrivateTransactionReceipt)
+            privateReader.deployRemote(publicReader.getContractAddress()).send();
+
+    assertEquals(0, transactionReceipt.getLogs().size());
+  }
+
+  @Test
+  public void privateContractMustNotBeAbleToCallPublicContractWhichSelfDestructs()
```
iikirilov:
```suggestion
  public void privateContractMustNotBeAbleToCallSelfDetructOfPublicContract()
```

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/core/ReadOnlyMutableAccount.java
```diff
@@ -0,0 +1,115 @@
...
+
+  private MutableAccount mutableAccount;
+
+  public ReadOnlyMutableAccount(final MutableAccount publicAccount) {
+    this.mutableAccount = publicAccount;
+  }
+
+  @Override
+  public void setNonce(final long value) {
+    throw new UnsupportedOperationException();
```
iikirilov: What are the implications of this if a new OPCode is added that we don't private contracts to call in the public state?

josh-richardson: We get into quite a nasty situation where pantheon tries to re-evaluate this over and over.

ethereum/core/src/main/java/tech/pegasys/pantheon/ethereum/mainnet/MainnetContractCreationProcessor.java
josh-richardson: I've run code coverage and this seems to be right. Will remove.

PR #1919, opened 2019-09-10
Events API: Transaction dropped, sync status, and renames

RatanRSur:
## PR description
Adds the ability to listen to SyncStatus changes and transaction dropped events to the `PantheonEvents` API


ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/state/SyncState.java
```diff
@@ -49,17 +51,22 @@ public SyncState(final Blockchain blockchain, final EthPeers ethPeers) {
         });
   }
 
-  private void publishSyncStatus() {
+  @VisibleForTesting
```
abdelhamidbakhta: (discussion) : why `@VisibleForTesting` if public ?

pantheon/src/test/java/tech/pegasys/pantheon/services/PantheonEventsImplTest.java
```diff
@@ -166,25 +190,55 @@ public void newTransactionEventFiresAfterSubscribe() {
...
+
+    assertThat(result.get()).isNull();
+    // The max pool size is configured to 1 so adding two transactions should trigger a drop
+    transactionPool.addLocalTransaction(TX1);
+    transactionPool.addLocalTransaction(TX2);
+
+    assertThat(result.get()).isNotNull();
+  }
+
+  @Test
```
abdelhamidbakhta: nice test 🎉

ethereum/eth/src/main/java/tech/pegasys/pantheon/ethereum/eth/sync/state/SyncState.java
RatanRSur: I didn't know this before but `@VisibleForTesting` is for documentation purposes only! I used to think it actually changed the visibility for the tests in some special way. It's to communicate that if not for the tests, this method would be less visible.

ethereum/jsonrpc/src/test/java/tech/pegasys/pantheon/ethereum/jsonrpc/health/ReadinessCheckTest.java
```diff
@@ -141,6 +141,7 @@ public void shouldNotBeReadyWhenCustomMaxBlocksBehindIsInvalid() {
   }
 
   private Optional<SyncStatus> createSyncStatus(final int currentBlock, final int highestBlock) {
-    return Optional.of(new SyncStatus(0, currentBlock, highestBlock));
+    return Optional.of(
+        new tech.pegasys.pantheon.ethereum.core.SyncStatus(0, currentBlock, highestBlock));
```
shemnon: If only java had type aliases on import...

PR #1889, opened 2019-08-28
Refactoring Rocksdb as a module

CjHare:
## PR description
Applying the API interfaces and refactoring the RocksDB key value store into a plugin.

PR #1932, opened 2019-09-12
Revert "Refactoring Rocksdb as a module (#1889)"

shemnon:
## PR description

This reverts commit 3e1829f4 / PR #1889

Two failures
1) this no longer works in IntelliJ
2) this no longer works on existing databases that are not segmented.


CjHare:
These two issues should now fixed by https://github.com/PegaSysEng/pantheon/pull/1934 (verified with local testing by myself and @pinges) ...so, hopefully this revert is not going to be required 😉

PR #1937, opened 2019-09-13
Expose in plugins api storage service a getByName so that plugins can use builtin storage plugin

jframe:
## PR description
Expose in plugins api storage service a getByName so that plugins can use builtin storage plugin


plugin-api/src/main/java/tech/pegasys/pantheon/plugin/services/StorageService.java
```diff
@@ -35,4 +36,13 @@
...
   List<SegmentIdentifier> getAllSegmentIdentifiers();
+
+  /**
+   * Retrieves a registered factory corresponding to the supplied factory name
+   *
+   * @param name The name of the factory to retrieve
+   * @return an optional containing the instance of the registered factory, or empty if the factory
+   *     hasn't been registered.
+   */
+  Optional<KeyValueStorageFactory> getByName(final String name);
```
CjHare: `final` modifier not needed on an interface

jframe: nice find. done

PR #1939, opened 2019-09-13
Ban junit assertions

RatanRSur:
## PR description
We prefer the more descriptive assertj for new tests so I thought I'd record some vim macros and hammer out the junits that were still around.


errorprone-checks/src/main/java/tech/pegasys/errorpronechecks/BannedMethod.java
```diff
@@ -39,8 +38,10 @@
 
   private static final ImmutableMap<Matcher<ExpressionTree>, String> BANNED_METHOD_LIST =
       ImmutableMap.of(
-          allOf(staticMethod().onClass("com.google.common.base.Objects").withAnyName()),
-          "Do not use com.google.common.base.Objects methods, use java.util.Objects methods instead.");
+          staticMethod().onClass("com.google.common.base.Objects").withAnyName(),
+          "Do not use com.google.common.base.Objects methods, use java.util.Objects methods instead.",
+          staticMethod().onClass("org.junit.Assert"),
+          "Do not use junit assertions. Use assertj assertions instead.");
```
mbaxter: Very nice!  There's also a `BannedMethodTest` that should probably be updated.

PR #1943, opened 2019-09-14
Support Rollbacks for RocksDB

RatanRSur:
## PR description
Primary:

This fixes an issue where rolling back the default database version could cause an error. It's important that the loading logic depends only on the database version so we can have full control. It's still fine to set auxiliary properties like isSegmentIsolationSupported as we go through the creation process.

Other:
* Extend the plugin interface with `Closeable`


plugins/rocksdb/src/main/java/tech/pegasys/pantheon/plugin/services/storage/rocksdb/RocksDBKeyValueStorageFactory.java
```diff
@@ -71,62 +87,61 @@ public KeyValueStorage create(
...
+            segmentedStorage =
+                new RocksDBColumnarKeyValueStorage(rocksDBConfiguration, segments, metricsSystem);
+          }
+          return new SegmentedKeyValueStorageAdapter<>(segment, segmentedStorage);
+        }
+      default:
+        {
+          throw new IllegalStateException(
+              String.format(
+                  "Developer error: A supported database version (%d) was detected but there is no associated creation logic.",
```
CjHare: nit; `Developer error:` isn't needed, besides it could also be a SysAdmin error (if they move around the db file)

PR #1947, opened 2019-09-17
Final Notice before Archive

EdJoJob:
This is to make it very obvious to everyone that this is no longer the repository for the codebase fomerly known as Pantheon.  We are now Hyperledger Besu!

**Note:** the Jenkins build will never pass on this PR, as there is no Jenkinsfile to run


ajsutton (approved):
LGTM.

